Issue #10, opened 2019-01-29
Fix up error handling


schnie:
Orbit has some functionality for selectively displaying error messages. Some unexpected error messages will be ignored and not displayed. We need to take a pass at triggering error conditions and ensuring that we have the error messages correctly mapped.

One example might be the error displayed when a user enters an invalid password.

ashb:
Just found a new one (making a note as working on something else) if you try to Sign up with an existing email you get a non-specific error page.

schnie:
Good catch, just saw this too. Testing a few other changes, but I'll patch this up in Houston/Orbit.

schnie:
Fixed with https://github.com/astronomer/houston-api/commit/97197714ef54ae0b164b453b26f33a2248435162.

This gets us back to compatible with how Orbit is treating errors, so I think I'll go ahead and mark this as closed for now and ship with 0.8.

I'm going to create a new ticket for 0.9 to re-work how some of these errors are handled to deal with potential user enumeration issues. The fixes for this would be a little more in-depth than our release timeline allowed now.

schnie:
New issue here: https://github.com/astronomer/astronomer/issues/270

PR #17, opened 2019-02-01
Send correct error message on Invalid password

ashb:
Orbit currently relies on the _text_ of the message to display invalid password error. To make it easier in the future we have added a custom `CODE` atttibute on the error that we should switch orbit to use.

This is the bit of orbit that defines the error partial strings it is looking for https://github.com/astronomer/orbit-ui/blob/master/src/modules/auth/Data/helpers.js

Part of (but not all of) #10


schnie:
:+1:

PR #23, opened 2019-02-06
Email confirmation for users before they can sign in

### LINKED ISSUE #9: Support sending and re-sending email verification
We need to add the sending of a verification email when users signup, as well as the ability to re-send the email. The signup mutation already exists, it just needs to be modified to send an email. The re-send mutation needs to be added.

ashb:
This PR adds a bit of prep-work and tidy up of some tests first.

Fixes #9.
 
It then changes the code to:

- send confirmation emails
- create users in the pending state (except for OAuth users, which should be active straight
  away)
- not return a "token" in the AuthUser response if the user can't log in
- add mutations for confirmEmail and resendConfirmation.
- check account status before logging in.


schnie:
Looks great!

PR #34, opened 2019-02-21
Create users as active and confirmed when using an invite token

ashb:
When a user follows an invite token and signs up we know they have
access to the email account specified (it's the only way they could
have gotten the invite URL) so there is no need to confirm it again.

To make mocking of the functions in the module possible they have been
changed from calling `validateInviteToken` to
`exports.validateInviteToken` etc


schnie:
Just pushed up a quick commit to fix a warning about duplicate import of `./index`. Also standardized to use the `userExports` object and renamed the original `exports` to `prismaExports` to match.

ashb:
I did wonder about that duplicate import, but thought it wasn't giving me any style warnings locally. (I could be wrong though!)

PR #35, opened 2019-02-26
Provide a signed JWT token from the Airflow authorization handler

ashb:
The Kubernetes ingress resource is already configured to pass along the
Authorization header from the handler.

This header will be ignored by the "classic" UI, but will be picked up
by a custom SecurityManager for the 1.10/RBAC UI that will create
users in the Airflow metadata store using the information from the
(validated) JWT token.

This is the platform side needed for https://github.com/astronomer/astronomer-fab-securitymanager to work.


ashb:
`aud` needs to be a URL on the python side, not a domain name. This'll need changing.

schnie:
@ashb sounds like we still need to patch up `aud` before merging this in?

ashb:
Yeah, lets hold off on merging this until I've done more testing.

ashb:
Right, no I was mis-understanding an error message. This PR is good to go - audience can be what ever we want it to be, and a domain name (rather than an HTTPS url) is the right value for it.

schnie:
@ashb this doesn't change anything for the existing auth workflows does it?

ashb:
It doesn't, no - we've added the new roles. They'll start receiving an `Authorization` header (which they should ignore).

That said I haven't checked if Grafana etc. might do something with this header. I don't expect them too.

schnie:
@ashb played around with this after merging and got this error in houston after following the oauth flow from orbit:
```
(node:27) UnhandledPromiseRejectionWarning: Error: invalid expiresIn,notBefore option for string payload
    at Object.module.exports [as sign] (/houston/node_modules/jsonwebtoken/sign.js:122:22)
    at createJWT (/houston/dist/lib/jwt/index.js:65:39)
    at _default (/houston/dist/routes/v1/oauth/callback/post.js:100:26)
    at process._tickCallback (internal/process/next_tick.js:68:7)
(node:27) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled
with .catch(). (rejection id: 1)
```
Reverted the PR for now while I test everything for the next release.

ashb:
Hmm curious. I tested it locally and it passed. I wonder if I somehow ended up with a different version of jsonwebtoken installed? (Guessing at the cause)

schnie:
Possibly. I didn't get a chance to dig into it since I was in the middle of testing a handful of other things. It's not required for our next planned release (hopefully this week), so just pulled it out until we can figure out what happened.

ashb:
What version was this runnig? The stack-trace shows line 100, but https://github.com/astronomer/houston-api/blob/38cd2e98618beb560613d4a5d0786c2cd56c32b2/src/routes/v1/oauth/callback/post.js (which was the commit before you reverted this PR) only shows it as having 92 lines.?

PR #36, opened 2019-02-26
Add Mailchimp integration upon app signup

petedejoy:


config/custom-environment-variables.yaml
```diff
@@ -38,6 +38,18 @@ email:
...
+  enabled: 
+    __name: "MAILCHIMP__ENABLED"
+    __format: "json"
+  apiKey: "MAILCHIMP__APIKEY"
+  listId: "MAILCHIMP__LISTID"
+
+registry:
+  auth:
+    __name: "REGISTRY__AUTH"
+    __form: "json"
```
ashb: This whole `registry` section was moved, can you remove it please. (I this was a merge conflict)

src/resolvers/mutation/create-user/index.js
```diff
@@ -35,5 +36,8 @@ export default async function createUser(parent, args, ctx) {
     }
   });
 
+  // Add user to Mailchimp list for onboarding.
+  await addMember(args.email);
```
ashb: Do we want to block on mailchimp on sign-up, or would a 'fire-and-forget' (i.e. remove `await `) be the right thing here?

ashb: Additionally - rather than adding this here and in the oauth route we could add it to the `_createUser` that both of them delegate to in `lib/users/index.js`

schnie:
@ashb we've been debating a little bit about including SaaS tool integrations into Houston.

One idea is to integrate one tool (Metarouter/Segment) and track events, that could integrate with all the sales/marketing tools (other than stripe). 

I think @petedejoy may also be looking at building some of these cloud-only features into Airflow DAGs.

Holding off on merging/not-merging until we figure something out for sure.

schnie:
@petedejoy any progress on airflowing this stuff? Should we close this PR for now? We can re-open if we decide to put it into Houston.

petedejoy:
@schnie I haven't had time to look into it yet- did we decide that we for sure don't want to go this way? Another option is throwing it all in Orbit if you want to keep Houston clean, but then we hit the same issue with EE installs, just in the front end.

schnie:
Skipping this for now, moving towards using analytics.js.

PR #43, opened 2019-04-04
Add userId and email for in-app tracking

benjamingregory:
Passes userId and email address back to in OAUTH response to Orbit for Identify calls.


src/routes/v1/oauth/callback/post.js
```diff
@@ -79,6 +79,10 @@ export default async function(req, res) {
   // Set the cookie.
   setJWTCookie(res, token);
 
+  // Add userId and email for in-app tracking
+  state['extras']['user_id'] = userId
+  state['extras']['email'] = email
```
schnie: @benjamingregory you can probably simplify this to something like `state.extras = { userId, email };`

schnie: And you'll have to change orbit to look at `userId` instead of `user_id`.

benjamingregory: `state.extras` already exists and contains the `onSuccess` redirect prior to appending on userId and email.
Screen Shot 2019-04-04 at 4 09 39 PM

benjamingregory: Not sure I understand what you mean by looking at `userId` instead of `user_id`.

schnie: Ah, my bad - how about `state.extras = { ...state.extras, userId, email }`, to spread the existing extras, and add `userId` and `email`. Was thinking we just change `user_id` to `userId` since we're standardized on camel case for variable names, which means you'd just have to match up the name change on the client side as well.

benjamingregory:
@schnie -- Changed to camelcase and updated orbit as well.

PR #44, opened 2019-04-12
Add unit tests for the `@auth` directive.

ashb:
It's quite an important component so I've also added some unit tests covering it!


schnie:
This is awesome, thanks for doing this @ashb! Pushed one minor thing, just removed an extra `await`. Promises are smart enough to unwrap nested ones, so this seems to work. Tests still pass, so I think it's good to go.

PR #46, opened 2019-04-18
Define explicit relationships with cascading deletes on RoleBinding

ashb:
Without this when deleting a deployment the RoleBinding would stay
around.

The mismatch in name (DeploymentRoleBindings vs RoleBindingToUser) is to
match the existing names. If Prisma had a way of baking migrations to
files we could change this in future, buut it doesn't.

Some useful queries/mutations to find and tidy these up:

```graphql
query allnull{
  roleBindings(
    where: {
      deployment: null,
      workspace: null,
      serviceAccount: null,
      role_not: SYSTEM_ADMIN
    }
  ) {
    id
    role
    deployment {
      id
    }
    workspace {
      id
    }
    serviceAccount {
      id
    }
  }
}

mutation delete_allnull {
  deleteManyRoleBindings(where: {
      deployment: null,
      workspace: null,
      serviceAccount: null,
      role_not: SYSTEM_ADMIN
  }) { count}
}
```


schnie:
@ashb looks good! While we're on the topic, what do you think about checking the generated code into git? Wondering if there's a downside to gitignoring the generated code.

ashb:
If the generated code is exclusively generated based on what is in the datamodel SDL file and doesn't depend upon the state of the (development) DB I'd go for not checking it in.

schnie:
To the best of my knowledge it's purely based on the datamodel. The schema and client should be both generated from `prisma generate`, which uses this config https://github.com/astronomer/houston-api/blob/master/database/prisma.yml#L10-L15.

We run `prisma generate` in our `npm run build`, which should ensure its existence in the final images, as well in our `npm run start`, so I think we're covered. We could try it for a bit and revert if needed.

PR #51, opened 2019-04-24
Grant Deployment Admin to all users with Workspace Admins when creating a deployment

ashb:
This takes us one step closer to not having to "synthesise" roles for
the current user.

Prisma doesn't provide anyway to run a data migration, so we will need
to create a migration script to add these missing roles to existing
deployments.


src/resolvers/mutation/create-deployment/index.js
```diff
@@ -53,6 +57,21 @@ export default async function createDeployment(parent, args, ctx, info) {
...
   const releaseName = generateReleaseName();
 
+  // Find anyone with WORKSPACE_ADMIN on the workspace and give them explicit
+  // DEPLOYMENT_ADMIN on deployment too
+  const workspaceAdminBindings = await ctx.db.query.roleBindings({
+    where: {
+      workspace: { id: args.workspaceUuid },
+      role: WORKSPACE_ADMIN
+    }
+  }).$fragment(`fragment Fields on RoleBinding {
```
schnie: @ashb just to standardize, we could pull this fragment out into the neighboring `fragment.js` (maybe rename to `fragments.js`), and used named exports instead of the current `default` export.

src/resolvers/mutation/create-deployment/index.js
```diff
@@ -53,6 +57,21 @@ export default async function createDeployment(parent, args, ctx, info) {
   // Generate a random space-themed release name.
   const releaseName = generateReleaseName();
 
+  // Find anyone with WORKSPACE_ADMIN on the workspace and give them explicit
```
schnie: @ashb if someone is added as a `WORKSPACE_ADMIN` down the road, they wouldn't have the corresponding `DEPLOYMENT_ADMIN`s on any of the deployments right?

If so, should we care about that? What do you think?

ashb: Yeah, I wasn't sure how to handle that. I think there are a few options:

1) make it clear in the UI that adding someone as a workspace admin won't add them to existing deployments.
2) Handle this automatically in Oribt (so when granting someone workspace admin in Orbit it could also grant them as Deployment Admins)
3) As in 2, but handled in Houston.

src/resolvers/mutation/create-deployment/index.js
ashb: Done.

ashb:
I'm not sure I'm handling ServiceAccounts right. The data model for ServiceAccounts seems to have a 1-1 between SA and RoleBinding, so this code may not _actually_ work. But we create SAs with WORKSPACE_EDITOR/DEPLOYMENT_EDITOR right now anyway, so it may be okay.

ashb:
Probably going to do this differently and cascade permissions from workspace down to all its deployments.

PR #52, opened 2019-04-26
Cascade permissions from workspaces down to deployments

ashb:
This changes the default roles to have deployment level permissions
against the workspace related roles - in effect a "GRANT x ON ALL
DEPLOYMENTS IN WORKSPACE";

hasPermission will now look for a matching permission on the "parent"
entity too - i.e. when asking for `deployment.config.get` on a Deployment,
also look for `deployment.config.get` on the parent workspace too.

Since we cascade from workspace we don't need an explicit grant when
creating new deployments any more either.

This is an alternative approach to #51 

Part of Part of astronomer/astronomer#283


src/lib/rbac/index.js
```diff
@@ -38,21 +36,40 @@ export function hasPermission(user, permission, entityType, entityId) {
...
+      const role = find(roles, { id: binding.role });
+      return role && permission in role.permissions;
+    }
+  });
+}
 
-  // Otherwise return if this role has an appropriate permission.
-  const role = find(config.get("roles"), { id: binding.role });
-  return includes(role.permissions, permission);
+export async function getParentEntity(entityType, entityId) {
```
schnie: @ashb I'm wondering if we could make an adjustment to the `roleBinding` data we get when we set the user on the context earlier in the request, and possibly skip the `prisma.deployment()` call here.

`getUserWithRoleBindings`, and `getServiceAccountWithRoleBindings` both import a fragment that they use when requesting the data for the user. Could we modify this to grab the `deployment { id }` on any `workspace` and use that data in this function instead?

If something like that makes sense, we'd be able to switch this check back to not being `async`. What do you think?

ashb: Oh yeah that probably makes sense. Let me try that.

ashb: We'll have to change the last arg of hasPermission from `entityId` to `entity`, but that's probably better (as with this fn as async it'd be easier to forget making it async and let through by mistake)

ashb: This makes the change a bit bigger (to make sure we get the right fields everywhere) but it is probably the right place to do it.

ashb:
Oh: I think this would have broken the `deployments { id }` query (as it relied on the synthesised roles to get list of deployments otherwise

ashb:
@schnie I've made the change. As expected it needed touching quite a few areas.

I have changed the sig of `hasPermission` etc to take just an entity (object) with a .__typename field instead of type and id strings. I think I got all the instances of it(?). It's the second commit in this PR as a fixup. I'm not sure if it's easier for you to review that commit on it's own or just everything all over.

Do you think its worth adding `"Deployment"` and `"Workspace"` as TYPE_DEPLOYMENT etc to constants?

schnie:
@ashb looks great. Are there any backward compatibility issues we need to worry about? Should we be able to upgrade a customer cluster without anything breaking? _Seems_ like we should be fine, but not totally sure.

ashb:
I _think_ it should be fine, yes as we don't have to add any new RoleBindings.

schnie:
Also, it seems like we could merge in today, and things would behave normally even without the rest of the RBAC related components merged yet.

Trying to pull together a release with all the rest of the commits on master and it seems like we could merge this in too without breaking anything. What do you think @ashb?

config/default.yaml
```diff
@@ -145,86 +145,77 @@ roles:
...
+      ? deployment.serviceAccounts.create
+      ? deployment.serviceAccounts.delete
 
-  - id: WORKSPACE_VIEWER
-    name: Workspace Viewer
-    permissions: []
+  - id: DEPLOYMENT_ADMIN
+    name: Deployment Admin
+    permissions: &PERMS__DEPLOYMENT_ADMIN
+      <<: *PERMS__DEPLOYMENT_ADMIN
```
schnie: @ashb should this be `*PERMS__DEPLOYMENT_EDITOR` to include the `permissions` from the role above?

Or maybe an array referencing the two like you are doing below?

ashb:
Yes should be safe to merge on its own, assuming I got all the cases uses of `has`/`checkPermission`

config/default.yaml
ashb: Um yeah. this ain't right.

ashb:
Clicking around the UI things still seem to work. It's not a very exhaustive test though.

The main thing that I think we might need to think about in the future is what permissions are given to what roles.

src/lib/rbac/index.js
```diff
@@ -35,24 +35,43 @@ export function hasPermission(user, permission, entityType, entityId) {
...
-  // Otherwise we have to ensure the user has access to the entity.
-  const binding = find(user.roleBindings, binding => {
-    // Return true if we're looking for a scoped permission,
-    // and this binding has the entity type we're interested
-    // in and the id matches.
-    const hasEntity = !!binding[entityType];
-    return hasEntity && binding[entityType].id === entityId;
+  const entityType = entity.__typename.toLowerCase();
+
+  const [parentType, parentId] = exports.getParentEntity(entity);
```
schnie: Not important at all, but this could be `getParentEntity()`, without the `exports`, unless there's a reason for this.

src/resolvers/query/deployments/index.js
```diff
@@ -89,10 +82,35 @@ export function deploymentsQuery(args, ctx) {
 
   if (!(deploymentUuid || releaseName || workspaceUuid)) {
     // Get a list of deployment ids that this user can access.
-    const deploymentIds = ctx.user.roleBindings.map(rb =>
-      rb.deployment ? rb.deployment.id : null
-    );
-    query.where.AND.push({ id_in: compact(deploymentIds) });
+    const deploymentIds = new Set();
```
schnie: @ashb not that we should change anything now, but I'm thinking we may want to think about restructuring our deployment fetching queries to account for RBAC.

Originally was just patching around some shortcomings in the first version of this query without breaking the API that Orbit and the CLI used, but planned on fixing the root issue as a follow up at some point.

If we restructured the API, do you think we could drop a lot of our RBAC logic here? I'm thinking we'd ideally structure the API so that we can properly use our `auth` directive. Interested to hear what you think.

src/lib/rbac/index.js
ashb: Oh yes it can now - it was `exports.` so that I could mock it to avoid hitting DB in the tests. But with the new object/fragments that isn't needed now.

src/resolvers/query/deployments/index.js
ashb: Yeah, we-working this API sounds good. Perhaps with the new role cascading we could remove the need to check different permissions for workspace and deployment which with also simplify the code.

ashb: Or we could probably quite easily extend the `@auth` to work on an arugment:

```
type Query {
  deployments(
    workspaceUuid:  Uuid @auth(permission: "workspace.deployments.list")
    releaseName: String @auth(permission: "deployment.config.get")
  ): [Deployment] @auth 
}
```

schnie: That could work. Does the `@auth` directive work on arguments like that? I was thinking it was extended to work on fields on the return type.

I was thinking we'd separate the queries a bit, like:

```
type Query {
  deployment(
    releaseName: String
  ): Deployment @auth(permission: "deployment.config.get")

  deployments(
    workspaceUuid:  Uuid
  ): [Deployment] @auth(permission: "workspace.deployments.list")
}
```

ashb: It doesn't currently work on arguments but could be extended to support it easily enough.

ashb:
Found a cased that this doesn't handle: createServiceAccount mutation.

ashb:
Damn it. I may have made this more complex than needed.

Rather than fetching the parent type in hasPermission I should have fetched it in the role bindings that we put on ctx.user. That probably means we don't need to do things like:

```js
  if (workspaceUuid) {
    const entity = await ctx.db.query.workspace(
      { where: { id: workspaceUuid } },
      rbacFragments.workspace
    );
    checkPermission(ctx.user, "workspace.deployments.list", entity);
  }
  if (deploymentUuid) {
    const entity = await ctx.db.query.deployment(
      { where: { id: deploymentUuid } },
      rbacFragments.deployment
    );
    checkPermission(ctx.user, "deployment.config.get", entity);
  }

  if (releaseName) {
    const entity = await ctx.db.query.deployment(
      { where: { releaseName } },
      rbacFragments.deployment
    );
    checkPermission(ctx.user, "deployment.config.get", entity);
  }
```

(where I had to load an object in the query just to pass it to checkPermission with the right info it needed. That info should (or could) already be available tunder ctx.user.roleBindings :(

ryw:
just noticed this old PR sitting out here

schnie:
I think we can close this. I think this is already implemented and we'll probably need to add actual deployment rbac soon.

PR #54, opened 2019-05-02
Specify role when adding users to a workspace

ashb:
This lets us specify the role we add a new user with.

To make the upgrade path easier we also default the value in the mutation to the previous behaviour, this way using an old client will still work.


src/schema.graphql
```diff
@@ -332,6 +333,8 @@ type Mutation {
   workspaceAddUser(
     workspaceUuid: Uuid
     email: String!
+    # For backwards compatibility specify the previous behaviour as default value
+    role: Role! = WORKSPACE_ADMIN
```
ashb: @andriisoldatenko This is how I have changed the mutation for adding a user.

ashb:
Think this one is good to go as it stands @schnie.

PR #55, opened 2019-05-02
Add mutation to change the Workspace role for a User.

ashb:
This works for users and invitees.

Part of astronomer/astronomer#283


schnie:
LGTM

PR #59, opened 2019-05-10
Feature/ui permissions

petedejoy:
Accidentally closed my other PR when trying to flip it live 🤦‍♂ . This is to go along with my Orbit PR- it adds the `workspaceCapabilities` type into the schema which will allow us to render UI components keyed off of permissions & other logic handled by Houston.


src/resolvers/types/workspace/index.js
```diff
@@ -29,4 +33,25 @@ export function deploymentCount(parent) {
   return size(parent.deployments);
 }
 
-export default { users, groups, invites, deploymentCount };
+export function workspaceCapabilities(parent, args, ctx) {
+  const editPermissions = hasPermission(
```
schnie: @petedejoy how about `updateIAM` instead of `editPermissions` to match the underlying permission name?

schnie:
Also looks like there might be a linting error, drone will show you the issue, or `npm run lint` locally.

PR #65, opened 2019-06-03
Prometheus Metrics Passthrough

samblackk:
This feature polls a series of Prometheus Queries (PromQL) via HTTP API and passes data through a `metrics` GraphQL Subscription. 

Port-forward prometheus so you can use localhost: 
```kubectl port-forward svc/aspiring-salamander-prometheus 9090:9090``` 

To test in playground you can use the following:
```
    subscription Metrics($deploymentUuid: Uuid!) {
      metrics(deploymentUuid: $deploymentUuid) {
        label
        result
      }
    }
```

Issue: https://github.com/astronomer/astronomer/issues/268

Please r/m

### CODE REVIEW

File: src/resolvers/subscription/metrics/index.js [hunk 1]
```diff
@@ -0,0 +1,60 @@
...
+  let { releaseName } = await db.query.deployment(
+    { where: { id: args.deploymentUuid } },
+    `{ releaseName }`
+  );
+
+  // To test on localhost uncomment this line
+  // releaseName = "boreal-jets-9045";
+
+  // Poll interval
+  const interval = 1000;
```
ashb: How often does our Prom collect metrics? Is it worth it to have it query every second?

samblackk: @ashb Grafana polls every 20s. We can replicate that timing. I set the polling to every second to match the polling frequency of the logs.

src/schema.graphql
```diff
@@ -436,4 +441,8 @@ type Subscription {
     timestamp: DateTime
     search: String
   ): DeploymentLog @auth(permission: "deployment.logs.get")
+
+  metrics(
+    deploymentUuid: Uuid!
+  ): [Metric] @auth(permission: "deployment.logs.get")
```
schnie: @samblackk thinking we may want to make a new permission here - `deployment.metrics.get`. All we'd need to do is change it here and then add it down in the `roles` section of `config/deployment.yaml`. We can duplicate the new permission everywhere the logs one is. Same end result but could possibly give enterprise users the ability to tweak individually if they want.

src/resolvers/subscription/metrics/index.js
schnie: The default settings of this particular dashboard in grafana are a 30min timeframe and refresh every 10s. That might be a fine place to start for most of these things.

What do y'all think about making these poll intervals config values instead of hard coded? Realized I forgot to do that for the elastic stuff.

samblackk:
@schnie @ashb All requested changes are resolved.

schnie:
@samblackk think there's anything worth writing a test or two for? Not thinking anything too crazy yet, just someting to give this module a little exercise.

src/resolvers/subscription/metrics/queries.js
```diff
@@ -0,0 +1,146 @@
...
+const namespace = `.*-.*-.*-[0-9]{4}`;
+
+// Poll interval (milliseconds)
+const step = config.get("prometheus.pollInterval");
+
+// UNIX start timestamp, minus step
+const start = Math.round((new Date().getTime() - step) / 1000);
+
+// UNIX current time timestamp
+const now = Math.round(new Date().getTime() / 1000);
```
ashb: This would only be evaluated once per node.js process at startup. Is that the behaviour we want?

config/default.yaml
```diff
@@ -78,6 +77,15 @@ elasticsearch:
   # If in NODE_ENV is development and this is enabled, generate sample log records.
   mockInDevelopment: true
 
+  # Interval that data is grabbed and sent
```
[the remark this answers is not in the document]
samblackk: Yes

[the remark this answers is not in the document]
samblackk: Yes

src/resolvers/subscription/metrics/queries.js
samblackk: Good catch @ashb - Fixed in the latest push.

samblackk:
@schnie Explored adding a unit test - however the major parts of the function are already covered by `createPoller` unit test. It's redundant to recreate it here.

src/resolvers/subscription/metrics/queries.js
```diff
@@ -0,0 +1,146 @@
...
+      query: query(`sum(airflow_ti_failures{deployment=~"${deployment}"})`)
+    },
+    {
+      name: "connectCount",
+      query: query(
+        `sum(pgbouncer_exporter_database_current_connections{name!="pgbouncer", deployment=~"${deployment}"})`
+      )
+    },
+    {
+      name: "waitingCount",
```
ashb:
```suggestion
      name: "dbConnwaitingCount",
```

Not suer where these names are used, but might be a bit ambigious otherwise? Similar for `connectCount`

samblackk: This makes sense.

src/resolvers/subscription/metrics/queries.js
```diff
@@ -97,7 +97,7 @@ export default function queries(deployment) {
       )
     },
     {
-      name: "waitingCount",
+      name: "dbConnwaitingCount",
```
ashb: Sorry I typoed the suggestion :(

samblackk:
Rebased and ready to go!

samblackk:
@schnie Queries are updated to better support the charts needed in Orbit.

PR #68, opened 2019-06-07
Ignore generated Prisma client/schema files

ashb:
These files are automatically generated every time we start up Houtson,
and if we want to get them back we can run `npm run build`


schnie:
Thanks for doing this!

PR #71, opened 2019-06-13
Feature/trial-UX

petedejoy:
This PR does a few things:

**Paywall/Suspension Functionality**
 - Adds `isSuspended` flag and `trialEndsAt` date to the Workspace table in the datamodel
 - Adds a mutation `suspendWorkspace` that allows a sysAdmin to control that flag
 - If `isSuspended == true` or the user is past their trial window (ie. the time now is after their `trialEndsAt` date, the user is blocked from creating deployments at the API layer and returns a `paywallEnabled` flag via the `workspaceCapabilities` type to orbit.

**Trial Limitations**
 - Uses the existence of a `stripeCustomerId` in the database to limit functionality
 - If a workspace has no associated payment information (ie. `stripeCustomerId == null`), users in that workspace cannot update resource configs on their deployments _or_ make more than one deployment.

**Other Small Fixes**
- Separated `billingEnabled` and `canUpdateBilling` into two separate bools on the `workspaceCapabilities` type. This grants us more granular control of what we're conditionally rendering in the UI.
 - Removed `active` field from Workspace type, as it was not formerly being used.


src/schema.graphql
```diff
@@ -424,6 +426,11 @@ type Mutation {
     company: String
     token: String!
   ): Card @auth(permission: "workspace.billing.update")
+
+  suspendWorkspace(
+    workspaceUuid: Uuid!
+    isSuspended: Boolean!
+  ): Workspace @auth(permission: "system.iam.update")
```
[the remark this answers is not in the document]
petedejoy: Yeah I'm cool with that! Sorry, forgot to ask how you wanted to handle it. I'll add that permission 
and update the PR now.

schnie:
Also, we already have `active` as a field on `Workspace`. I'm thinking that was meant to serve the same purpose in Houston1. I doubt it's being used, but want to make sure and remove it?

petedejoy:
Oh yeah I never noticed that before. It looks like we're only using the `active` flag on the user level and then we just set it and forget it in the workspace table. Should I remove it from the datamodel for the Workspace too?

schnie:
Probably, since we're replacing it. Orbit may need to update to cope with that change.

schnie:
Yea probably this one since we're technically replacing it (kinda).

petedejoy:
@schnie tests are in here and passing! Should be good to begin review on this.

config/default.yaml
```diff
@@ -176,14 +181,15 @@ roles:
     permissions:
       - system.monitoring.view
       - system.iam.update
+      - system.workspace.suspend
+      - system.workspace.extend
```
schnie: @petedejoy maybe `extendTrial` to be a little more specific?

src/resolvers/mutation/add-card/index.js
```diff
@@ -20,7 +21,9 @@ export default async function addCard(parent, args, ctx) {
 
   await ctx.db.mutation.updateWorkspace({
     data: {
-      stripeCustomerId: stripeCustomerId
+      stripeCustomerId: stripeCustomerId,
+      isSuspended: false,
+      trialEndsAt: moment().format() //Changes trialEndAt timestamp to be now
```
schnie: `new Date()` might work here to simplify

src/resolvers/types/workspace/index.js
```diff
@@ -33,27 +34,42 @@ export function deploymentCount(parent) {
...
-  return { canUpdateIAM, canUpdateBilling };
+  // Check for hard override of trial paywall logic and throw paywall
+  const workspace = await ctx.db.query.workspace(
+    { where: { id: parent.id } },
+    `{ trialEndsAt, isSuspended }`
+  );
+
+  const now = moment();
+
+  let isTrialing = moment(workspace.trialEndsAt).isAfter(now) ? true : false;
```
schnie: This seems like it could be declared with `const`, and I think you can drop the ternary expression since `isAfter` will already return a `Boolean`.

src/resolvers/types/workspace/index.js
```diff
@@ -33,27 +34,42 @@ export function deploymentCount(parent) {
...
+
+  let paywallEnabled = workspace.isSuspended ? true : !isTrialing;
```
schnie: Can prob `const` this too

src/schema.graphql
```diff
@@ -137,6 +136,8 @@ type DeploymentConfig {
 type WorkspaceCapabilities {
   canUpdateBilling: Boolean
   canUpdateIAM: Boolean
+  billingEnabled: Boolean
+  paywallEnabled: Boolean
```
schnie: Thinking about where these belong... :thinking:

petedejoy: I'll follow you guys on this one. I do agree it doesn't belong here forever, but think it'd be ok to leave it for now then work to build a system-level config, permissions field, etc as part of a different PR.

petedejoy:
@schnie Schema structure is fixed here. Now to figure out why my formatter all the sudden started doing things differently...

src/resolvers/mutation/add-card/index.js
```diff
@@ -20,7 +20,9 @@ export default async function addCard(parent, args, ctx) {
 
   await ctx.db.mutation.updateWorkspace({
     data: {
-      stripeCustomerId: stripeCustomerId
+      stripeCustomerId: stripeCustomerId,
```
schnie: @petedejoy it might be clearer to do what we do in some other places and define some variables ahead of time and pass into the mutation, like...

```
const data = { stripeCustomerId: blah, blah: blah };
const where = { blah: blah };
await ctx.db.mutation.updateWorkspace({ data, where });
```

src/resolvers/mutation/create-deployment/index.js
```diff
@@ -38,12 +38,20 @@ export default async function createDeployment(parent, args, ctx, info) {
     {
       where: { id: args.workspaceUuid }
     },
-    `{ stripeCustomerId }`
+    `{ stripeCustomerId, isSuspended, deployments { id } }`
```
schnie: We could also separate the `where` out here and import a fragment from `./fragments.js`, where we can keep this.

src/resolvers/mutation/extend-trial/index.js
```diff
@@ -0,0 +1,12 @@
+export default async function extendTrial(parent, args, ctx) {
+  const workspace = await ctx.db.mutation.updateWorkspace({
+    data: {
+      trialEndsAt: args.trialEndsAt
+    },
+    where: {
+      id: args.workspaceUuid
+    }
+  });
+
```
schnie: @petedejoy to flatten things out you could do something here too like...

```
const { trialEndsAt, workspaceUuid } = args;
const data = { trialEndsAt };
const where = { id: workspaceUuid };
return ctx.db.mutation.updateWorkspace({ data, where });
```

You can also skip the `await` here and just return the promise and it should work as expected. May want to test that.

petedejoy: ✅

src/resolvers/mutation/add-card/index.js
petedejoy: ✅

src/resolvers/mutation/create-deployment/index.js
petedejoy: Pulled out the where here, but not completely sure I know how to abstract this into fragments.js. Would we just add this info to the fragment and return `addFragmenttoInfo(info, fragment)` here?

schnie: Yea I think the current file exports a `default` fragment. We'll probably want to refactor and use two named exports.

petedejoy: Hm, so what would that look like? Two separate fragments? I assume we don't want to add all those fields to every info that gets returned throughout that mutation

schnie: You can just take the fragment you have inline right now and move that to a separate, named export in `fragments.js` and import it by name and plop it in place of your current string.

petedejoy: Sorry, should've been more clear- there already exists a fragments.js that exports a default fragment for `deployment`. Don't want to override that or mess anything up, what would setting a new one up in that file look like?

src/lib/oauth/providers/okta/index.js
```diff
@@ -9,14 +9,15 @@ import querystring from "querystring";
...
+
 /*
  * Generate the authentication url for Okta.
  * @return {String} The authentication url.
  */
 export function authUrl(state, redirectUrl, integration = "self") {
   const params = {
     client_id: cfg.clientId,
-    response_type: "code id_token token",
+    response_type: "id_token token",
```
schnie: @petedejoy is this supposed to be in here?

petedejoy: It's not specific to this PR, but it's supposed to be a change made in Houston. That `code` is only needed for Auth Code flow, and as we discussed we use Implicit. Removing it just cleans things up without changing the functionality.

Would you rather me do it as a separate PR?

schnie: No worries, just making sure it wasn't something that was leftover.

PR #72, opened 2019-06-13
Add 0.8.x to 0.9.x migration

schnie:


schnie:
This is a little goofy and we may want to get rid of this as soon as prisma 2 is released, but here's a first crack at a migration here. Its weird because we have to grab the data from the old tables, _then_ run `prisma deploy --force`, _then_ add that data back into the new tables/columns.

src/migrations/20190612121012_fix_relationships.js
```diff
@@ -0,0 +1,115 @@
...
+  // Collect all the records from the old join tables (A,B mappings).
+  const data = await Promise.all(
+    ops.map(op => {
+      log.debug(`Getting records for ${op.src}`);
+      return knex
+        .withSchema(schema)
+        .select("*")
+        .from(op.src);
+    })
+  );
```
ashb: Could we write this data out to a temporary file on disk (as JSON probably?), _just_ in case something bad happens during the migration that way we at least have the relastionship data to recover from, if manually.

package.json
```diff
@@ -4,15 +4,17 @@
...
     "test": "prisma generate && jest",
     "coverage": "jest --coverage",
     "lint": "prisma generate && eslint src --ignore-pattern src/lib/generated --ext .js --ext .graphql",
     "token": "prisma token",
     "playground": "graphql playground",
+    "migrate": "npm run knex-migrate && npm run prisma-deploy",
     "knex": "babel-node node_modules/.bin/knex",
-    "migrate-latest": "npm run knex migrate:latest"
+    "knex-migrate": "npm run knex migrate:latest",
+    "prisma-deploy": "prisma deploy --force"
```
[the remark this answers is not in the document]
schnie: Yea, thinking through this again, I think we can make our default `prisma deploy` a non-forced one and just make the one that the knex migration calls a `--force`, which we do need to actually automate it and get it to go through.

src/migrations/20190612121012_fix_relationships.js
schnie: Yea, good call, was thinking about that too, but we really only have a handful of users that will ever hit this migration, so I skipped it. I'm going to merge this and cut an alpha so we can test a full upgrade. I might follow up in a bit to do this.

Issue #73, opened 2019-06-17
Passing email into delete user mutation removes all users in workspace


vparekh94:
When trying to delete a user from a workspace, a user tried passing in an email instead of a UUID:

```
mutation WorkspaceRemoveUser {
  workspaceRemoveUser (
    workspaceUuid: "cjx0owkil00pt0c05p74ohl8o"
    email:"vmpvmp94@gmail.com"
  )
  {
    uuid
    label
    description
    users { emails {address} }
  }
}
```

This resulted in all users in the workspace. When deleting from the UI, a UUID is used instead of an email:

```

mutation removeUser($id: Uuid!, $workspaceId: Uuid!) {
  workspaceRemoveUser(userUuid: $id, workspaceUuid: $workspaceId) {
    id: uuid
    __typename
  }
}

```

PR #74, opened 2019-06-17
Enforce fields on remove user mutation and remove email as an arg

petedejoy:
Addresses the issue we ran into with City of Montreal (https://github.com/astronomer/houston-api/issues/73) where attempting to delete a user while passing an email removed all users from the workspace.

This PR enforces that `userId` and `workspaceId` are passed with the mutation. It also removes email as a passable arg.


schnie:
Thanks @petedejoy!

PR #80, opened 2019-06-28
Metrics: Allow metric group selection

samblackk:
Updates the metrics subscription to received an array of `enums` from the client and returns only the data that matches the keys provided. This paves the way to "subscribing to specific metrics" in Orbit. 

Attached to https://github.com/astronomer/astronomer/issues/339

Do not merge until accompanying Orbit PR is ready - this will break the current metrics tab. 
Please review!


schnie:
@samblackk playing around with this now, looks great! Is there a way we could have houston fallback to some default mode in development to avoid a `DAG gone it!` message if we don't have a prometheus listening on localhost? I added something for the logging stuff, so it'd fallback by default to some fake logs, but configurable so I could hook up to a live ES cluster. Mostly just so it works out of the box if a new developer runs the project(s).

samblackk:
@schnie Yeah, we can make this happen. I'll throw a flag in the config for prom testing and we can use to trigger pulling test data from Houston.

samblackk:
@schnie Sample data is GTG.

schnie:
@samblackk awesome! Playing around with this again and I noticed that when I first load the page or add a new metric group the view completely refreshes, temporarily just showing the "Add a metrics group" button for a second. Wondering if there's something we can do to prevent the flash or make it a little less jumpy :thinking:

schnie:
@samblackk are you able to reproduce this ^. Happy to jump on a call and show you what I mean.

samblackk:
@schnie yeah - I can. working on a solution for it.

samblackk:
Just an update - this is GTG. Can't fix the "flash" that happens between adding groups due to the Apollo HOC. The refactor I'm working on will fix this.

PR #85, opened 2019-07-09
Rework auth providers to use OpenID Connect library

ashb:
This does a few things:

   - it reduces similar/duplicated code needed for the (at present) three Auth integrations we've got (Auth0, Google, Okta)
   - uses the IdP published discovery document to find URLs and signing keys.
   - makes the checks of the OAuth2 tokens more stringent (checks JWT signing keys, at_hash, issuer and much more for us out of the box).
   - Allows us to more easily add new OpenID Connect providers without having to add a custom auth provider modules

(The way Orbit gets/shows the auth providers should change to be more dynamic. That will be a future change)


src/routes/v1/oauth/redirect/static/index.html
```diff
@@ -25,8 +25,9 @@
         var params = location.hash.replace('#', '').split("&");
         for(var i = 0; i < params.length; i++) {
           var parts = params[i].split("=");
+          parts[1] = decodeURIComponent(parts[1]);
```
ashb: This was needed as the format of some of the token comes in URI-escaped for some IdPs (`+` as `%2b` for microsoft for example). Without this change the new extra checks that openid-client adds were failing.

src/lib/oauth/config/index.js
```diff
@@ -1,8 +1,55 @@
...
+      authUrlParams: {
+        audience: "astronomer-ee"
+      }
+    };
+  }
+
+  if (config.has("auth.google") && config.get("auth.google.enabled")) {
+    let cfg = config.get("auth.google");
+
+    providerCfg["google"] = {
```
ashb: This will enable the old `AUTH__GOOGLE__CLIENT_ID` env var to keep working - and I think that's useful to be able to continue setting the common settings that way.

ashb:
@petedejoy This should fix our Okta URL woes.

petedejoy:
Awesome, thanks @ashb! Did you get a chance to test the okta integration with our dev cluster and the new URL they provided?

ashb:
Yeah I tested it on our Okta dev account, the url included `/oauth2` and logged in successfully.

petedejoy:
💯 💯 💯 Thanks a ton!

config/default.yaml
```diff
@@ -118,31 +118,39 @@ auth:
...
+      # If blank we will go via Auth0
+      clientId: ~
+      discoveryUrl: https://accounts.google.com
+
+    okta:
+      enabled: false
+      clientId: ~
+      discoveryUrl: ~
 
   # Github oauth configuration.
```
schnie: @ashb unrelated, but we can get rid of this direct github config block right?

src/lib/oauth/config/index.js
```diff
@@ -1,8 +1,57 @@
...
+export const DEFAULT_CLIENT_ARGS = {
+  scope: "openid profile email",
+  response_type: "token id_token",
+  response_mode: "fragment"
+};
+
+// "Upgrade" the easier-to-set auth0/okta/google auth configs to use OpenID
+// connect
+function upgradeConfigs() {
+  if (config.has("auth.auth0") && config.get("auth.auth0.enabled")) {
```
petedejoy: I get an error thrown here when `npm start`ing. It looks like the config has been changed so that `auth0` is nested under `auth.openidConnect`.

src/lib/oauth/config/index.js
schnie: I guess this is related to my other comment, but wondering if we should nest the env vars like the `default.yaml`. It was nice being able to not upgrade my `.env` file, but we control most of these env vars in production envs via helm for most right?

While we were building Houston 1, we kept a ton of code like this to upgrade various things as we were drastically changing the platform pretty rapidly back then. The amount of code we had for this stuff got a little out of hand and ended up being another factor in scrapping the whole thing.

Do you see this type of code being deprecated over time or staying around?

src/lib/oauth/config/index.js
schnie: Hmm, mine booted right up. I am wondering about syncing up the env var names (managed separately) with the new config structure though.

petedejoy: Hm that's weird- changing these to `auth.openidConnect.auth0` and `auth.openidConnect.auth0.enabled` fixes the issue for me 🤔.

schnie: How are you setting your config? Are you modifying `default.yaml` directly or using `.env`?

petedejoy: Looks like it's getting set from my default.yaml. I should probably override it with a .env, but isn't it still a problem that the default config doesn't work?

petedejoy: Ah scratch that, I get your comment about syncing up env var names with the config structure now :)

ashb: I tried to leave the existing config names working (there's the `upgradeConfig` fn that should handle it) I may have missed an edge case or two

config/default.yaml
ashb: Sort of - it's not strictly an oidc provider it goes via Auth0, but is handled specially so I'm not sure.

src/lib/oauth/config/index.js
ashb: Possibly we could have `AUTH__GOOGLE__ENABLED` env var set the auth.openidConnect.google` setting - I didn't think about that option

ashb: My thinking for keeping these was it's quite handy to have simple env vars to tweak settings without having to go full in to passing a json block like this:

```
AUTH__GITHUB__ENABLED=true \
AUTH__OPENID_CONNECT_PROVIDERS='{"microsoft": { "enabled": true, "authUrlParams": {"response_mode": "fragment"}, "discoveryUrl": "https://login.microsoftonline.com/common/v2.0", "clientId": "9a665ede-0e00-4a0b-9ca1-75d3956109de"}}' \
npm start
```

schnie: Yea I was kind of thinking we could move `AUTH__GOOGLE__ENABLED` over to `AUTH__OPENID_CONNECT__GOOGLE__ENABLED` and so on. Definitely verbose, but primarily concerned with staying correctly nested to match the `default.yaml` since this would be the only exception to that.

ashb:
Cool I'll change this. Do we have a place to note upgrading instructions?

src/lib/oauth/config/index.js
ashb: Removed this function and renamed the config vars to match new structure, anyone using these on an enterprise install will need to update the config.

PR #88, opened 2019-07-11
Sysadmin: Delete Deployment

petedejoy:
This PR allows anyone with the `system.deployment.delete` permission to run the `deleteDeployment` mutation. It will enable System Admins (ie. me and @paolaperaza for now) to delete rogue deployments that aren't paid for on new Cloud (think: someone signs up for a trial, doesn't convert, but leaves a deployment hanging). 

This also sets the stage for more granular permissions management going forward- we're now able to pass multiple permissions as an array with each mutation and a `PermissionsOperator` enum that tells the auth directive whether to check for all of the permissions (`AND`) or just one of them (`OR`).


schnie:
Should we change `permission` to `permissions`? `permissionOperator` feels a little verbose too, maybe just `op` or something like that? Trying to avoid a crazy looking schema file.

petedejoy:
Yeah I was thinking about that permission-->permissions change. Definitely makes more sense to me since it's now an array regardless of how many there are. Could we just call it `Operator`? It's just `AND` and `OR` - no reason that needs to be specific to permissions right?

src/schema.graphql
```diff
@@ -28,13 +29,18 @@ enum EntityType {
   SYSTEM
 }
 
+enum Operator {
```
ashb:
```suggestion
enum PermissionOperator {
```

Otherwise Operator as a top-level enum is a little bit generic.

src/schema.graphql
```diff
@@ -28,13 +29,18 @@ enum EntityType {
...
+}
+
 #
 # Types
 #
 type AuthUser {
   user: User
   token: Token
-  permissions: JSON
+  permissionss: JSON
```
ashb: Bad search-and-replace here :)

src/schema.graphql
ashb: (but keep the kwarg to the directive as `op`)

src/schema.graphql
```diff
@@ -5,7 +5,8 @@
 #
 # Implements authorization rules.
 directive @auth(
-  permission: String
+  permissions: [String]
+  op: Operator
```
ashb: I think we can do this
```suggestion
  op: Operator = AND
```

That way we don't need to deal with the `null` case in the JS

src/lib/directives/auth/index.js
```diff
@@ -67,8 +71,19 @@ export default class AuthDirective extends SchemaDirectiveVisitor {
...
+        // true for all permissions in the array
+        const deny =
+          this.args.permissionOperator == null ||
+          this.args.permissionOperator == "AND";
+
+        permissions.map(p => {
+          if (deny && !this.hasPermission(ctx.user, p, entityType, entityId))
+            throw new PermissionError();
+
+          this.checkPermission(ctx.user, p, entityType, entityId);
```
ashb: I'm not sure this logic is right. Could you extend `src/lib/directives/auth/index.unit.test.js` to exercise the OR case?

(also there's a directive defined in the mini schema in that file which will need updating).

src/schema.graphql
ashb: Oh though this _is_ generic, so maybe that's okay here.

src/lib/directives/auth/index.js
ashb:
```js

    if (requireAllPermissions) {
        for (let p of permissions) {
            this.checkPermission(ctx.user, p, entityType, entityId);
        }
    } else {
        (() => {
            for (let p of permissions) {
                if (this.hasPermission(ctx.user, p, entityType, entityId)) {
                    // Having any of the permissions is enough.
                    return;
                }
            }
            // If we get to here no permissions passed.
            throw new PermissionError();
        })
    }
```

src/schema.graphql
petedejoy: Yeah - it's really just `AND`/`OR` and isn't specific to permissions at all. I was thinking maybe we'll have use for it elsewhere- what do you think the best bet is?

src/schema.graphql
petedejoy: 🤦‍♂

src/schema.graphql
petedejoy: Ah that's a good call. That null case is annoying.

src/lib/directives/auth/index.js
petedejoy: That logic looks good- I was thinking about this last night and think I need to change a bunch of this up today, as we pass that permissions object in a few places around Houston and things might have gotten messed up when we redefined the permissions as an array. 

I think I'm going to create a separate helper that handles all of this logic (looping thru the permissions array) and call that in the auth directive and other places it's needed.

ashb: Sounds good - makes sense for that to live in rbac/index.js alongside hasPermission etc.

schnie:
@petedejoy this looks like it may have some commits from other branches that are already merge in it. Seems to contain a lot of the things @ashb already has committed on master. Seems like you only really changed 2 or 3 files but its showing 18.

petedejoy:
Lemme rebase and try to clean it up. I also am still hacking on fixing these unit tests- hoping to have 'em done today, then this will be ready to roll.

petedejoy:
Nightmare of a rebase, but should be cleaned up here, @schnie. Now onto that unit test...

src/schema.graphql
```diff
@@ -331,23 +337,27 @@ type Mutation {
...
     config: JSON
     env: JSON
     sync: Boolean
-  ): Deployment @auth(permission: "deployment.config.update")
+  ): Deployment @auth(permissions: ["deployment.config.update"])
 
   deleteDeployment(
     deploymentUuid: Uuid!
-  ): Deployment @auth(permission: "deployment.config.delete")
+<<<<<<< HEAD
```
schnie: @petedejoy looks like some merge conflicts here

src/lib/directives/auth/index.js
```diff
@@ -67,8 +72,20 @@ export default class AuthDirective extends SchemaDirectiveVisitor {
...
+
+        const fn = this.args.permissionOperator === "AND" ? every : some;
+        const check = curry(this.hasPermission)(
+          ctx.user,
+          _,
+          entityType,
+          entityId
+        );
+        const allowed = fn(permissions, check);
+        console.log(allowed);
```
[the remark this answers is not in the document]
petedejoy: what the...I don't have this log locally. It probably got roped in with the rebase than I deleted it and re-committed. Everythings probably getting caught by drone, but the unit test for this directive is still failing.

src/schema.graphql
```diff
@@ -258,7 +264,7 @@ type Query {
     workspaceUuid: Uuid
     deploymentUuid: Uuid # Ignored, only here for backward compatibility.
     releaseName: String
-  ): [Deployment] @auth # Permissions checked in resolver, due to limitations with current API design.
+  ): [Deployment] @auth #  checked in resolver, due to limitations with current API design.
```
schnie: @petedejoy looks like ya might have accidentally deleted this

schnie:
@petedejoy looks like we may have a merge conflict now. We merged some new metrics updates that also added another `enum` to the schema. Looks like its confused, we want to keep both.

PR #89, opened 2019-07-12
Support authenticating with Google via Auth0 "disabled"

ashb:
The default config we ships has Auth0 disabled but Google and Github
enabled. For this to work we need to ask for the "google" provider when
processing the callback, not auth0.

I'm not 100% happy with this solution, but it does work.


schnie:
Cool, good enough for now - pretty much how it was before. Definitely a bit awkward.

PR #90, opened 2019-07-15
Add config flag for clock skew and implement in oauth lib

petedejoy:
This PR addresses the local dev issue I had where I'd get ```AssertionError [ERR_ASSERTION]: id_token issued in the future``` when trying to authenticate into the app. It adds a `auth.openidConnect.clockTolerance` config flag to set a clock skew for the client. It's currently defaulted to 5 seconds.


config/default.yaml
```diff
@@ -119,6 +119,8 @@ auth:
     enabled: true
 
   openidConnect:
+    # Clock skew for slightly out-of-sync clocks
+    clockTolerance: 5
```
ashb: Lets add a environment variable setting for this to in config/custom-environment-variables.yaml.

And perhaps leave this as the default of `0` as it works fine for most people.

petedejoy: Ah yes that's a great call. One moment...

petedejoy: ✅

config/custom-environment-variables.yaml
```diff
@@ -78,6 +78,7 @@ auth:
       __format: "json"
 
   openidConnect:
+    clockTolerance: "AUTH__OPENID_CONNECT__CLOCK_TOLERANCE"
```
ashb: And we might need to make this json-y so that the value set by an env var is a number, not a string.

PR #91, opened 2019-07-16
Only use redirector.astronomer.io for default Auth0 base domain

### LINKED ISSUE #78: Fix redirect_uri in Okta integration
As of now, we set the auth0 baseDomain to be `auth0.astronomer.io` by default. This causes the universal oauth `redirect_uri` to be `https://redirect.astronomer.io`, even though that should only be true if the user is leveraging the Auth0 integration.

The `redirect_uri` for the Okta integration should be the customer's default URL. I'm going to make a new `useGlobalRedirect` config that causes the `redirect_uri` to be set to the correct URL if the config is false.

ashb:
We should use the customer's houston callback directly for other Auth0
clients or for other OIDC providers.

Closes #78 

@petedejoy I did this slightly differently than in the issue report, but I think it should have the desired effect.

### DIFF
--- a/src/lib/oauth/config/index.js
+++ b/src/lib/oauth/config/index.js
@@ -23,30 +23,6 @@ export function oauthUrl() {
 }
 
 /*
- * Return full oauth redirect url.
- * @return {String} The oauth redirect url.
- */
-export function oauthRedirectUrl() {
-  const isProd = process.env.NODE_ENV === "production";
-  const { discoveryUrl: auth0Url } = providerCfg.auth0;
-  const defaultAuth0 = auth0Url === "https://astronomerio.auth0.com";
-
-  // If we're in prod with the default auth0 domain configured,
-  // return the shared redirect url. This is for the shared
-  // auth0 account that is enabled by default. Auth0 requires a list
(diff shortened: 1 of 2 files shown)


src/lib/oauth/config/index.js
```diff
@@ -146,6 +122,32 @@ function subclassClient(issuer, clientId, integration, providerName) {
...
+        // at a custom domain, the redirect won't work. To get around this,
+        // we host a well-known url to use by default.
+        if (isProd && defaultAuth0) {
+          return "https://redirect.astronomer.io";
+        }
+      }
+
+      // Otherwise return the redirect url of the installation. If a user, brings
+      // their own auth0 account, this will be used and will skip the shared url.
+      return `${houston()}/${version()}/oauth/redirect`;
```
petedejoy: So this `$houston()` just ends up being the basedomain of the Astro deployment right?

schnie: Should be `houston.baseDomain`

ashb:
@petedejoy Yeah, correct - Okta will by default use `https://houston.$basedomain/oauth/redirect` as the redirect_url

PR #94, opened 2019-07-19
Set role for Service Accounts

samblackk:
Attached to https://github.com/astronomer/astronomer/issues/364

This may need additional updates - I'm running into a `You don't have permission` GraphQL error when attempting to update the Service Worker. @ashb any insights here?


src/resolvers/mutation/create-service-account/index.js
```diff
@@ -1,5 +1,5 @@
 import fragment from "./fragment";
-import { checkPermission, serviceAccountRoleMappings } from "rbac";
```
ashb: Is `serviceAccountRoleMappings` used anywhere else now, or should we delete it?

src/resolvers/mutation/update-service-account/index.js
```diff
@@ -23,7 +24,22 @@ export default function updateServiceAccount(parent, args, ctx, info) {
...
   const where = { id: args.serviceAccountUuid };
+
+  // Get role bindings
+  const roleBindings = await ctx.db.query.roleBindings(
+    { where: { serviceAccount: where } },
+    "{ id }"
+  );
+
+  // Update the rolebinding if included in payload
+  if (data.role) {
```
ashb: I wonder if we could do this all via something like https://www.prisma.io/docs/prisma-client/basic-data-access/writing-data-JAVASCRIPT-rsc6/#nested-object-writes

samblackk: I'm not sure about this one.

src/resolvers/mutation/create-service-account/index.js
samblackk: It's used elsewhere - let's leave it.

schnie:
@samblackk you may have figured this out already but this is one of the currently weird mutations that we should refactor, but for right now I'm guessing you are hitting that `checkPermissions` and getting your error because your user does not have `${entityType}.serviceAccounts.update` permission. What role is your user?

samblackk:
@schnie I'm a SYS_ADMIN and WORKSPACE_ADMIN

Issue #77, opened 2019-06-25
Fix users query


schnie:
The `users` query is too loose now. Details to come.

petedejoy:
A few notes from our chat:
 - Refactor current users query to be a systemUsers query. Add system admin permissions here.
 - Add workspaceUsers query that requires workspaceUuid as an arg. Add workpsace admin permissions here.

ryw:
How is this going @petedejoy?

petedejoy:
Should have a pair of PRs ready for review today! Got sucked into a million other things the last 2 weeks

petedejoy:
Houston PR: https://github.com/astronomer/houston-api/pull/95
Orbit PR: https://github.com/astronomer/orbit-ui/pull/124

Will work on writing a unit test for my new Houston resolver today as well.

petedejoy:
@schnie - just pushed up the changes we talked about this afternoon. I added a third `workspaceUser` query that takes workspaceUuid and username as required args. It basically just pulls some of the functional stuff we were doing to Houston. I'm pretty stoked about how it turned out- it's much cleaner and aligns super nicely with the "a resolver for everything" GraphQL philosophy. I also have unit tests in there for all three of the new resolvers.

petedejoy:
Closing out here now that the PRs have been merged 🎉

PR #95, opened 2019-07-24
Fix users query

petedejoy:
This PR addresses the following issue: https://github.com/astronomer/houston-api/issues/77

I separated out our old, non-permission bound `users` query into two new `workspaceUsers` and `systemUsers` queries. The former is now what we use in orbit to get a user's information and requires `workspaceUuid` as an arg, meaning you can't get a user's information without a workspace Id. The `systemUsers` query is the same as our old `users` query, but with an added `system.users.list` permission.


src/resolvers/query/workspace-users/index.js
```diff
@@ -0,0 +1,21 @@
...
+export default async function workspaceUsers(parent, args, ctx, info) {
+  return ctx.db.query.users(
+    {
+      where: {
+        roleBindings_some: {
+          workspace: { id: args.workspaceUuid }
+        }
+      }
+    },
+    info ? addFragmentToInfo(info, userFragment) : info
```
schnie: @petedejoy whats going on with the ternary here? Do we need this or should it just be the `addFragmentToInfo(info, userFragment)`?

petedejoy: Good question- to be honest, I stole this chunk from the `workspaces` query. I think you're right- it should just be `addFragmentToInfo`

schnie:
@petedejoy looks like we got some merge conflicts.

@andriisoldatenko do we need to adjust anything in the CLI for this change?

petedejoy:
@schnie Just rebased up- pretty sure it was my permissions PR that we merged in yesterday :).

Also changed the permission on the new resolver to match the updated array structure.

andriisoldatenko:
@schnie on it.

andriisoldatenko:
PR https://github.com/astronomer/astro-cli/pull/259

src/schema.graphql
```diff
@@ -243,12 +243,21 @@ type Metric {
...
-  users(
+ systemUsers(
     userUuid: Uuid
     username: String
     email: String
+  ): [User] @auth(permissions: ["system.users.list"])
+
+  workspaceUsers(
+    workspaceUuid: Uuid!
   ): [User] @auth
```
schnie: @petedejoy should we add some permissions here now that we have it broken apart? Was hoping that this refactor would let us slap some permissions onto the new workspace-level queries.

petedejoy: Hm- I’m not sure. Do you foresee us limiting which roles are able to list workspace users (ie a viewer can’t see other users in the workspace)? It’s fairly well locked down asking for that workspaceUuid as a required arg as far as I can tell.

petedejoy: Hm- I’m not sure. Do you foresee us limiting which roles are able to list workspace users (ie a viewer can’t see other users in the workspace)? It’s fairly well locked down asking for that workspaceUuid as a required arg as far as I can tell.

src/resolvers/query/workspace-user/index.js
```diff
@@ -0,0 +1,25 @@
...
+import { addFragmentToInfo } from "graphql-binding";
+import { find } from "lodash";
+/*
+ * Get a single user on a workspace.
+ * @param {Object} parent The result of the parent resolver.
+ * @param {Object} args The graphql arguments.
+ * @param {Object} ctx The graphql context.
+ * @return users on this specific workspace
+ */
+export default async function workspaceUser(parent, args, ctx, info) {
```
ashb: Do we need a separate query here, or would we instead use

```
query {
    workspace($workspaceUuid) {
        users { username id }
    }
}
```

sort of thing from Orbit?

src/resolvers/query/workspace-user/index.unit.test.js
```diff
@@ -0,0 +1,48 @@
...
+    // Create vars.
+    const vars = {
+      workspaceId: casual.uuid,
+      username: casual.word
+    };
+
+    // Run the graphql mutation.
+    const res = await graphql(schema, query, null, { db }, vars);
+    expect(res.errors).toBeUndefined();
+    expect(db.query.users.mock.calls).toHaveLength(1);
```
ashb:
```suggestion
    expect(db.query.users).toHaveBeenCalled();
```

src/resolvers/query/workspace-user/index.js
```diff
@@ -0,0 +1,25 @@
...
+  const workspaceUsers = await ctx.db.query.users(
+    {
+      where: {
+        roleBindings_some: {
+          workspace: { id: args.workspaceUuid }
+        }
```
ashb:
```suggestion
        },
        username: args.username
```

Then we don't need to filter the user in JS but can let the server handle it. (I think that works)

src/resolvers/query/system-users/index.unit.test.js
```diff
@@ -13,12 +13,12 @@ const schema = makeExecutableSchema({
 
 // Define our mutation
 const query = `
-  query users(
+  query systemUsers(
```
ashb: I'm not sure what we decided with our naming convention, but my first though would be to leave this as `users` (and have `invites` as the companion top level query)

src/schema.graphql
ashb: We should filter it by _something_ otherwise any auth'd user can ask for the users of a workspace they don't belong to if they know the workspace ID (not a disaster, but not great)

 We have `workspace.invites.get` on viewer already - maybe we add a new `workspace.users.get` (`.list`?)?

src/resolvers/query/workspace-user/index.js
ashb: Maaaybe. That might be tricky to actually make work

petedejoy: This is actually how we had it to begin with- chatted with @schnie and it felt a little weird to return a list of users and find the right one every time we displayed a user config screen in Orbit.

src/resolvers/query/system-users/index.unit.test.js
schnie: Yea I've been debating the `system` prefix on these too. I think that makes sense!

ashb: Cool, let's change the global ones to be just `users` and `invites` then.

schnie: Looks like we already use `invites`. That should probably be switched to `workspaceInvites`, and use `invites` for the system level one.

ashb:
Okay so we have `workspaceInvites` that takes an email address. Should we split that out in to two `workspaceInvite` and `workspaceInvites` queries?

And we don't have any global-level `query { invite($email) }` query. We probably need that for providing the data when we click on an invite from the system level screens.

petedejoy:
Diving back in here- looks like @ashb covered all of the above changes with his commit. Is there anything else that needs to be done here?

petedejoy:
Pushed up a couple of fixes here- tested locally with the Orbit counterpart and all looks good!

Issue #100, opened 2019-08-05
Pull deploymentAddAlert mutation from updateDeployment resolver


petedejoy:
In reference to https://github.com/astronomer/astronomer/issues/394#issuecomment-518333398

We need to have a separate mutation for adding an email for alerts to a deployment. Otherwise, an error is thrown if the workspace is on a trial since updating deployment configs is not allowed for trial users.

schnie:
Orbit and Houston PRs merged!

PR #101, opened 2019-08-06
Feature/deployment update alerts

petedejoy:
Backend work for https://github.com/astronomer/houston-api/issues/100

Pulls out the logic that updates alert emails on a deployment into a separate resolver.


src/schema.graphql
```diff
@@ -370,6 +370,12 @@ type Mutation {
     sync: Boolean
   ): Deployment @auth(permissions: ["deployment.config.update"])
 
+
+  deploymentUpdateAlerts(
+    deploymentUuid: Uuid!
+    payload: JSON!
```
schnie: @petedejoy can we drop the `payload` field here and use something more strongly typed, maybe something like `[EmailAddress!]` or maybe just `[String!]` instead.

Trying to drop all usage of the `JSON` type from our schema, in favor of more strongly typed options. This should make the resolver a little simpler too.

petedejoy: Yeah I definitely agree, but this was the quickest way to get this fix out if we need it for cloud 2- all of the orbit stuff was set up to pass it as a json payload and its a little fragile. I tried to set it up to pass an array of emails as an arg and ran into a bunch of issues, but can definitely keep hacking on it.

schnie: An array of Strings would probably be good enough for now. I think we should go ahead and fix it while we're at it, since we're going to be fixing all these JSONs up before 1.0 anyway.

petedejoy: Sweet, all cleaned up on this on and the orbit PR! ✅

petedejoy:
Adjusted naming convention to match `scopeObjectVerb` here.

PR #105, opened 2019-08-14
Fix ServiceAccount bugs to do with being able to edit the roles

ashb:
- Validate the role type matches the entity type, otherwise odd things happen
- Update the role of the ServiceAccount when asked to (we didn't _actually_ change the role)
- Correctly set entityType field on ServiceAccount now roles can be set (it was null if a role other than Editor was chosen at create time)


ashb:
astronomer/orbit-ui#138 relates to this fix too.

ryw:
is this the last item before next RC @ashb?

PR #106, opened 2019-08-14
Turn async exceptions in express route handlers into 500s

ashb:
Without this if the handlers raise an async error the request will never
get a 500 but simply timeout after 30s and an UnhandledPromiseException
message logged in the server.


schnie:
@ashb should we wrap every route with this?

schnie:
@ashb modified to wrap all of our express routes, leaving the graphql route as is. Let me know if you see anything, but going to merge in now and kick off a build for v0.10.0.

ashb:
Post-hoc review: LGTM.

Issue #102, opened 2019-08-14
Workspaces query label arg doesn't work as expected


petedejoy:
When running the `workspaces` query, I'd expect it to return a single workspace given an arg (it takes `userUuid`, `workspaceUuid`, and `label`). However, if you run the query with just a specific `label` as an arg, it returns an array of all workspaces on a cluster.

@schnie we can set up some time to talk through this and figure out what the best way to tighten this up is.

petedejoy:
Update: This is also an issue with the `userUuid` arg. The resolver only uses `workspaceUuid`, so passing any string with `label` or `userUuid` results in all workspaces being returned. As far as I can tell, there are 2 ways to go about fixing this:

1. Turn this into a singular `workspace` query that takes `workspaceUuid` and/or `label` as args and returns a single workspace object.
2. Keep it as a `workspaces` query that takes `userUuid` as an arg and returns the array of workspaces a user has access to. Need to figure out how to handle permissions if we take this route- even if you're an admin on one workspace, you shouldn't be able to get the info for a workspace where you're a viewer.
3.  Some combination of the two solutions above.

PR #109, opened 2019-08-16
Clean up workspaces query

petedejoy:
This PR addresses https://github.com/astronomer/houston-api/issues/102.

We stripped the args from `workspaces`, as they didn't serve much of a purpose previously. Now, any authed user can call the query and return the list of workspaces they're a member of.

We also went ahead and removed the concept of `Groups` from the codebase, as we don't use them anywhere in the application.


schnie:
@andriisoldatenko is there anything we need to adjust in the CLI for this?

schnie:
Looks like we will prob need to rip out the args from here: https://github.com/astronomer/astro-cli/blob/master/houston/queries.go#L213-L218.

petedejoy:
@schnie Good catch! Should be fixed now.

PR #110, opened 2019-08-19
Fix "users" return & add search to invites

samblackk:
Attached to https://github.com/astronomer/astronomer/issues/426

Adds search to invites and allows all users to return on the users query. 

Please r/m

### DIFF
--- a/src/resolvers/query/users/index.js
+++ b/src/resolvers/query/users/index.js
@@ -22,7 +22,7 @@ export default async function users(parent, args, ctx, info) {
  * @param {Object} ctx The graphql context.
  * @return {Object} The users query.
  */
-export function usersQuery(args, ctx) {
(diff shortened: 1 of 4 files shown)


src/resolvers/query/users/index.js
```diff
@@ -22,20 +22,22 @@ export default async function users(parent, args, ctx, info) {
...
+  if (username) return { where: { username_contains: username } };
 
   // If we have an email use it.
   if (address)
-    return { where: { emails_some: { address: address.toLowerCase() } } };
+    return {
+      where: { emails_some: { address: address.toLowerCase() } }
+    };
 
   // Otherwise just return a query for the current user.
```
schnie: @samblackk guess this comment doesn't apply anymore

PR #112, opened 2019-08-19
System deployment update & delete

samblackk:
Attached to https://github.com/astronomer/astronomer/issues/316

Grants system admin permission to view all deployments on the platform as well as make updates to the deployment config.


schnie:
Looks good, but I'm wondering if we should work on this `deployments` query, breaking it up into multiple queries with corresponding permissions, rather than expanding on the weird logic we already had in there for permissions. @petedejoy did something similar with `workspaces` recently.

What you thinking @samblackk?

samblackk:
@schnie We can clean it up now. 

Essentially we would need:
1. A `deployment` query with a `userUuid` and `deploymentUuid` arg that returns a single deployment if the user has permission to access. 
2. Update `deployments` query to only handle returning a list of deployments that are attached to the user.
3. A `systemDeployments` query with a `userUuid` arg that returns all the deployments on the platform if the user has the correct permissions to pull them. 

Sound about right?

petedejoy:
If we're sticking to the syntax used for the workspace queries, I believe it'd be `userDeployment`, `userDeployments`, and `deployments` (no scope for system-level queries`) in that order. Otherwise, that sounds right on. Happy to help out if there's anything I can do!

schnie:
Should it be `workspaceDeployments` or `userDeployments`? Don't we only want a list of deployments for a workspace? Is there somewhere we query for the deployments for all workspaces?

samblackk:
[SysAdmin: View + Delete Deployments](https://trello.com/c/21QAubXl/10-sysadmin-view-delete-deployments)

src/schema.graphql
```diff
@@ -249,9 +249,10 @@ type Query {
...
   ): DeploymentConfig @auth
 
-  deployments(
+  deployments: [Deployment] @auth(permissions:["system.deployment.list"])
+
+  workspaceDeployments(
     workspaceUuid: Uuid
-    deploymentUuid: Uuid # Ignored, only here for backward compatibility.
     releaseName: String
   ): [Deployment] @auth # Permissions checked in resolver, due to limitations with current API design.
```
schnie: @samblackk I think we can remove this old comment and pass `workspace.deployments.list` into the `@auth` resolver now right?

src/resolvers/query/workspace-deployments/index.js
```diff
@@ -0,0 +1,92 @@
...
+ * @return {[]Deployment} List of Deployments.
+ */
+export default async function workspaceDeployments(parent, args, ctx, info) {
+  // Pull out some args.
+  const { releaseName, workspaceUuid } = args;
+  const admin = hasSystemPermission(ctx.user, "system.deployment.update");
+
+  // Perform extra checks here if passing in releaseName,
+  // because it passes right through the auth directive.
+  if (workspaceUuid && !admin) {
```
schnie: @samblackk we can use the `@auth` resolver and cut out the `checkPermission` calls here right?

src/schema.graphql
```diff
@@ -249,11 +249,12 @@ type Query {
...
 
-  deployments(
+  deployments: [Deployment] @auth(permissions:["system.deployment.list"])
+
+  workspaceDeployments(
     workspaceUuid: Uuid
-    deploymentUuid: Uuid # Ignored, only here for backward compatibility.
     releaseName: String
-  ): [Deployment] @auth # Permissions checked in resolver, due to limitations with current API design.
+  ): [Deployment] @auth(permissions: ["deployment.config.update", "system.deployment.update"] op: OR)
```
```suggestion
  ): [Deployment] @auth(permissions: ["workspace.deployments.list"])
```

schnie: @samblackk whacha think here?

samblackk: @schnie We need to keep the system one as well. Or the admin panel won't list all of the deployments on the platform. can be updated to `system.deployments.list`

schnie: @samblackk I thought the system stuff used the `deployments` query up above.

samblackk:
@schnie Good to go.

src/resolvers/query/workspace-deployments/index.js
```diff
@@ -0,0 +1,61 @@
...
+ * @return {Object} The deployments query.
+ */
+export function deploymentsQuery(args, ctx) {
+  // Pull out some args.
+  const { deploymentUuid, releaseName, workspaceUuid } = args;
+
+  // Build query structure.
+  let query = { where: { AND: [] } };
+
+  // If we have deploymentId, add to filter.
```
schnie: @samblackk we dropped `deploymentId` from the args in the schema def. Seems like we can remove this chunk. What you think?

src/resolvers/query/workspace-deployments/index.js
```diff
@@ -0,0 +1,61 @@
...
+  if (releaseName) {
+    query.where.AND.push({ releaseName });
+  }
+
+  // If we have workspaceUuid, add to filter.
+  if (workspaceUuid) {
+    query.where.AND.push({ workspace: { id: workspaceUuid } });
+  }
+
+  if (!(deploymentUuid || releaseName || workspaceUuid)) {
```
schnie: and here I guess

src/resolvers/query/workspace-deployments/index.unit.test.js
```diff
@@ -0,0 +1,136 @@
...
+
+    // Run the graphql mutation.
+    const res = await graphql(schema, query, null, { db, user });
+    expect(res.errors).toBeUndefined();
+    expect(deployments.mock.calls.length).toBe(1);
+  });
+});
+
+describe("deploymentsQuery", () => {
+  test("deploymentId is correctly used in query", async () => {
```
schnie: and this test?

samblackk:
@schnie Cleaned up!

PR #113, opened 2019-08-19
Allow createDeployment mutation to set labels on Airflow k8s namespaces

sjmiller609:
Pre-integrated with this PR https://github.com/astronomer/houston-api/pull/108 , so it includes duplicate code. merge that one first, then I can do a pull / push to remove the duplicate diff from this PR.

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -299,7 +299,7 @@ deployments:
   logHelmValues: false
 
   # Default airflow version to deploy. This is used as the default image tag.
-  defaultAirflowVersion: "1.10.3"
+  defaultAirflowVersion: "1.10.4"
 
   # This is the database connection that we use to connect
   # and create other databases/schemas/users for each
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -109,7 +109,12 @@ repository: "https://helm.astronomer.io/"
 helm:
   baseDomain: ~
   registryAuthSecret: ~
+  registryPort: 5000
   releaseName: ~
+  # Map of labels to apply to an Airflow deployment's namespace.
+  # Does not apply in the case of single-namespace mode.
+  deploymentNamespaceLabels:
+    test-label: yep
   releaseNamespace: ~
   releaseVersion: ~
   singleNamespace: false
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -299,9 +299,8 @@ roles:
 #
 deployments:
 
+  # Labels specifed here are applied to Airflow deployment namespaces.
   namespaceLabels: {}
-
-  registryPort: 5000
 
   # Log out final helm values whenever they are generated, before deployment.
   logHelmValues: false
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -109,9 +109,7 @@ repository: "https://helm.astronomer.io/"
 helm:
   baseDomain: ~
   registryAuthSecret: ~
-  registryPort: 5000
   releaseName: ~
-  deploymentNamespaceLabels: {}
   releaseNamespace: ~
   releaseVersion: ~
   singleNamespace: false
@@ -297,6 +295,11 @@ roles:
 # Airflow Deployment Configurations
 #
 deployments:
+
+  namespaceLabels: {}
+
+  registryPort: 5000
+
   # Log out final helm values whenever they are generated, before deployment.
   logHelmValues: false
 
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -111,10 +111,7 @@ helm:
   registryAuthSecret: ~
   registryPort: 5000
   releaseName: ~
-  # Map of labels to apply to an Airflow deployment's namespace.
-  # Does not apply in the case of single-namespace mode.
-  deploymentNamespaceLabels:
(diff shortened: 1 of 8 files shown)


config/default.yaml
```diff
@@ -109,7 +109,9 @@ repository: "https://helm.astronomer.io/"
 helm:
   baseDomain: ~
   registryAuthSecret: ~
+  registryPort: 5000
   releaseName: ~
+  deploymentNamespaceLabels: {}
```
schnie: @sjmiller609 similar to the registry port, what do you think about moving `namespaceLabels` under the `deployments` section? That section is dedicated for platform configs that deal with the airflow deployments.

src/lib/deployments/naming/index.js
```diff
@@ -25,6 +25,24 @@ export function generateNamespace(releaseName) {
...
+/*
+ * Return an empty array if single namespace mode,
+ * otherwise return the labels from the config file
+ * @return array of objects in the form {"key": ****, "value": ***}
+ */
+export function generateDeploymentLabels() {
+  const { deploymentNamespaceLabels, singleNamespace } = config.get("helm");
+  // When the labels are serialized, the format we want is an array
+  // of { key: *** , value: *** } objects.
+  var labels = []
```
schnie: @sjmiller609 we might be able to reuse a function for this looping operation. We have https://github.com/astronomer/houston-api/blob/master/src/lib/deployments/config/index.js#L359-L366, which does the same thing it seems. We may want to make the function name more generic though if we use it here too.

schnie:
@sjmiller609 I think this PR contains some of the stuff that we merged in last week. Might need a rebase / resolve conflicts.

src/lib/deployments/config/index.js
```diff
@@ -97,8 +97,7 @@ export function limitRange() {
   const containerLimit = {
     type: "Container",
     default: min,
-    defaultRequest: min,
-    min
```
sjmiller609: I forget what this change is about. tracking it down

src/lib/deployments/naming/index.js
sjmiller609: not sure about this one because we want our objects like this:
```
{ key: ***, value: ***}
```
as opposed to
```
{ ***, ***}
```

src/lib/deployments/config/index.js
sjmiller609: With change:
```
Resource Limits
 Type       Resource  Min  Max      Default Request  Default Limit  Max Limit/Request Ratio
 ----       --------  ---  ---      ---------------  -------------  -----------------------
 Pod        cpu       -    3500m    -                -              -
 Pod        memory    -    13440Mi  -                -              -
 Container  cpu       -    -        100m             100m           -
 Container  memory    -    -        384Mi            384Mi          -
```

Without change:
```
Resource Limits
 Type       Resource  Min    Max      Default Request  Default Limit  Max Limit/Request Ratio
 ----       --------  ---    ---      ---------------  -------------  -----------------------
 Pod        cpu       -      3500m    -                -              -
 Pod        memory    -      13440Mi  -                -              -
 Container  cpu       100m   -        100m             100m           -
 Container  memory    384Mi  -        384Mi            384Mi          -
```

I recall now. The istio sidecars will be small, and it isn't necessary to limit the size of containers on the low end. Istio sidecars fail to deploy in the namespace with this limit in place.

config/default.yaml
```diff
@@ -298,6 +298,11 @@ roles:
 # Airflow Deployment Configurations
 #
 deployments:
+
```
schnie: @sjmiller609 can you write up a quick comment here, indicating how these are used downstream?

Also, do we need `registryPort` here or can we reference `registry.port` from the last PR?

src/lib/deployments/config/index.js
```diff
@@ -361,7 +360,7 @@ export function envArrayToObject(arr = []) {
  * @param {Object} An array of objects with key/value pairs.
  * @return {[]Object} The object with key/value pairs.
  */
-export function envObjectToArray(obj = {}) {
+export function objectToArrayOfKeyValue(obj = {}) {
```
schnie: @sjmiller609 this was very loosely related to the `envArrayToObject` function right above it. Since we changed the naming to be more generic here, I'm thinking we should do the same there.

src/lib/deployments/naming/index.js
```diff
@@ -25,6 +26,23 @@ export function generateNamespace(releaseName) {
...
+ * Return an empty array if single namespace mode,
+ * otherwise return the labels from the config file
+ * @return array of objects in the form {"key": ****, "value": ***}
+ */
+export function generateDeploymentLabels() {
+  const { singleNamespace } = config.get("helm");
+  const { namespaceLabels } = config.get("deployments");
+
+  // When the labels are serialized, the format we want is an array
+  // of { key: *** , value: *** } objects.
```
schnie: @sjmiller609 can prob delete this comment since its covered by the function comment.

sjmiller609: will do right now

PR #115, opened 2019-08-21
Add removeUser mutation

samblackk:
Attached to https://github.com/astronomer/astronomer/issues/439 & https://github.com/astronomer/orbit-ui/pull/146

Adds removeUser mutation.
Please r/m


schnie:
@samblackk it looks like this removes the `roleBindings` but not the actual user record. Should we delete the actual user too?

src/resolvers/mutation/index.js
```diff
@@ -35,6 +36,7 @@ export default {
   createSystemRoleBinding,
   createToken,
   createUser,
+  removeUser,
```
schnie: nit but these are alphabetical currently

samblackk: womp. reordering

src/resolvers/mutation/remove-user/index.js
```diff
@@ -0,0 +1,18 @@
...
+ * @param {Object} parent The result of the parent resolver.
+ * @param {Object} args The graphql arguments.
+ * @param {Object} ctx The graphql context.
+ * @return {Workspace} The Workspace.
+ */
+export default async function removeUser(parent, args, ctx) {
+  // Pull out some args, ignoring email - remove later.
+  const { userUuid } = args;
+
+  // Remove the RoleBinding.
```
schnie: nit but the comments on the different chunks are reversed. Also, now that I think about it, I think you can get away just deleting the user and skipping the rolebindings. We have a cascade delete defined so prisma should handle it. https://github.com/astronomer/houston-api/blob/master/database/datamodel.graphql#L20

samblackk:
Updated!

PR #116, opened 2019-08-22
Prevent removing all system admins

schnie:
This PR prevents the deletion of a system admin, if there are no other admins in the system.


schnie:
Hate to do it, but merging to build an RC. Let me know if you see anything.

PR #119, opened 2019-08-30
Email case checking in oauth

samblackk:
Fixes https://github.com/astronomer/astronomer/issues/326

Created a test email account: `SUPERsamBLACKtest@gmail.com` and verified the invited email in all lowercase will match the signup oauth email that's variable in case. 

Please r/m


src/routes/v1/oauth/callback/post.js
```diff
@@ -34,7 +34,9 @@ export default async function(req, res, next) {
   // Grab user data
   // Some IDPs don't return useful info, so fall back to the claims if we don't have it
   const userData = merge(await provider.userinfo(tokenSet.access_token), {
```
ashb: I dont think this PR will do what we need, as the upper case email will come from the return value of `provider.userinfo(tokenSet.access_token)`, at least for Office365.

samblackk: Suggestions to get this to do what we need?

schnie:
Merging this in now to patch up for v0.10.1. We can return to this if we need to.

PR #123, opened 2019-09-10
Validate extraAU

samblackk:
Attached to https://github.com/astronomer/astronomer/issues/469
Please r/m


src/lib/deployments/validate/index.js
```diff
@@ -4,19 +4,35 @@ import validateProperties from "./properties";
...
+  deploymentId,
+  deployment
+) {
   // Validate environment variables.
   validateEnvironment(args.env);
 
+  const { executors } = config.get("deployments");
+
+  // Get the executor on this deployment.
+  const executor = get(deployment, "config.executor", AIRFLOW_EXECUTOR_CELERY);
```
schnie: @samblackk I _think_ we're doing this check / fallback in multiple places now. Do you think it makes sense to add a new constant called like `AIRFLOW_EXECUTOR_DEFAULT` and have that defined as 
```
export const AIRFLOW_EXECUTOR_DEFAULT = AIRFLOW_EXECUTOR_CELERY
```
so when we switch to kubernetes/knative as default we only change it one place?

Then we could replace these fallbacks with the new one.

samblackk: Good idea!

samblackk:
@schnie GTG

src/resolvers/mutation/update-deployment/index.js
```diff
@@ -62,7 +61,12 @@ export default async function updateDeployment(parent, args, ctx, info) {
   });
 
   // Validate our args.
-  await validate(deployment.workspace.id, mungedArgs, args.deploymentUuid);
+  await validate(
```
schnie: @samblackk wonder if we should trim `validate()` down to 3 args or even 2. `deployment.workspace.id`, `args.deploymentUuid` and `deployment` could all be derived from `deployment`. 

Whacha think?

samblackk: Cleaned up in the lates commit

PR #124, opened 2019-09-10
Add Deployment Capabilities

schnie:
Have not tested this yet. Will later on. Just kicking this up to get it started.


schnie:
Just tested, seems to work fine. Cleaned up a few things on the workspace side of the house too.

schnie:
@andriisoldatenko this is merged. You should be able to consume this from the CLI now.

PR #127, opened 2019-09-12
Fix updateDeployment mutation

schnie:
A recent update to the validation logic broke a few things on the `updateDeployment` mutation. When calling this mutation, both `deploymentId` and `workspaceId` were null when passed to the `checkExistence()` function. This caused any update on the form to throw a `DuplicateDeploymentLabelError` no matter what.

This PR ensures both of these ids are passed correctly. It also removes a second query for the `workspace` to check for trial status, and instead asks for a few more workspace details on the first query for the deployment.


samblackk:
@schnie Need to check the react lifecycle `componentDidUpdate()` and ensure we're refetching the correct values if needed. Since the issue corrects itself during refresh, then the issue lies in how the form is updating after submission.

PR #129, opened 2019-09-24
Stripe customer exists without card

petedejoy:
This PR handles the edge case in which a customer exists in Stripe without an associated credit card. This might happen if we sign an annual deal with a Cloud customer and invoice them for the year up-front rather than requiring a credit card for monthly payment.

astronomer/issues#400


src/resolvers/query/card/index.js
```diff
@@ -10,17 +10,17 @@ export default async function card(parent, args) {
   const response = await getStripeCard(stripeCustomerId);
 
   const customerInfo = response;
-  const cardInfo = response.sources.data[0] || {};
+  const cardInfo = response.sources.data[0] || null;
```
schnie:
```suggestion
  const cardInfo = response.sources.data[0];
```
Purely a nitpick style thing but I think we can omit the `|| null`, this will make `cardInfo` `undefined` instead. Both are falsey, so shouldn't matter.

src/resolvers/query/card/index.js
```diff
@@ -10,17 +10,17 @@ export default async function card(parent, args) {
...
+    return {
+      expMonth: cardInfo.exp_month,
+      expYear: cardInfo.exp_year,
+      last4: cardInfo.last4,
+      name: cardInfo.name,
+      brand: cardInfo.brand,
+      billingEmail: customerInfo.email,
+      company: customerInfo.metadata.company
+    };
+  }
```
schnie: Another purely style thing, but could avoid a branch (if statement) and use a ternary as the `return` statement.
```
return cardInfo ? { key: values, go: here } : null;
```

petedejoy: ✅

petedejoy:
@schnie - just committed all the fixes here!

src/resolvers/query/card/index.js
```diff
@@ -10,17 +10,17 @@ export default async function card(parent, args) {
   const response = await getStripeCard(stripeCustomerId);
```
schnie:
```suggestion
  const customerInfo = await getStripeCard(stripeCustomerId);
```

then you can delete the next line and just replace `response` below with `customerInfo`.

PR #131, opened 2019-09-25
Add Stripe Customer mutation

petedejoy:
This PR addresses https://github.com/astronomer/issues/issues/401

System Admins will now be able to add a stripeCustomerId directly to the database from the playground. This will be particularly useful for:
- Migrating old cloud customers to new cloud and using their existing stripeCustomerId so they do not need to re-submit their payment info
- Allowing a user to bypass our trail restrictions without submitting real credit card information to our system (ie. cases where we invoice annually or want to unlock environments for Astronomer employees or partners)


src/schema.graphql
```diff
@@ -467,6 +467,11 @@ type Mutation {
     token: String!
   ): Card @auth(permissions: ["workspace.billing.update"])
 
+  addCustomer(
```
schnie: @petedejoy should we use `createCustomer` here to standardize? Looking at it, thinking we may want to make `addCard` `workspaceAddCard` to standardize our naming, but we can deal with that later. Just thinking out loud.

config/default.yaml
```diff
@@ -208,6 +208,7 @@ roles:
       - system.deployment.logs
       - system.deployment.metrics
       - system.users.list
+      - system.billing.update
```
schnie: Thinking more about this, maybe we go more granular with `system.billing.createCustomer`, or maybe even `system.workspace.createCustomer`. :thinking: SO DIFFICULT

src/schema.graphql
petedejoy: Hm, good point. I'm hesitant to go with `createCustomer` here since we're not really creating a customer anywhere, we're just adding a customer that already exists to the database. We could go with something like `addStripeId` maybe?

schnie: ah, maybe then `system.workspace.addCustomer`? Then it'd mesh with `workspace.billing.addCard` if we change the other one later.

schnie: Good point, maybe `addCustomerId` just to not mention stripe.

config/default.yaml
schnie: `system.workspace.addCustomerId`?

schnie: Trying to standardize with some of the other ones.

schnie:
And maybe a quick unit test or two? Obviously a pretty simple resolver, but maybe a lil exercise couldn't hurt.

schnie:
Looks good!

PR #132, opened 2019-09-26
Catch workspace delete with deployments

samblackk:
Handles catching a situation where user can delete a workspace while it still has deployments.
Please r/m


src/resolvers/mutation/delete-workspace/index.js
```diff
@@ -6,6 +8,16 @@
...
 export default async function deleteWorkspace(parent, args, ctx) {
+  const deployments = ctx.db.query.deployments(
+    {
+      where: { id: args.workspaceUuid }
+    },
+    `{ id }`
+  );
+
+  // Don't delete if there are deployments present
+  if (deployments) throw new WorkspaceDeleteError();
```
schnie: @samblackk if no deployments are found, what is returned here? Empty array?

As an alternative we can also do something like this to get the count - https://github.com/astronomer/houston-api/blob/master/src/lib/users/index.js#L97-L100.

PR #134, opened 2019-09-30
Add /v1/metrics route

sjmiller609:
This PR is a part of this issue: https://github.com/astronomer/issues/issues/27

I followed these instructions to add prometheus to houston: https://www.npmjs.com/package/express-prom-bundle

It ran locally and successfully exposed the metrics using docker-compose up + npm start.

A few points to review:
- "require" vs "import"
- when the prometheus middleware is added: the plugin notes that the order in which it's added to the app matters because only that which is added after will be monitored.
- does package-lock.json also need an update?

Thanks for your patience, I am new to JS/Node/Express.


package.json
```diff
@@ -86,4 +88,4 @@
     "prettier": "^1.15.1",
     "supertest": "^3.3.0"
   }
-}
+}
```
schnie: @sjmiller609 newline here

src/index.js
```diff
@@ -39,6 +39,11 @@ app.use(
   })
 );
 
+// Set up prometheus metrics
+const promBundle = require("express-prom-bundle");
```
schnie:
```suggestion
import promBundle from "express-prom-bundle"
```
Should be able to do something like this. Guessing its not a named export so no curly braces required. Then you can toss this up top. Linter should order it for you.

sjmiller609: yep, that worked. thanks!

schnie:
Wonder if we should push this under `/v1` or leave it up at the root :thinking:

sjmiller609:
> Wonder if we should push this under `/v1` or leave it up at the root 🤔

I kind of like it being default. also, i don't think it's versioned with our stuff.

schnie:
So we have every other REST route defined under `v1` now. It feels weird to have one outlier. I know `/metrics` is the default, but I think its only recommended because that's the default that Prometheus will scrape at if not told otherwise.

If we want to leave `/metrics` at the root, what do you think about moving `/healthz` to root as well. They seem to fall into the same bucket.

sjmiller609:
We will be adding in application-specific metrics to the client, so it does make sense to put it under /v1

sjmiller609:
image

src/index.js
```diff
@@ -39,6 +40,14 @@ app.use(
   })
 );
 
+// Set up prometheus metrics
+const metricsMiddleware = promBundle({
+  includeMethod: true,
+  metricsPath: "/v1/metrics"
+});
+
+app.use(metricsMiddleware);
```
schnie: @sjmiller609 looks good but it may make sense to leave this top level file as-is and move this block here - https://github.com/astronomer/houston-api/blob/master/src/routes/v1/index.js#L19.

I'm not sure, but since we'd be attaching it to a nested router (already under `/v1`), you may even be able to get away with a one-liner in that file, letting it default to `/metrics`, which would really be exposed at `/v1/metrics`.

`router.use(promBundle({ includeMethod: true }));`

sjmiller609: I gave that a shot, but it didn't work.

schnie: Gotcha. Can we still move this block to the `v1` router though?

sjmiller609: I did not figure out how to make it work with router.use, I tried a few different things... seems like it was just not showing up. Do you want me to dig in here to understand the difference b/w app.use and route.use? If you mean use the path /v1/metrics, then that is already the case

schnie: I just committed an update to move it to the `v1` module. Also not able to drop that hard-coded path, so just left it.

PR #135, opened 2019-10-04
Init workspace-aware insights

samblackk:
Attached to https://github.com/astronomer/issues/issues/432 & https://github.com/astronomer/issues/issues/431 

Add's platform context to `generateDeploymentLabels()` - generating`workspaceId` and `platformReleaseName` to `namespaceLabels` when a deployment is created.

Returns a key/value object - 
```
{
  deployment: "label-name",
  workspace: "cjladu123",
  release: "asdfi2d",
  trial: true,
}
```

With this information we can send workspace-aware deployment information to granfa/prom and back into the alert emails. 

Please review & make updates elsewhere as needed.


schnie:
@samblackk nice, looking good. 

What would happen when a user goes and adds a credit card? Seems like we'd need to `helm upgrade` the airflow chart to push out an update on `trial`.

schnie:
@samblackk if the trial thing is an issue, we could merge this now without the trial one, so we can patch up our emails. Then we can follow up and figure out how to deal with the trial label separately.

samblackk:
@schnie Removed the trial tag. Other thing - this label update happens on deployment creation. Is there a way we can update existing deployments to contain the workspace label?

schnie:
These particular labels only apply to the kubernetes namespace object that contains the rest of the deployments/secrets/etc, so we could always go and manually tag a few deployments.

We could also add `namespaceLabels` to the `updateDeployment` request to commander and pass them through here - https://github.com/astronomer/houston-api/blob/d111c9f4dfb10e8eeac67a23412958a4d3e8b4b3/src/resolvers/mutation/update-deployment/index.js#L95, just like `createDeployment`.

We'd need to make some minor changes to commander to support this though.

This may be how we handle the trial thing too. We can call `updateDeployment`, with an updated `trial` flag to update the namespace label.

samblackk:
@schnie Added `namespaceLabels` to `updateDeployment` - and updated the email template url strings as well (instead of throwing them into a separate PR).

sjmiller609:
Hold up for my review. I remember there was an issue around passing the labels in the update, or some reason why I didn't do that in the first place

src/resolvers/mutation/update-deployment/index.js
```diff
@@ -93,6 +102,7 @@ export default async function updateDeployment(parent, args, ctx, info) {
         name: DEPLOYMENT_AIRFLOW,
         version: updatedDeployment.version
       },
+      namespaceLabels: generateDeploymentLabels(platform),
```
sjmiller609: namespace labels are not a part of the updateDeployment protocol buffer. Neither is namespace. https://github.com/astronomer/commander/blob/faf453bc898a9e2efca4592f82805c5b236783b8/_proto/deployment.proto#L33

In addition, namespaces are not handled in the underlying function. https://github.com/astronomer/commander/blob/faf453bc898a9e2efca4592f82805c5b236783b8/provisioner/kubernetes/kubernetes.go#L84

schnie:
@samblackk since we need to wait until we add support to `updateDeployment` in Commander, what do you think about cutting the changes to `updateDeployment` out here? We can follow up with those changes when we have that done.

The changes to the email templates should be enough to get this working.

PR #136, opened 2019-10-07
Make Deployment Admin be an admin inside Airflow too

ashb:
This lets admins configure the permissions that the roles have in
Airflow (manually, using the UI, not via any automated means)


schnie:
@ashb so is `admin` higher than `op`?

ashb:
@schnie Yeah, admin can to absolutely everything in Airflow, Op can do most thing, but critically not manage role/permission bindings.

PR #137, opened 2019-10-09
Refactor ServiceAccount Queries & Mutations

samblackk:
Attached to https://github.com/astronomer/issues/issues/349
Refactors and delegates scoped Service Account queries and mutations. Leaves the old `ServiceAccounts` query and `createServiceAccount` intact until all platforms are migrated to the new schema.


samblackk:
@schnie Totally! Just want to get over this first hurtle with getting the tests to pass. It's weird I'm getting null back for the new items in the schema.

samblackk:
Screen Shot 2019-10-09 at 12 58 58 PM
Screen Shot 2019-10-09 at 1 02 02 PM

schnie:
@samblackk oh you may need to add those files to `mutations` and `queries` `index.js` files - https://github.com/astronomer/houston-api/blob/f280115ff3e801172240d42c18d7fc262cbbf240/src/resolvers/query/index.js.

samblackk:
@schnie @ashb ready to be tested thoroughly!

schnie:
Just to confirm, this should work with the CLI and Orbit, as is right?

samblackk:
Yes! The original query and mutation are still there and unchanged.

src/schema.graphql
```diff
@@ -204,7 +204,9 @@ type ServiceAccount {
   label: String
   category: String
   entityType: String!
-  entityUuid: Uuid!
+  entityUuid: Uuid @deprecated # backwards compatibility
```
schnie: @samblackk so once we get the clients updated, can we drop `entityType` and `entityUuid`?

schnie:
@samblackk do you not think we should use `x.x.list` on the queries where we pull multiple `serviceAccounts`?

Even if we group them both together role-wise, I like having the option to override, just in case.

src/schema.graphql
samblackk: Yeah - `entityType` and `entityUuid` will be deprecated entirely.

schnie: should we @depricated the `entityType` here too?

samblackk:
@schnie I'm not sure it makes a different - a get is a get. And the directive if it's multiple or not is in the query itself `workspaceServiceAccount` vs `workspaceServiceAccounts`

schnie:
Maybe its fine for now, I'm just thinking that the _ability_ to list and enumerate all the `serviceAccounts`, `deployments`, `workspaces`, could be separate from getting a single ones details. Thinking that someday we may only expose certain fields through the `list` operation, and show more detail or something with the single query. This would mean that `get`ing a certain object may be reserved for higher up roles.

Also thinking it'd give enterprise users the ability to tweak that behavior by overriding the houston configs for the roles.

src/schema.graphql
```diff
@@ -203,8 +203,10 @@ type ServiceAccount {
   apiKey: String
   label: String
   category: String
-  entityType: String!
-  entityUuid: Uuid!
+  entityType: String! @deprecated # backwards capability
+  entityUuid: Uuid @deprecated # backwards capability
+  workspaceUuid: Uuid @alias(field: "entityUuid")
+  deploymentUuid: Uuid @alias(field: "entityUuid")
```
ashb: This works and you can have two fields aliased to the same thing?

src/resolvers/query/workspace-service-account/index.unit.test.js
```diff
@@ -0,0 +1,95 @@
...
+    };
+
+    const vars = {
+      serviceAccountUuid: casual.uuid
+    };
+
+    // Run the graphql mutation.
+    const res = await graphql(schema, query, null, { db, user }, vars);
+    expect(res.errors).toBeUndefined();
+    expect(workspaceServiceAccount.mock.calls.length).toBe(1);
```
ashb: I prefer the slightly more "literate" version:

```suggestion
    expect(workspaceServiceAccount).toHaveBeenCalledTimes(1);
```

(or `.toHaveBeenCalled()` if we don't care about how many times.)

ashb: It may also be worth asserting that we call this mutation with using `expect(workspaceServiceAccount).toHaveBeenCalledWith` -- see https://github.com/astronomer/houston-api/blob/58b30f8c0e16d48e437c5a3fd6a58fee9dcc4a96/src/resolvers/mutation/update-service-account/index.unit.test.js#L74-L83 for an example

src/resolvers/query/deployment-service-accounts/index.js
```diff
@@ -0,0 +1,37 @@
+import fragment from "./fragment";
+import { addFragmentToInfo } from "graphql-binding";
+
+/*
+ * Get a list of deployment service accounts
+ * @param {Object} parent The result of the parent resolver.
+ * @param {Object} args The graphql arguments.
+ * @param {Object} ctx The graphql context.
+ * @return {AuthConfig} The auth config.
```
ashb:
```suggestion
 * @return {[ServiceAccount]} The service accounts.
```

ashb: Not sure if that is the right syntax for list-of in doc comment

src/resolvers/query/deployment-service-accounts/fragment.js
```diff
@@ -0,0 +1,8 @@
+export default `{
+  id
+  roleBinding {
+    role
+    workspace { id }
```
ashb: Do we need workspace here in the return? Would it perhaps make more sense to have it be `deployment { id workspace { id } }`?

src/resolvers/mutation/create-workspace-service-account/fragment.js
```diff
@@ -0,0 +1,9 @@
+export default `fragment EnsureFields on ServiceAccount {
+  id
+  roleBinding {
+    role
+    workspace { id }
+    deployment { id }
```
ashb: This is always going to be empty isn't it?

samblackk: Ah, yes. I'll clean these up.

schnie:
@samblackk was thinking we'd separate `deleteServiceAccount` and `updateServiceAccount` into `deleteWorkspaceServiceAccount`, `updateWorkspaceServiceAccount`, `deleteDeploymentServiceAccount`, `updateDeploymentServiceAccount`, each of these taking a `serviceAccountUuid` as a param and either a `workspaceUuuid` _or_ `deploymentUuid` purely for the `@auth` directive, and make both args _required_ for all of these. Whacha think?

src/schema.graphql
schnie: We're going to follow up and fix orbit and the CLI immediately so ideally we can drop these `entity...` fields with that.

schnie:
Thanks @samblackk!

PR #139, opened 2019-10-15
Add verifyEmail mutation

schnie:
This PR adds a `verifyEmail` mutation. This can be used as a system admin override in case email links get stripped or some other issue.


petedejoy:
Looks good to me. Thanks for jumping on this @schnie!

PR #141, opened 2019-10-16
When "accepting" an invite, accept them all

ashb:
Right now there's an edge case when being invited to multiple workspaces:

1. Sent User an invite to Workspace1 (before they made an account)
2. Sent User an invite to Workspace2 shortly thereafter (also before they made an account)
3. User accepts invite to Workspace1, gets access to only one workspace, can't accept invite to Ws2

I have changed it so that when accepting an invite it also joins all other pending workspaces/invites for that user. This makes it the same behaviour as if the user already existed in the system (When adding a user to a workspace when the user already has an account, that user is added directly, they don't need any extra confirmation step.)

Closes https://github.com/astronomer/issues/issues/474


paolaperaza:
Nice @ashb  - I think this covers the issue! I just went back through the list of invites I've sent and confirmed that the only users blocked from Workspace access when accepting invites were those who I invited to more than 1 Workspace, but I realized that everyone else got through as expected.

Thank you!

PR #144, opened 2019-10-16
Add netlify to CORS host for new orbit dev

samblackk:
Add's host for the netlify build of Orbit so we can preview the app without running it locally.
(netlify uses staging env of houston) https://orbit-dev.netlify.com/


schnie:
@samblackk think we could make a new config like

```
cors:
  allowedOrigins:
    - my-first-host
    - another-host
```
and append those if defined, instead of hard-coding one in here?

samblackk:
@schnie Good idea. Update incoming.

schnie:
@samblackk got anything for this? Cleaning up my pull reminders :)

samblackk:
@schnie Updated.

config/default.yaml
```diff
@@ -6,6 +6,13 @@ webserver:
   endpoint: "/v1"
   subscriptions: "/ws"
 
+# CORS Access
+cors:
+  allowedOrigins:
+    - http://app.local.astronomer.io:5000
+    - http://app.local.astronomer.io:8080
+    - http://orbit-dev.netlify.com
```
schnie: @samblackk trying to keep the one off netlify domain out of the codebase, so we can add it up at the terraform level to override here. What do you think about moving the two default ones back into the code and having this be a blank array for now? Then @sjmiller609 can add the netlify host to our staging terraform config.

samblackk: Updated! Makes sense.

PR #145, opened 2019-10-17
Add plaintext link into invite email

schnie:
This PR adds a plaintext link into our user invite email. This is to _hopefully_ avoid the stripped links that Outlook produces.


schnie:
Closes https://github.com/astronomer/houston-api/pull/145

schnie:
@andrewhharmon confirmed that the plaintext link does make it through. Merging!

PR #146, opened 2019-10-18
Allow sysadmins access to Airflow deployments

ashb:
THe authorization handler has been updated to understand all the
`sysadmin.airflow.*` permissions, but for now I have only given the
sysadmin role the "get"/read-only permission.

We may find that this Viewer role in Airflow is missing some permissions
on the Airflow side so we might have to up this, or add missing
permissions to Airflow in our SecurityManager


src/routes/v1/authorization/handler.js
```diff
@@ -65,13 +65,25 @@ export function airflowJWT(user, roles, hostname) {
 
 export function mapLocalRolesToAirflow(user, deploymentId) {
   const entityType = ENTITY_DEPLOYMENT.toLowerCase();
-  if (hasPermission(user, "deployment.airflow.admin", entityType, deploymentId))
+  if (
+    hasPermission(user, "deployment.airflow.admin", entityType, deploymentId) ||
+    hasPermission("sysadmin.airflow.admin")
```
schnie: @ashb don't we need to pass `user` in as the first arg on these new `hasPermission()` calls? Also, shouldn't they be `system.airfow.get`, rather than `sysadmin.airflow.admin`?

ashb: Oh No

ashb: > shouldn't they be system.airfow.get, rather than sysadmin.airflow.admin

I don't think so, not here. This would let an enterprise customer configure what roles the SysAdmin has, but by default it's read only.

ashb: That's what I get for not having any tests.

src/routes/v1/authorization/handler.js
```diff
@@ -65,13 +65,25 @@ export function airflowJWT(user, roles, hostname) {
...
-  if (hasPermission(user, "deployment.airflow.op", entityType, deploymentId))
+  if (
+    hasPermission(user, "deployment.airflow.op", entityType, deploymentId) ||
+    hasPermission(user, "sysadmin.airflow.op")
+  )
     return ["Op"];
-  if (hasPermission(user, "deployment.airflow.user", entityType, deploymentId))
+  if (
+    hasPermission(user, "deployment.airflow.user", entityType, deploymentId) ||
+    hasPermission(user, "sysadmin.ariflow.user")
```
schnie:
```suggestion
    hasPermission(user, "sysadmin.airflow.user")
```

src/routes/v1/authorization/handler.js
```diff
@@ -65,13 +65,25 @@ export function airflowJWT(user, roles, hostname) {
 
 export function mapLocalRolesToAirflow(user, deploymentId) {
   const entityType = ENTITY_DEPLOYMENT.toLowerCase();
-  if (hasPermission(user, "deployment.airflow.admin", entityType, deploymentId))
+  if (
+    hasPermission(user, "deployment.airflow.admin", entityType, deploymentId) ||
+    hasPermission(user, "sysadmin.airflow.admin")
```
schnie: @ashb `sysadmin` should be `system` on all these right?

config/default.yaml
```diff
@@ -214,6 +214,7 @@ roles:
       - system.workspace.addCustomerId
       - system.workspace.suspend
       - system.workspace.extendTrial
+      - system.airflow.get
```
schnie: @ashb won't we need to add the other `system.airflow.*` permissions up into these roles?

src/routes/v1/authorization/handler.js
ashb: Yuup :( Speedy PRs never are.

config/default.yaml
ashb: The ticket calls for Read-only, so no?

src/routes/v1/authorization/handler.js
ashb:
```suggestion
    hasPermission(user, "system.airflow.admin")
```

src/routes/v1/authorization/handler.js
```diff
@@ -65,13 +65,25 @@ export function airflowJWT(user, roles, hostname) {
...
+  )
     return ["Admin"];
-  if (hasPermission(user, "deployment.airflow.op", entityType, deploymentId))
+  if (
+    hasPermission(user, "deployment.airflow.op", entityType, deploymentId) ||
+    hasPermission(user, "sysadmin.airflow.op")
```
ashb:
```suggestion
    hasPermission(user, "system.airflow.op")
```

src/routes/v1/authorization/handler.js
```diff
@@ -65,13 +65,25 @@ export function airflowJWT(user, roles, hostname) {
...
+  )
     return ["Op"];
-  if (hasPermission(user, "deployment.airflow.user", entityType, deploymentId))
+  if (
+    hasPermission(user, "deployment.airflow.user", entityType, deploymentId) ||
+    hasPermission(user, "sysadmin.airflow.user")
```
ashb:
```suggestion
    hasPermission(user, "system.airflow.user")
```

src/routes/v1/authorization/handler.js
```diff
@@ -65,13 +65,25 @@ export function airflowJWT(user, roles, hostname) {
...
+  )
     return ["User"];
-  if (hasPermission(user, "deployment.airflow.get", entityType, deploymentId))
+  if (
+    hasPermission(user, "deployment.airflow.get", entityType, deploymentId) ||
+    hasPermission(user, "sysadmin.airflow.get")
```
ashb:
```suggestion
    hasPermission(user, "system.airflow.get")
```

PR #147, opened 2019-10-18
Report current image version to users 

andriisoldatenko:
Fix https://github.com/astronomer/issues/issues/477
related astro-cli PR https://github.com/astronomer/astro-cli/pull/303


andriisoldatenko:
@ashb @samblackk @schnie what is the best way to test it?
i found api docker webhook api and trying to run some integration test using `curl`:
```
curl -vvv -d '{"events":[{"id":"uuid","action":"push","target":{"repository":"cosmic-dust-1234/airflow","tag":"cli-1"},"request":{"host":"casual.domain"}}]}' -H "Content-Type: application/vnd.docker.distribution.events.v1+json" -X POST http://localhost:8871/v1/registry-events/
```

based on this unit tests src/routes/v1/registry/events/index.unit.test.js:11, but no luck with error:
**POST body missing. Did you forget use body-parser middleware?%**

any ideas how to test/debug it?

src/resolvers/mutation/update-deployment/index.js
```diff
@@ -44,7 +44,8 @@ export default async function updateDeployment(parent, args, ctx, info) {
   const updatablePayload = pick(args.payload, [
     "label",
     "description",
-    "version"
+    "version",
+    "currentTag"
```
ashb: This would let someone send a mutation to update the currentTag - given that would just affect the DB but not what is actually deployed we shouldn't allow that.

andriisoldatenko:
fixed. it works :)

```
curl -vvv -d '{"events":[{"id":"uuid","action":"push","target":{"repository":"blazing-ionization-2662/airflow","tag":"cli-1"},"request":{"host":"casual.domain"}}]}' -H "Content-Type: application/vnd.docker.distribution.events.v1+json" -X POST http://localhost:8871/v1/registry/events
```

houston-api logs:
```
2019-10-22T12:48:10 INFO Commander disabled, skipping call to #updateDeployment
2019-10-22T12:48:10 DEBUG POST /v1/registry/events 200 73.254 ms - 2
```

```
postgres=# select "releaseName", "currentTag" from "houston$default"."Deployment";
       releaseName       | currentTag
-------------------------+------------
 blazing-ionization-2662 | cli-1
(1 row)
```

src/resolvers/query/deployments/fragment.js
```diff
@@ -5,4 +5,5 @@ export default `fragment EnsureFields on Deployment {
   alertEmails
   airflowVersion
   config
+  currentTag
```
ashb: I don't think we need to add most of these - we only need to ask for the columns that the user has requested - nothing else in our system needs this column to exist.

src/resolvers/mutation/update-deployment/index.js
andriisoldatenko: removed

src/resolvers/query/deployments/fragment.js
andriisoldatenko: removed

src/schema.graphql
```diff
@@ -191,6 +191,7 @@ type Deployment {
   updatedAt: DateTime!
   roleBindings: [RoleBinding!]
   deploymentCapabilities: DeploymentCapabilities
+  currentTag: String
```
ashb: This info is already in `deployInfo` structure. We shouldn't add a new column, but instead change how that works.

andriisoldatenko:
@ashb @schnie PTAL

schnie:
@andriisoldatenko what happens if a user pushes an image tag that is not generated via the CLI, so no `cli-` prefix?. I noticed we dropped usage of the `findLatestTag()` field, which was used to remove all non-`cli-`-prefixed tags before generating the new one. I'm thinking that situation may not be accounted for with this.

@ashb do you think we should keep `currentTag`, or somehow integrate with the new `DockerImage` table in #155? https://github.com/astronomer/houston-api/blob/078e8202cd20c9c936b633f10dc461f9aa58d36d/database/datamodel.graphql#L132-L143

ashb:
I had this PR sort-of in mind when I wrote mine, and we can use `currentTag` of this to find the "right" DockerImage row to look at (my PR keeps a row per image that is pushed, and no indication other than implicit by createdAt date which oneis )

schnie:
@ashb ok that makes sense.

schnie:
@andriisoldatenko @ashb I don't think this will work if a user pushes a docker tag that is a git SHA or build number. That tag will get stored in the record and the [generateNextTag()](https://github.com/astronomer/houston-api/blob/master/src/lib/deployments/config/index.js#L437-L444) won't work properly without the [findLatestTag()](https://github.com/astronomer/houston-api/blob/master/src/lib/deployments/config/index.js#L429-L435) filtering out to find the _last_ `cli-` tag.

Is there a better way to generate a unique identifier as a tag from the CLI? Should we keep the `cli-n` convention? If we want to keep it, we may need to start tracking that number in a new field, separate from the `currentTag`. What do you guys think?

ashb:
> Is there a better way to generate a unique identifier as a tag from the CLI? Should we keep the cli-n convention? If we want to keep it, we may need to start tracking that number in a new field, separate from the currentTag. What do you guys think?

With the DockerImages model added in my other PR we have a list of all the tags, so we could filter use that to find the latest `cli-n` tag.

schnie:
@ashb oh sweet, that makes sense. @andriisoldatenko what do you think about making that update?

So, let's say we roll these changes out in our next release. The first time someone goes to `astro deploy` their project, they'd probably get `cli-1` back from houston as the tag to use, even though most of them probably have already pushed that tag previously.

Seems like we'll need some kind of migration in place to at least get this table sync'd up with the latest we have in the registry. Or is there a better way?

ashb:
> Seems like we'll need some kind of migration in place to at least get this table sync'd up with the latest we have in the registry. Or is there a better way?

Oh dang. Yeah totally. Either a migration thing to walk the images in the registry and populate the DockerIamge (using the API so we don't push the images and update any deployments) or else we could special case when there are no DockerImage rows and fall back to still asking the registry.

andriisoldatenko:
TODO for me:
- [ ] create data migration to fetch latest tag and insert into new `DockerImage` model
- [ ] refactor findLatestTag to reuse `DockerImage` and `tag_starts_with` query tnx @ashb for suggestion

src/resolvers/types/deployment/index.js
```diff
@@ -114,18 +111,40 @@ export async function deployInfo(parent) {
...
-    const repo = await request({
-      method: "GET",
-      uri,
-      json: true,
-      headers: { Authorization: `Bearer ${dockerJWT}` }
-    });
-
-    // Generate the response based on list.
-    const latest = findLatestTag(repo.tags);
+    let { tags } = await ctx.db.query.dockerImages(
```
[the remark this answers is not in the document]
schnie:
```suggestion
    const { tags } = await ctx.db.query.dockerImages(
```

nitpick, but just for consistency

src/resolvers/types/deployment/index.js
```diff
@@ -114,18 +111,40 @@ export async function deployInfo(parent) {
...
+      {
+        where: {
+          AND: [{ id: parent.id }, { tag_starts_with: "cli-" }]
```
schnie:
```suggestion
          AND: [{ deployment { id: parent.id } }, { tag_starts_with: "cli-" }]
```
@andriisoldatenko `parent.id` here would be the `Deployment` ID wouldn't it? Thinking it would look something like this suggestion. Haven't tested that though.

src/resolvers/types/deployment/index.js
```diff
@@ -114,18 +111,40 @@ export async function deployInfo(parent) {
...
+        }
+      },
+      `{ tag }`
+    );
+    if (!tags) {
+      tags = registryTags(parent);
```
schnie: @andriisoldatenko I thought we were going to run a migration instead of falling back to the registry.

src/resolvers/types/deployment/index.js
```diff
@@ -114,18 +111,40 @@ export async function deployInfo(parent) {
...
+    }
+    const deployment = await ctx.db.query.deployment(
```
schnie: @andriisoldatenko wouldn't we be able to drop this query if the query above works?

src/resolvers/types/deployment/index.js
andriisoldatenko: yes it's correct, but i can't find quick way to build events json from registry api to save in DockerImage model...

src/resolvers/types/deployment/index.js
```diff
@@ -87,19 +87,16 @@ export function type() {
...
-
+export async function registryTags(parent) {
   // Create a JWT for the registry request.
   const dockerJWT = await createDockerJWT("houston", [
     {
       type: "repository",
-      name: repo,
+      // Build the repo name.
+      name: `${parent.releaseName}/${DEPLOYMENT_AIRFLOW}`,
       actions: ["push", "pull"]
```
ashb: Not that you changed this line, but while we are here: we don't need push permission to get the tags, so lets fix that:

```suggestion
      actions: ["pull"]
```

src/resolvers/types/deployment/index.js
ashb: Yes, this seems more likely to be correct. Also can we just do this as

```suggestion
          { deployment: { id: parent.id }, tag_starts_with: "cli-" }
```

src/resolvers/types/deployment/index.js
ashb: We don't need the events, strictly. The pseudo code for the migration is this:

1. For each deployment record in the DB:
    1. Get list of tags by request to `https://${registry}/v2/${releaseName}/airflow/tags/list`
    1. For each tag
    ```js
        const ev = { 
          target: { 
            url: `http://${registry}/v2/${releaseName}/airflow/manifests/${tag}`, 
            mediaType: "application/vnd.docker.distribution.manifest.v2+json"
          }
        }

        const imageMetadata = await extractImageMetadata(ev);
        await prisma.createDockerImage({
            name: `${releaseName}/airflow:${tag}`,
            deployment: {
              connect: { releaseName }
            },
            labels: imageMetadata.labels,
            env: imageMetadata.env,
            tag: tag,
            digest: ev.target.digest
        });
    ```

src/resolvers/types/deployment/index.js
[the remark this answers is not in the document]
andriisoldatenko: fixed

src/resolvers/types/deployment/index.js
andriisoldatenko: fixed.

src/migrations/20191101143903_migrate_initial_docker_image.js
```diff
@@ -0,0 +1,80 @@
...
+export async function up(knex) {
+  const prismaConfig = config.get("prisma");
+  const prisma = new Prisma({
+    typeDefs: "src/lib/generated/schema/prisma.graphql",
+    endpoint: prismaConfig.endpoint,
+    secret: prismaConfig.secret,
+    debug: prismaConfig.debug
+  });
+
+  const { releaseNames } = await prisma.query.deployments(
```
andriisoldatenko: @ashb @schnie Regarding migrations. i found `knex` supports only postgresql connection, that's why i'm trying to add something like it ^ it doesn't work, anything i forget?

andriisoldatenko: or may be i need to switch to `SELECT` and `INSERT` instead of GraphQL?

src/migrations/20191101143903_migrate_initial_docker_image.js
```diff
@@ -0,0 +1,74 @@
...
+
+    for (tag of tags) {
+      // For each tag create record
+      const ev = {
+        target: {
+          url: `http://${registry}/v2/${repo}/manifests/${tag}`,
+          mediaType: "application/vnd.docker.distribution.manifest.v2+json"
+        }
+      };
+      const imageMetadata = await extractImageMetadata(ev);
```
andriisoldatenko: @ashb @schnie this line doesn't work for me... i see: `i see level=warning msg="error authorizing context: insufficient scope"` during calling registry https://github.com/astronomer/houston-api/blob/f352bf7e73cf252d6881ce38425fb2c29897eedd/src/routes/v1/registry/events/post.js#L125

andriisoldatenko: fixed in https://github.com/astronomer/houston-api/pull/147/commits/0739081591b1f74a179c8017c3b69e87da5fb04c

andriisoldatenko:
@schnie @samblackk finally it's ready for review:
history of practical tests:

Docker build and docker tag and push to local registry:
```
docker tag advanced-quark-6540/airflow:latest localhost:5001/advanced-quark-6540/airflow:cli-1
docker push localhost:5001/advanced-quark-6540/airflow:cli-1
```

Run migration:
```
➜  houston-api git:(report-current-image-version-to-users) ✗ JWT__REGISTRY__CERT_PATH=. JWT__REGISTRY__SERVICE=docker-registry JWT__REGISTRY__ISSUER=houston NODE_ENV=development PUBLIC_SIGNUPS=true npx babel-node node_modules/.bin/knex --inspect-brk "migrate:latest"
Debugger listening on ws://127.0.0.1:9229/cef3600a-754d-4f1c-b447-703d4ab72761
For help, see: https://nodejs.org/en/docs/inspector
Debugger attached.
Debugger listening on ws://127.0.0.1:9229/cef3600a-754d-4f1c-b447-703d4ab72761
For help, see: https://nodejs.org/en/docs/inspector
Debugger attached.
Using environment: development
2019-11-04T13:51:28 DEBUG Requesting docker tags for advanced-quark-6540 at http://localhost:5001/v2/advanced-quark-6540/airflow/tags/list
(node:84626) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
2019-11-04T13:51:31 INFO Fetching manifest from http://localhost:5001/v2/advanced-quark-6540/airflow/manifests/cli-1
2019-11-04T13:51:31 DEBUG Fectching image config from http://localhost:5001/v2/advanced-quark-6540/airflow/blobs/sha256:7d8b9918d209e47f3fa6383ec74b1c8bb8ab980145e23ee491597b5a56418ab2
Batch 2 run: 1 migrations
20191101143903_migrate_initial_docker_image.js
Waiting for the debugger to disconnect...
```

Check db:
```
postgres=# select "digest", "deployment" from "houston$default"."DockerImage";
                                 digest                                  |        deployment
-------------------------------------------------------------------------+---------------------------
 sha256:7d8b9918d209e47f3fa6383ec74b1c8bb8ab980145e23ee491597b5a56418ab2 | ck2gh74fs002h0720940bgdhj
(1 row)
```

🎉

src/migrations/20191101143903_migrate_initial_docker_image.js
```diff
@@ -0,0 +1,73 @@
...
+      // For each tag create record
+      const ev = {
+        target: {
+          url: `http://${registry}/v2/${repo}/manifests/${tag}`,
+          repository: repo,
+          mediaType: "application/vnd.docker.distribution.manifest.v2+json"
+        }
+      };
+      const imageMetadata = await extractImageMetadata(ev);
+      await prisma.createDockerImage({
```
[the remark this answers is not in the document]
andriisoldatenko: do you mean silently skip on data integrity error?

src/migrations/20191101143903_migrate_initial_docker_image.js
```diff
@@ -0,0 +1,77 @@
...
+  const {
+    releaseNamespace: namespace,
+    releaseName: platformReleaseName
+  } = config.get("helm");
+
+  const registryPort = config.get("registry.port");
+
+  let deployment;
+
+  for (deployment of deployments) {
```
andriisoldatenko: let's add try / catch if registry doesn't exists.

andriisoldatenko: fixed

andriisoldatenko:
@schnie PTAL

src/resolvers/types/deployment/index.js
```diff
@@ -91,50 +87,25 @@ export function type() {
...
+    `{ tag }`
+  );
+  const tags = Array.from(images, x => x.tag);
+  let latest = findLatestTag(tags);
+  if (!latest) {
+    const deployment = await ctx.db.query.deployment(
+      { where: { id: parent.id } },
+      `{ currentTag }`
+    );
+    latest = deployment.currentTag;
```
schnie: @andriisoldatenko can't we delete this block?

src/resolvers/types/deployment/index.js
```diff
@@ -91,50 +87,25 @@ export function type() {
...
-    // If we get a 404, that means nobody has pushed yet, so just return empty.
-    if (err.statusCode === NOT_FOUND) return {};
-
-    // Otherwise throw the actual error.
-    throw err;
+      where: { deployment: { id: parent.id }, tag_starts_with: "cli-" }
+    },
+    `{ tag }`
+  );
+  const tags = Array.from(images, x => x.tag);
```
schnie: @andriisoldatenko I like to use `lodash` to do a lot of these kinds of things. It's super helpful

https://lodash.com/docs/#pick
```suggestion
  const tags = pick(images, "tag");
```

src/resolvers/types/deployment/index.js
andriisoldatenko: do you think we should remove `currentTag`?

src/resolvers/types/deployment/index.js
andriisoldatenko: fixed using `map` https://lodash.com/docs/#map
```
var users = [
  { 'user': 'barney' },
  { 'user': 'fred' }
];
 
// The `_.property` iteratee shorthand.
_.map(users, 'user');
// => ['barney', 'fred']
```

src/resolvers/types/deployment/index.js
andriisoldatenko: done.

andriisoldatenko:
@schnie pls take another look

src/resolvers/types/deployment/index.js
```diff
@@ -91,50 +87,17 @@ export function type() {
...
-
-    // Otherwise throw the actual error.
-    throw err;
-  }
+      where: { deployment: { id: parent.id }, tag_starts_with: "cli-" }
+    },
+    `{ tag }`
+  );
+  const tags = map(images, "tag");
+  let latest = findLatestTag(tags);
```
schnie: @andriisoldatenko nit but this could be `const`

andriisoldatenko: @schnie fixed.

PR #156, opened 2019-10-28
Grant registry PULL permission to other images to allow layer sharing

ashb:
One of the "silly" slow steps of pushing the first layer to a deployment
is that no layers are shared with existing images, so we have to push
the whole thing up.

The Docker Registry will allow layers to be shared from other images
that the user has access to; so to make this initial push quicker I have
granted universal PULL to a set of "baseImages" (pushable by sysadmins.)

Additionally to share layers between deployments (i.e. when re-tagging
an image for staging to promote it to prod) the scopes returned now
include read permissions to any other deployment the user has access
to.

(With service tokens this would allow layer sharing in a workspace, but
there is no way currently to share layers across workspaces when using
service tokens to push.)

I have tested this manually, but I think this is probably worth adding more unit testing to cover it.

Example of this working:

**Pushing the base image**

```
$ docker push localhost:5001/base-images/airflow:0.10.2-1.10.5
The push refers to repository [localhost:5001/base-images/airflow]
356252343ba8: Pushed
88235de03c4e: Pushed
b2f2d6ca1b0a: Pushed
233ea8b8acb1: Pushed
b9043bc56416: Pushed
5de3948d4d61: Pushed
3845e2a7b37f: Pushed
03901b4a2ea8: Pushed
0.10.2-1.10.5: digest: sha256:590fda7c8920076c0f92d52750d66637079a8d6af5755804ec02a4d1fee15511 size: 1984
```

**Pushing an image with one change from base**:

```
$ docker push localhost:5001/native-intensity-7824/airflow:cli-2
The push refers to repository [localhost:5001/native-intensity-7824/airflow]
8b4214c20750: Pushed
8a85d994c094: Pushed
356252343ba8: Layer already exists
88235de03c4e: Layer already exists
b2f2d6ca1b0a: Layer already exists
233ea8b8acb1: Layer already exists
b9043bc56416: Mounted from base-images/airflow
5de3948d4d61: Layer already exists
3845e2a7b37f: Layer already exists
03901b4a2ea8: Layer already exists
cli-2: digest: sha256:8d1b342140e1ab418068e0857acea877acd95e4f006001dc92aabea9f626698d size: 2402
```

**Pushing an image with one change from another image in workspace**:
```
$ docker push localhost:5001/telescopic-sun-2787/airflow:cli-3
The push refers to repository [localhost:5001/telescopic-sun-2787/airflow]
8b4214c20750: Mounted from native-intensity-7824/airflow
8a85d994c094: Mounted from native-intensity-7824/airflow
356252343ba8: Mounted from base-images/airflow
88235de03c4e: Mounted from base-images/airflow
b2f2d6ca1b0a: Mounted from base-images/airflow
233ea8b8acb1: Mounted from base-images/airflow
b9043bc56416: Mounted from native-intensity-7824/airflow
5de3948d4d61: Mounted from base-images/airflow
3845e2a7b37f: Mounted from base-images/airflow
03901b4a2ea8: Mounted from base-images/airflow
cli-3: digest: sha256:dd05a2bd3db4d35b6d8df0d213188486c8216fc2abeba16d52a80661c0a4dbc5 size: 2402
```


andriisoldatenko:
@ashb great! it's super cool feature!

src/lib/deployments/validate/docker-tag/index.js
```diff
@@ -1,13 +1,15 @@
...
  */
-const ACTIONS = {
+export const ACTIONS = {
   PUSH: "push",
   PULL: "pull",
   MOUNT: "mount",
   DELETE: "delete"
 };
 
+export const VALID_RELEASE_NAME = /^(?<releaseName>[a-z]+-[a-z]+-[0-9]{0,4})\/airflow$/;
```
andriisoldatenko: copy/paste? do you want to add few tests?!

src/routes/v1/registry/authorization/handler.js
```diff
@@ -69,42 +79,91 @@ export default async function(req, res) {
...
+      if (includes(baseImages, parsed.name)) {
+        // Only sys admins can push to the base images, but anyone can pull from them
+        if (isPushAction) {
+          const allowed = hasPermission(user, "system.registryBaseImages.push");
+
+          if (!allowed) {
+            return sendError(
+              res,
+              REGISTRY_CODES.DENIED,
+              `You do not have permission to manage ${parsed.name}`
```
andriisoldatenko: i think better to add permission name to error message for better troubleshooting?

src/lib/deployments/validate/docker-tag/index.js
ashb: Yes to adding tests. Don't think this is c/p'd from anywhere else now?

schnie:
@ashb awesome! What happens if the `base-image/airflow` does not exist in the repository?

ashb:
Nothing at all :) It behaves exactly as it does currently.

ashb:
It would be sweet if we could work out how to pre-pull that image as part of the platform deploy, but I'm not quite sure how we can do that as even from inside the registry container we'd need valid auth to push to that repository.

schnie:
Yea, I was thinking it could be a Kubernetes `Job` that gets fired after a `helm install/upgrade` so it'd work in Enterprise installs out of the box. Would still need to figure that auth part out.

schnie:
@ashb thinking through the situation where we push image version 1 then we follow up a month later and push version 2 that changes something in the base layer(s), like a dependency. Both versions (tags) would exist in the repository right? Meaning that as soon as another version is pushed up, a user would have access to them and they become "usable" right?

ashb:
> Yea, I was thinking it could be a Kubernetes `Job` that gets fired after a `helm install/upgrade` so it'd work in Enterprise installs out of the box. Would still need to figure that auth part out.

Oh that could work. If we run it in the helm container (but with a different entrypoint) we can use the `createDockerJWT` function to generate the command and then run `docker pull/docker tag/docker push`? (Or to not need a running docker daemon which is more tricky in container than it sounds we could use the HTTP api to handle the push/pull etc.)

ashb:
> @ashb thinking through the situation where we push image version 1 then we follow up a month later and push version 2 that changes something in the base layer(s), like a dependency. Both versions (tags) would exist in the repository right? Meaning that as soon as another version is pushed up, a user would have access to them and they become "usable" right?

Correct. It does the sharing based on the id's of the layers, not anything to do with the tag of the images. So unless we delete the tags in base-images/airflow it would still be available for sharing.

src/routes/v1/registry/authorization/handler.js
ashb: Done.

ashb:
Working on adding some more tests to this now. This endpoint is sorely lacking in unit tests!

src/routes/v1/registry/authorization/index.unit.test.js
```diff
@@ -1,18 +1,158 @@
...
-import request from "supertest";
+import * as prisma from "generated/client";
+import { createJWT, decodeJWT } from "jwt";
+import * as rbac from "rbac";
+import casual from "casual";
+import supertest from "supertest";
 import express from "express";
+import { DEPLOYMENT_ADMIN } from "constants";
+
+jest.mock("rbac");
```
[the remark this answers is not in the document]
ashb: /cc @schnie @samblackk @andriisoldatenko This might be a nicer way of writing our unit tests in Houston (easier to mock I think)

src/routes/v1/registry/authorization/index.unit.test.js
```diff
@@ -1,18 +1,158 @@
...
+
+      expect(res.statusCode).toBe(401);
+      expect(res.body).not.toHaveProperty("token");
+    });
+
+    test.each`
+      scope          | expectedPermission          | expected
+      ${"push,pull"} | ${"deployment.images.push"} | ${["push", "pull"]}
+      ${"pull"}      | ${"deployment.images.pull"} | ${["pull"]}
+    `(
```
ashb: Slightly odd syntax here: https://jestjs.io/docs/en/api#2-testeach-table-name-fn-timeout (I used this one form rather than the other one as I couldn't get the test names to be set as nicely with just the `test.each([[a,b]], () => {})` style

schnie:
Looks great!

PR #159, opened 2019-10-30
Fix: Query WorkspaceUsers by fullName

samblackk:
Attached to https://github.com/astronomer/issues/issues/488 & https://github.com/astronomer/orbit-ui/pull/182

Allows querying `workspaceUsers` by `fullName`.

### DIFF
--- a/src/resolvers/query/workspace-users/index.js
+++ b/src/resolvers/query/workspace-users/index.js
@@ -9,7 +9,7 @@ import { addFragmentToInfo } from "graphql-binding";
  */
 export default async function workspaceUsers(parent, args, ctx, info) {
   // Build the users query.
-  const query = invitesQuery(args);
+  const query = usersQuery(args);
 
   return ctx.db.query.users(
     {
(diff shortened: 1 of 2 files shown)


src/resolvers/query/workspace-users/index.js
```diff
@@ -30,12 +30,15 @@ export default async function workspaceUsers(parent, args, ctx, info) {
...
  * @return {Object} The users query.
  */
-export function invitesQuery(args) {
+export function usersQuery(args) {
   // Pull out some args.
-  const { email } = args;
+  const { fullName, email } = args;
+
+  // If we have username, use it.
+  if (fullName) return { fullName_contains: fullName };
```
schnie: @samblackk if the orbit change is going to always send `fullName` and `email`, then wouldn't this always return here and never get to email? Wouldn't we want to return results that match either here?

samblackk: Fixed in a new PR.
https://github.com/astronomer/houston-api/pull/161

PR #161, opened 2019-10-30
Fix: Workspaceusers OR query

samblackk:
Related to comment on https://github.com/astronomer/houston-api/pull/159
Build's query via an OR statement instead of replacing.


src/resolvers/query/workspace-users/index.js
```diff
@@ -14,9 +14,13 @@ export default async function workspaceUsers(parent, args, ctx, info) {
...
-          workspace: { id: args.workspaceUuid }
-        },
+        AND: [
+          {
+            roleBindings_some: {
+              workspace: { id: args.workspaceUuid }
+            }
+          }
+        ],
         ...query
```
schnie: @samblackk would we want the `...query` to be a part of this `AND` array?

src/resolvers/query/workspace-users/index.js
```diff
@@ -34,11 +38,15 @@ export function usersQuery(args) {
   // Pull out some args.
   const { fullName, email } = args;
 
-  // If we have username, use it.
-  if (fullName) return { fullName_contains: fullName };
-
-  // If we have an email use it.
-  if (email) return { emails_some: { address: email.toLowerCase() } };
+  // If we have username and email, use it.
+  if (fullName && email)
```
schnie: @samblackk I know orbit is sending both of these up, but it seems like it should be possible to build up this `OR` block more granularly.

If we have `fullName`, add the first chunk, and if we have `email`, add the second two chunks here.

I just like to think about all possible clients, including folks making direct API calls themselves. What do you think?

src/resolvers/query/workspace-users/index.js
samblackk: I don't think so - because we need the roleBinding in the query at all times plus _any_ of the items in the `OR` portion. You can't put `OR` in `AND`. However, with the other change you suggested - the use of `OR` would probably be unnecessary.

src/resolvers/query/workspace-users/index.js
schnie: @samblackk think we could get this separated out today?

samblackk: Yes. Working on it this morning.

src/resolvers/query/workspace-users/index.js
```diff
@@ -32,13 +21,35 @@ export default async function workspaceUsers(parent, args, ctx, info) {
  */
 export function usersQuery(args) {
   // Pull out some args.
-  const { fullName, email } = args;
+  const { fullName, email, workspaceUuid } = args;
+
+  // Init query
+  let query = {
+    where: {
+      AND: [
```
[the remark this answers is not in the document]
samblackk: The query wouldn't resolve correctly without it.

PR #162, opened 2019-11-01
Add raw link to verify email html

petedejoy:
We previously added this to the invite email html template, but forgot to add it to the confirm email template as well.


petedejoy:
All tested here! 
image

PR #163, opened 2019-11-06
Clean up deleted deployment data

andriisoldatenko:
fix https://github.com/astronomer/issues/issues/262
related PR https://github.com/astronomer/helm.astronomer.io/pull/235


src/schema.graphql
```diff
@@ -189,6 +189,7 @@ type Deployment {
   workspace: Workspace
   createdAt: DateTime!
   updatedAt: DateTime!
+  deletedAt: DateTime
```
ashb: This exposes the deletedAt to consumers of the Houston API, and I don't think we should do that - if we want to be able to un-delete deployments that is probably a separate (admin-only?) mutation in our API

src/resolvers/mutation/delete-deployment/index.unit.test.js
```diff
@@ -87,7 +87,7 @@ describe("deleteDeployment", () => {
       );
 
       expect(res.errors).toBeUndefined();
-      expect(deleteDeployment.mock.calls.length).toBe(1);
+      expect(updateDeployment.mock.calls.length).toBe(1);
```
ashb: It's probably important that we test _what_ we update about the deployment, so all of the cases in this should look something like this:
```suggestion
      expect(updateDeployment).toBeCalledTimes(1);
      expect(updateDeployment).toBeCalledWith({
        where: { id },
        data: { deletedAt: expect.any(Date) }
      });
```

(`toBeCalledTimes` is just my personal preference, but is functionally the same as what is already there)

src/schema.graphql
schnie: Yea, I think we should remove this too.

andriisoldatenko: done.

src/scripts/registry-cleanup.js
```diff
@@ -0,0 +1,79 @@
...
+import request from "request-promise-native";
+import config from "config";
+import { DEPLOYMENT_AIRFLOW } from "constants";
+
+log.info("Starting registry cleanup");
+
+function getManifestoForSelectedTag(dockerJWT, registry, repo, tag) {
+  let uri;
+  uri = `https://${registry}/v2/${repo}/airflow/manifests/${tag}`;
+  let { headers } = request({
```
schnie: @andriisoldatenko do we want to `await request` here? I'm guessing `request` is returning a `Promise` here? If we `await`, we'll want to make this an `async function` and `await` it down below as well.

schnie:
@andriisoldatenko as part of this, we'll also want to `DROP` the database associated with this deployment we are cleaning up.

andriisoldatenko:
@schnie can you point me in code how to find correct DB? somewhere in k8s/helm?

schnie:
Code related to DB creation is here https://github.com/astronomer/houston-api/blob/master/src/lib/deployments/database/index.js.

I think it uses this env var to connect and create DBs https://github.com/astronomer/helm.astronomer.io/blob/e4ff96a4cc444e965cadb94e32e1ba86f0a97a46/charts/astronomer/templates/houston/houston-deployment.yaml#L134

src/scripts/registry-cleanup.js
andriisoldatenko: fixed.

andriisoldatenko:
> Code related to DB creation is here https://github.com/astronomer/houston-api/blob/master/src/lib/deployments/database/index.js.
> 
> I think it uses this env var to connect and create DBs https://github.com/astronomer/helm.astronomer.io/blob/e4ff96a4cc444e965cadb94e32e1ba86f0a97a46/charts/astronomer/templates/houston/houston-deployment.yaml#L134

@schnie something like this: 577770762c16defb397345403ea636447378361d ?

src/scripts/cleanup.js
```diff
@@ -0,0 +1,123 @@
...
+
+async function cleanup() {
+  log.info("Starting registry cleanup");
+  const deployments = await prisma.deployments(
+    { where: { deletedAt_not: null } },
+    `{ releaseName }`
+  );
+  log.info(deployments);
+  if (deployments == null) {
+    return;
```
schnie: @andriisoldatenko how about a `log.info()` here to indicate that the job didn't find anything to help keep an eye on things?

src/resolvers/query/workspace-deployments/index.js
```diff
@@ -13,6 +13,7 @@ export default async function workspaceDeployments(parent, args, ctx, info) {
   // Build the deployments query.
   const query = deploymentsQuery(args, ctx);
 
+  query.where.AND.push({ deletedAt: null });
```
schnie: @andriisoldatenko how about moving this part to the `deploymentsQuery` function, since the scope of that function is to generate the full query.

schnie:
Similar to `updateDeployent`, we should probably update the query in `upgradeDeployment` - https://github.com/astronomer/houston-api/blob/master/src/resolvers/mutation/upgrade-deployment/index.js#L15-L20.

This is the resolver that is responsible for `helm upgrade`ing to a new chart version when the little "Upgrade" button shows up in the UI.

andriisoldatenko:
> updateDeployent

@schnie after quick look into source code, i'm a-little bit confused:
https://github.com/astronomer/houston-api/blob/efecc11d6205826520528b5c4133f73037800c93/src/resolvers/mutation/upgrade-deployment/index.js#L17

```
export default async function upgradeDeployment(parent, args, ctx, info) {
  // Update the deployment in the database.
  const where = { id: args.deploymentUuid };
  const data = { version: args.version };
  const updatedDeployment = await ctx.db.mutation.updateDeployment(
    { where, data },
    addFragmentToInfo(info, fragment)
  );
```
`upgradeDeployment` relates on `updateDeployment`.
what should i change? 
cc @ashb

andriisoldatenko:
what about:
```
export default async function logs(parent, args, ctx) {
  const { releaseName } = await ctx.db.query.deployment(
    { where: { id: args.deploymentUuid } },
    `{ releaseName }`
  );
https://github.com/astronomer/houston-api/blob/aa450497371aff756b8667ddede7a90873527ba7/src/resolvers/query/logs/index.js#L13
```

src/scripts/cleanup.js
```diff
@@ -0,0 +1,125 @@
...
+  const repo = `${releaseName}/${DEPLOYMENT_AIRFLOW}`;
+  const registry = `${platformReleaseName}-registry.${namespace}:${registryPort}`;
+
+  // Create a JWT for the registry request.
+  const dockerJWT = createDockerJWT("houston", [
+    {
+      type: "repository",
+      // Build the repo name.
+      name: repo,
+      actions: ["pull"]
```
ashb:
```suggestion
      actions: ["pull", "delete"]
```

Then we could use the same JWT token for both requests

src/scripts/cleanup.js
```diff
@@ -0,0 +1,125 @@
...
+    uri,
+    json: true,
+    headers: {
+      Authorization: `Bearer ${dockerJWT}`,
+      "User-Agent": `houston/${version()}`
+    }
+  });
+}
+
+async function cleanupRegistry(deployment) {
```
ashb:
```suggestion
async function cleanupImagesForDeployment(deployment) {
```

src/scripts/cleanup.js
```diff
@@ -0,0 +1,125 @@
...
+      tag
+    );
+    await deleteManifest(registry, repo, digestHash);
+  }
+}
+
+async function cleanup() {
+  log.info("Starting registry cleanup");
+  const deployments = await prisma.deployments(
+    { where: { deletedAt_not: null } },
```
ashb: This is going to delete any deleted deployment as soon as we run this script -- we probably want to have the script take a number of days old to look for, i.e. `bable-node src/scripts/cleanup.js --older-than 14`

ashb:
Oh, and well have to add the deleteAt=Null to the  src/resolvers/types/deployment/index.js change that you add in #147 so we need to merge that one first.

src/scripts/cleanup.js
andriisoldatenko: nice catch!

package.json
```diff
@@ -14,7 +14,8 @@
     "migrate": "npm run knex-migrate && npm run prisma-deploy",
     "knex": "babel-node node_modules/.bin/knex",
     "knex-migrate": "npm run knex migrate:latest",
-    "prisma-deploy": "prisma deploy --force"
+    "prisma-deploy": "prisma deploy --force",
+    "registry-cleanup": "babel-node src/scripts/registry-cleanup.js"
```
schnie:
```suggestion
    "cleanup": "babel-node src/scripts/cleanup.js"
```

schnie: @andriisoldatenko need to update this to match the new name.

ashb: I think `registry-cleanup` (or `deleted-deployment-cleanup`) is a better name.

andriisoldatenko: @ashb it's not only about registry cleanup it's registry clean up + database cleanup.
what about `deployment-cleanup` ?

src/resolvers/mutation/update-deployment/index.js
```diff
@@ -44,7 +44,8 @@ export default async function updateDeployment(parent, args, ctx, info) {
   const updatablePayload = pick(args.payload, [
     "label",
     "description",
-    "version"
+    "version",
+    "deletedAt"
```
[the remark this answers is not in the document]
andriisoldatenko: done.

src/scripts/cleanup.js
andriisoldatenko: fixed.

src/scripts/cleanup.js
andriisoldatenko: fixed.

andriisoldatenko:
@schnie @ashb PTAL

src/scripts/cleanup.js
andriisoldatenko: done.

package.json
andriisoldatenko: fixed.

package.json
```diff
@@ -14,7 +14,8 @@
     "migrate": "npm run knex-migrate && npm run prisma-deploy",
     "knex": "babel-node node_modules/.bin/knex",
     "knex-migrate": "npm run knex migrate:latest",
-    "prisma-deploy": "prisma deploy --force"
+    "prisma-deploy": "prisma deploy --force",
+    "deployment-cleanup": "babel-node src/scripts/deployment-cleanup.js --older-than 14"
```
schnie: @andriisoldatenko if you remove the `--older-than 14` here, can we have the `CronJob`'s command be `npm run deployment-cleanup --older-than 14`? I'm thinking that we'd want to have the retention time configurable/template-able from helm.

src/scripts/deployment-cleanup.js
```diff
@@ -0,0 +1,129 @@
...
+import log from "logger";
+import { version } from "utilities";
+import { createDockerJWT } from "registry/jwt";
+import { prisma } from "generated/client";
+import { removeDatabaseForDeployment } from "deployments/database";
+import request from "request-promise-native";
+import config from "config";
+import { DEPLOYMENT_AIRFLOW } from "constants";
+
+const MEDIATYPE_DOCKER_MANIFEST_V2 =
```
schnie: @andriisoldatenko @ashb I think this string is now hardcoded at least twice in Houston. Think we should move it to `constants.js`? https://github.com/astronomer/houston-api/blob/master/src/lib/constants/index.js

src/scripts/deployment-cleanup.js
```diff
@@ -0,0 +1,129 @@
...
+    log.info("There is no deployments to delete :(");
+    return;
+  }
+  for (let deployment of deployments) {
+    await cleanupImagesForDeployment(deployment);
+    await removeDatabaseForDeployment(deployment);
+  }
+}
+
+const yargs = require("yargs");
```
schnie: @andriisoldatenko do we need to add `yargs` to `package.json`? Usually you can `npm install yargs --save` to ensure it gets saved to the file.

Also, could this be moved to the top of the file and use `import` syntax for consistency?
`import yargs from "yargs"` should work.

andriisoldatenko: fixed.

src/scripts/deployment-cleanup.js
andriisoldatenko: fixed.

package.json
andriisoldatenko: fixed.

src/scripts/deployment-cleanup.js
```diff
@@ -0,0 +1,125 @@
...
+import { createDockerJWT } from "registry/jwt";
+import { prisma } from "generated/client";
+import { removeDatabaseForDeployment } from "deployments/database";
+import request from "request-promise-native";
+import config from "config";
+import yargs from "yargs";
+import { DEPLOYMENT_AIRFLOW, MEDIATYPE_DOCKER_MANIFEST_V2 } from "constants";
+
+async function getManifestoForSelectedTag(dockerJWT, registry, repo, tag) {
+  let uri = `http://${registry}/v2/${repo}/manifests/${tag}`;
```
schnie: Could be `const`

src/scripts/deployment-cleanup.js
```diff
@@ -0,0 +1,125 @@
...
+  log.debug(`Requesting docker tag manifest for ${uri} at ${tag}`);
+  let response = await request({
```
schnie: same

src/lib/deployments/database/index.js
```diff
@@ -100,6 +100,29 @@ export async function createDatabaseForDeployment(deployment) {
...
+
+  // Parse the connection into a standard format.
+  const parsedConn = parseConnection(connection);
+
+  // Create connection as root user.
+  const rootConn = createConnection(parsedConn);
+
+  const dbName = generateDatabaseName(deployment.releaseName);
+  log.info(`Droping database: ${dbName}`);
+  await rootConn.raw(`DROP DATABASE ${dbName};`);
```
andriisoldatenko: @schnie what do you think about this code? not sure how to test it...

andriisoldatenko: cc @ashb

src/scripts/deployment-cleanup.js
```diff
@@ -0,0 +1,125 @@
...
+import { version } from "utilities";
+import { createDockerJWT } from "registry/jwt";
+import { prisma } from "generated/client";
+import { removeDatabaseForDeployment } from "deployments/database";
+import request from "request-promise-native";
+import config from "config";
+import yargs from "yargs";
+import { DEPLOYMENT_AIRFLOW, MEDIATYPE_DOCKER_MANIFEST_V2 } from "constants";
+
+async function getManifestoForSelectedTag(dockerJWT, registry, repo, tag) {
```
schnie: @andriisoldatenko typically we like to document each function with a description, and documentation of args and return type. Here's an example - https://github.com/astronomer/houston-api/blob/master/src/lib/deployments/config/index.js#L32-L37.

andriisoldatenko: @schnie fixed. ****

schnie: I think most of these would be `{String}` instead of `{Object}`.

andriisoldatenko: fixed.

schnie:
@andriisoldatenko @ashb I found the following entries where we are searching for a deployment, but are not filtering with `deletedAt: null`. We should probably add that to these too right? For instance, if an alert gets fired by prometheus, but the deployment has actually been deleted recently, then we wouldn't want to return that deployment to send alert emails

- https://github.com/astronomer/houston-api/blob/master/src/routes/v1/alerts/post.js#L21

- https://github.com/astronomer/houston-api/blob/master/src/routes/v1/authorization/handler.js#L33

- https://github.com/astronomer/houston-api/blob/master/src/lib/deployments/validate/authorization/index.js#L25

- https://github.com/astronomer/houston-api/blob/master/src/routes/v1/registry/authorization/handler.js#L125

andriisoldatenko:
@schnie fixed in a8068809ec0a7470058b56f5e1cb3c9cb69f89ab

schnie:
@andriisoldatenko can you search the code for any other references where we may need to adjust our query logic to add the `deletedAt: null` part? I only did a quick search for that last batch.

Also, I think we may want to patch up this https://github.com/astronomer/houston-api/blob/master/src/lib/deployments/validate/existence/index.js#L17-L22. The comment explains why we need it currently. 

Is there actually a `unique` constraint on `releaseName` in the underlying db table? I forget. This is kind of a separate issue, but I think this PR could exaggerate it if we never delete them.

This makes me wonder if our actual cleanup-deployments script should also finally delete the actual underlying record as well to prevent this.

@ashb @andriisoldatenko what are your thoughts?

andriisoldatenko:
> can you search the code for any other references where we may need to adjust our query logic to add the deletedAt: null part? I only did a quick search for that last batch.

done
>Also, I think we may want to patch up this https://github.com/astronomer/houston-api/blob/master/src/lib/deployments/validate/existence/index.js#L17-L22. The comment explains why we need it currently.

done

> Is there actually a unique constraint on releaseName in the underlying db table? I forget. This is kind of a separate issue, but I think this PR could exaggerate it if we never delete them.

yes 
https://github.com/astronomer/houston-api/blob/8a44d3d8dbf5ca1089ef722b0cb94a337354327e/database/datamodel.graphql#L120

>This makes me wonder if our actual cleanup-deployments script should also finally delete the actual underlying record as well to prevent this.

i don't see any risks, ok with me to remove it also every 15 days.

@ashb ?

src/routes/v1/registry/events/post.js
```diff
@@ -31,7 +31,9 @@ export default async function(req, res) {
       );
 
       // Get the existing config for this deployment.
-      const config = await prisma.deployment({ releaseName }).config();
+      const config = await prisma
```
schnie: @andriisoldatenko @ashb what would happen here if we receive a webhook for a deployment that has been deleted recently? Probably an edge-case, but wondering if we should add a guard here:

`if (!config) return;`

schnie:
@andriisoldatenko thinking we should go ahead and remove the record from the DB in the cleanup script. This keeps the overall behavior similar to how it is today and we don't introduce any _new_ complications. We can change this behavior later if we decide to.

src/routes/v1/registry/events/post.js
andriisoldatenko: fixed in d8531e3

andriisoldatenko:
> @andriisoldatenko thinking we should go ahead and remove the record from the DB in the cleanup script. This keeps the overall behavior similar to how it is today and we don't introduce any _new_ complications. We can change this behavior later if we decide to.

done in d8531e3

andriisoldatenko:
@schnie @ashb PTAL

PR #164, opened 2019-11-11
Add ability to get current non cli- tag

andriisoldatenko:
+ fixed 2 warnings.
fix astronomer/issues#477
related PR https://github.com/astronomer/astro-cli/pull/303


src/resolvers/types/deployment/index.js
```diff
@@ -97,7 +97,17 @@ export async function deployInfo(parent, args, ctx) {
...
+  const imagesCreated = await ctx.db.query.dockerImages(
+    {
+      where: { deployment: { id: parent.id } },
+      order: [["created", "DESC"]],
+      limit: 1
+    },
+    `{ tag }`
+  );
+  let currentTag = map(imagesCreated, "tag")[0];
+  return { latest, next, currentTag };
```
schnie: @andriisoldatenko feels like we may want to figure out a better way to name these with the new addition. 

Maybe we stay consistent and omit `tag`, or find better names altogether. We'd need to update clients though.

src/resolvers/types/deployment/index.js
```diff
@@ -97,7 +97,17 @@ export async function deployInfo(parent, args, ctx) {
...
+
+  const imagesCreated = await ctx.db.query.dockerImages(
+    {
+      where: { deployment: { id: parent.id } },
+      order: [["created", "DESC"]],
+      limit: 1
+    },
+    `{ tag }`
+  );
+  let currentTag = map(imagesCreated, "tag")[0];
```
schnie:
```suggestion
  const current = map(imagesCreated, "tag")[0];
```

src/resolvers/types/deployment/index.js
schnie:
```suggestion
  return { latest, next, current };
```

src/schema.graphql
```diff
@@ -167,6 +167,7 @@ type DeploymentUrl {
 type DeployInfo {
   latest: String
   next: String
+  currentTag: String
```
```suggestion
  current: String
```

ashb: We could use `@deprecated` and change this to be

```graphql
type DeployInfo {
  current: String
  nextCli: String
  latest: String @deprecated
  next: String @deprecate @alias(field: "nextCli")
```

(I feel we don't need `latest` which is the latest `cli-` tagged one anymore

andriisoldatenko: @ashb @schnie fixed!

src/resolvers/types/deployment/index.js
andriisoldatenko: fixed.

src/resolvers/types/deployment/index.js
andriisoldatenko: fixed.

andriisoldatenko:
@ashb @schnie PTAL

src/resolvers/types/deployment/index.js
```diff
@@ -97,7 +97,17 @@ export async function deployInfo(parent, args, ctx) {
...
+  const imagesCreated = await ctx.db.query.dockerImages(
+    {
+      where: { deployment: { id: parent.id } },
+      order: [["created", "DESC"]],
+      limit: 1
+    },
+    `{ tag }`
+  );
+  const current = map(imagesCreated, "tag")[0];
+  return { latest, next, current };
```
schnie: @andriisoldatenko don't we need to update here too?

src/schema.graphql
```diff
@@ -165,8 +165,10 @@ type DeploymentUrl {
 }
 
 type DeployInfo {
-  latest: String
-  next: String
+  current: String
+  nextCli: String
+  latest: String @deprecated
+  next: String @deprecated @alias(field: "nextCli")
```
schnie: @samblackk do we use these fields in orbit? Anything we need to update there?

src/resolvers/types/deployment/index.js
andriisoldatenko: fixed.

andriisoldatenko: fixed.

PR #165, opened 2019-11-13
Add deployment status web subscription

samblackk:
Adds a web subscription that returns only the PromQL scheduler heartbeat metric. To be used as the data for good/bad indicator in the new orbit deployment lists. 

Attached to https://github.com/astronomer/new-orbit/pull/67

Not ready for merge, is producing this error: 

Screen Shot 2019-11-13 at 2 09 05 PM


schnie:
@samblackk did you get this patched up? Just let me know whenever its working and ready to merge.

samblackk:
Not yet @schnie - the fix for this is more involved than expected and has to do with how the `createPoller()` function is returning a copy of `AsyncIterator`. Will be refactoring tomorrow.

samblackk:
@schnie Issue is resolved. Refactoring done to `createPoller()`.
Tested for issues in current orbit subscriptions - metrics & logs - and everything is a-okay. 

Please test & merge!

src/lib/pubsub/poller/index.js
```diff
@@ -38,13 +38,13 @@ export default function createPoller(
...
   // when a subscription is cancelled.
-  return {
-    ...iterator,
-    return: () => {
-      log.info(`Disconnecting subscription ${topic}`);
-      clearInterval(poll);
-      clearTimeout(kill);
-      return iterator.return();
-    }
+  iterator.return = () => {
```
schnie: @samblackk so what happened here? What made it decide to break now?

src/schema.graphql
```diff
@@ -241,6 +241,10 @@ type Metric {
   result: JSON
 }
 
+type DeploymentStatus {
+  result: JSON
```
schnie: @samblackk is it possible to define this new type out instead of using `JSON`? Trying to get rid of all the `JSON` types as we refactor. How complex is the nested typing we'd need?

samblackk: This is the return result:

```  {
    "metric":{
      "deployment": "frozen-neutron-5201",
      "instance": "frozen-neutron-5201-statsd.datarouter-frozen-neutron-5201.svc:9102",
      "job": "airflow",
      "type": "counter"
    },
    "value":[
      1562779535,
      "1"
    ]
  }```

src/lib/pubsub/poller/index.js
samblackk: I'm not entirely sure how it worked it in the first place.. but it may be because we returned promises in the other subscriptions? It's not clear, honestly. However, with this update, we're not copying `AsyncIterator` (which is a no-go) and should resolve future issues we would have ran into.

src/schema.graphql
samblackk: Actually - this would be pretty hard to do. The results from PromQL don't always have the same hierarchy or keys in `metric`.

PR #166, opened 2019-11-13
Include a friendly name in the "from" header in emails

ryanahamilton:
Before/after:

image


schnie:
@ryanahamilton looks good! I think it failed a linting check though. Should be able to follow the link and see in drone.

PR #168, opened 2019-11-13
Global Service Accounts

petedejoy:
Add mutations and unit tests for creating and deleting system-level service accounts. These accounts will set the stage to allow us programmatically invite users to various Cloud instances from ATC.

https://github.com/astronomer/issues/issues/553


src/resolvers/mutation/delete-system-service-account/index.js
```diff
@@ -0,0 +1,38 @@
...
+) {
+  // Pull out some variables.
+  const { serviceAccountUuid } = args;
+
+  // Look for the service account.
+  const serviceAccount = await ctx.db.query.serviceAccount(
+    {
+      where: { id: serviceAccountUuid }
+    },
+    `{ roleBinding { workspace { id }, deployment { id } } }`
```
petedejoy: @schnie I'm thinking I can get away with just returning a top-level `id` here- is my suspicion correct?

ashb: Yes, probably can. The return for the delete mutation doesn't matter much

src/resolvers/mutation/create-system-service-account/index.js
```diff
@@ -0,0 +1,41 @@
...
+ * @return {ServiceAccount} The new ServiceAccount.
+ */
+export default async function createSystemServiceAccount(
+  parent,
+  args,
+  ctx,
+  info
+) {
+  // Pull out some variables.
+  const { label, category, role } = args;
```
ashb: We should validate that role starts with `system.`

src/resolvers/mutation/delete-system-service-account/index.js
schnie: Yep, sounds good

src/resolvers/mutation/create-system-service-account/index.js
schnie: @petedejoy could throw a test in for it as well

petedejoy: Just threw in this validation layer and unit test!

src/resolvers/mutation/delete-system-service-account/index.js
petedejoy: Done. Thanks, guys!

src/resolvers/mutation/create-system-service-account/index.js
```diff
@@ -0,0 +1,44 @@
...
+
+  if (!role.includes("SYSTEM")) throw new InvalidRoleError();
```
schnie: @petedejoy maybe we use `startsWith` here to be just a little bit more specific? Or did you copy this from somewhere else?

src/resolvers/mutation/create-system-service-account/index.unit.test.js
```diff
@@ -0,0 +1,105 @@
...
+    const vars = {
+      label: casual.word,
+      category: casual.word,
+      role: WORKSPACE_ADMIN
+    };
+
+    // Run the graphql mutation.
+    const res = await graphql(schema, mutation, null, { db, user }, vars);
+    expect(res.errors).toHaveLength(1);
+    expect(createServiceAccount.mock.calls).toHaveLength(0);
```
andriisoldatenko: a while ago @ashb suggested to write `.toHaveBeenCalledTimes(0)` instead of `.mock.calls).toHaveLength(0);`

andriisoldatenko: also pls add:
```
expect(createServiceAccount).toBeCalledWith(query, ...);```

File: src/resolvers/mutation/create-system-service-account/index.js (hunk 3 above)
Comment (petedejoy) [reply src/resolvers/mutation/create-system-service-account/index.js]: Yeah probably. I didn't copy this from somewhere, just honestly didn't know `startsWith` was a thing 🙂.

File: src/resolvers/mutation/create-system-service-account/index.unit.test.js (hunk 4 above)
Comment (petedejoy) [reply src/resolvers/mutation/create-system-service-account/index.unit.test.js]: On it. Thanks Andrii!

Comment (petedejoy) [reply src/resolvers/mutation/create-system-service-account/index.unit.test.js]: @andriisoldatenko I'm having trouble using `.toBeCalledWith` because the `createServiceAccount` function is called with an `apiKey` arg. Since that is randomly generated within the actual mutation, I'm not sure how to get Jest to match it. Any ideas?

Comment (andriisoldatenko) [reply src/resolvers/mutation/create-system-service-account/index.unit.test.js]: you can use: `expect.Anything()`

Comment (andriisoldatenko) [reply src/resolvers/mutation/create-system-service-account/index.unit.test.js]: @petedejoy e.g.:
```
expect(deployments).toBeCalledWith(
      {
        where: { id, deletedAt: null, api: expect.anything() }
      },
      expect.anything()
    );
```

petedejoy: Hm- makes sense, but if I do that via:
```    const where = { role: vars.role };
    const data = {
      active: true,
      apiKey: expect.Anything(),
      category: vars.category,
      label: vars.label,
      roleBinding: {
        create: { role: SYSTEM_ADMIN }
      }
    };

    expect(createServiceAccount).toHaveBeenCalledWith(
      { where, data },
      expect.any(Object)
    );```

I still get

```    - Expected
    + Received

      Object {
        "data": Object {
          "active": true,
    -     "apiKey": Anything,
    +     "apiKey": "69e9e5ea5b148eb2209623ff020da7a6",
          "category": "recusandae",
          "label": "nostrum",
          "roleBinding": Object {
            "create": Object {
              "role": "SYSTEM_ADMIN",
            },
          },
    -   },
    -   "where": Object {
    -     "role": "SYSTEM_ADMIN",
        },```

petedejoy: @schnie @andriisoldatenko any ideas here? Having trouble figuring out how to get that `expect.anything()` working correctly here and can't seem to circumvent this error.

schnie: I don't think you need the `expect.anything()` down under `{where, data}`, only up in your `data` object.

petedejoy: Hm- do you mean the `expect.any(Object)` there? Removing it produces the same issue- seems like the `expect.anything();` in my `data` object doesn't do what I'd expect it to do and still throws that above error.

schnie: Ah, you may need `expect.objectContaining` as well. Check the example here: https://jestjs.io/docs/en/expect#expectobjectcontainingobject

PR #170, opened 2019-11-15
Change how we define our roles to make it easier to tweak permissions

ashb:
The "config" module we use supports loading config files from multiple
sources, but the behavior with arrays is that you can't merge them, so
if you want to change any of the permissions at all you have to
duplicate the entire block of permissions to make a single change. This
makes upgrades error prone (and puts the onus on the admins of the
platform to merge changes correctly).

So instead this changes the layout form list-of-lists to dict-of-dicts,
as the config layering means we can tweak _exactly_ the permissions we
need, so for instance if you didn't want to allow Editors to push images
you could set:

    roles:
      DEPLOYMENT_EDITOR:
        permissions:
          deployment.images.push: false

and then _just_ that one permission is changed. This is easier to
maintain across upgrades as we add more permissions, and clearer to
understand what the intent is and what is actually changed.

In order to support anyone who currently has a custom role config we
upgrade the old config in memory before using it.


src/lib/rbac/index.js
```diff
@@ -51,15 +64,16 @@ export function hasPermission(user, permission, entityType, entityId) {
...
-  const permissions = flatten(
-    user.roleBindings.map(binding => {
-      const role = find(config.get("roles"), { id: binding.role });
-      if (!role) return false;
-      return filter(role.permissions, p => p.startsWith("system"));
-    })
-  );
 
-  return includes(permissions, permission);
+  for (let binding of user.roleBindings) {
```
andriisoldatenko: i used `const binding of...` @schnie ?

schnie: I do like to standardize on `const` when it can be

src/lib/rbac/index.js
```diff
@@ -229,12 +243,54 @@ export function accesibleDeploymentsWithPermission(user, permission) {
...
+          )
+        }
+      ];
+    })
+  );
+}
+
+// The config module doesn't let us edit the config at runtime, so we can't set
+// this back in place.
+const ROLES = exports.upgradeOldRolesConfig(config.get("roles"));
```
schnie: @ashb could we move this line to the top of the file? I was confused scanning through haha.

schnie: Also, you don't need to prefix with `exports.` right?

src/lib/rbac/index.js
ashb: Oh yes, no need for `let` here.

src/lib/rbac/index.js
ashb: Oh with function hoisting it's probably okay, yeah

ashb: Done.

src/lib/rbac/index.js
ashb: Done.

PR #173, opened 2019-11-16
Add support for changes to labels in the oss airflow chart branch

schnie:
This moves the label handling from the `createDeployment` mutation down to the `generateHelmConfig` sequence. We now apply more generic labels to the airflow chart, rather than the platform-specific platform section.

This should be backward compatible for handling both versions.


schnie:
@ashb its slightly changing the logic so we have more control over some of the airflow helm values that are generated.

This change is to support existing airflow chart versions as well as prepare for merging the oss helm branch. It's to account for the change seen here (repeated on every chart object): https://github.com/astronomer/helm.astronomer.io/pull/213/files#diff-edb7ac86c3bb9a79e12fa35d11cdc53aR10-R19.

PR #174, opened 2019-11-17
Ensure we run prisma deploy before using client library

schnie:
Now that we are running `prisma generate` earlier, the prisma client library is updated by the time we run this migration and query for deployments (and create docker images). This means the client library and schema in the db are out of sync, which causes and issue when migrating from a previous version. This makes sure we run `prisma deploy` before attempting to use the client library.


andriisoldatenko:
LGTM

PR #177, opened 2019-11-21
Expose and airflow chart and image versions

schnie:
This PR re-arranges and adds new configuration to Houston for the default Airflow chart version, as well as the default Image tag and closes https://github.com/astronomer/issues/issues/589.


schnie:
@andriisoldatenko we can consume this new image tag when we `astro dev init` now.

andriisoldatenko:
@schnie do you mean to change astro-cli according to this new query?

schnie:
Yep this is the API portion of what we are discussing over here https://github.com/astronomer/issues/issues/335

src/schema.graphql
```diff
@@ -140,9 +140,11 @@ type DeploymentConfig {
   astroUnit: AstroUnit
   maxExtraAu: Int
   executors: JSON
-  latestVersion: String
   singleNamespace: Boolean!
   loggingEnabled: Boolean!
+  latestVersion: String @deprecated
+  airflowChartVersion: String!
+  airflowImageTag: String!
```
ashb: This one feels slightly out of placed - tag for what image?

ashb: Oh sorry, I thought this was on Deployment, not DeploymentConfig

schnie:
This is tied with this change in the airflow helm chart https://github.com/astronomer/helm.astronomer.io/blob/65c1a61e725476e8a69e79e9cb8fd3861916add0/charts/airflow/values.yaml#L59

PR #179, opened 2019-11-25
Metrics: Update heartbeat rate Prom query

samblackk:
A small adjustment that updates the heartbeat prom query. Needed for new orbit.


schnie:
@samblackk what is this change for and how will affect old orbit if we do a release with it next week?

samblackk:
@schnie This is a non-breaking change for current orbit. Good to go in new release.

PR #180, opened 2019-11-27
Add support for multiple Airflow versions

schnie:
This PR adds support for a few more `DeploymentConfig` properties that show the supported airflow versions, tags and defaults.


schnie:
Gotta fix / add tests.

config/default.yaml
```diff
@@ -313,15 +313,25 @@ roles:
...
-    # Default Airflow image settings
-    image:
-      version: 1.10.5
+  # Airflow chart settings
+  # Static helm configurations for this chart are found below.
+  chart:
+    # This version number controls the default Airflow chart version that will be installed
+    # when creating a new deployment in the system. This is also used to ensure all
+    # child airflow deployments are kept up to date and on the latest version.
+    version: 0.10.3
```
schnie: @andriisoldatenko this is the value the auto-updater will want to look at now.

config/default.yaml
```diff
@@ -313,15 +313,25 @@ roles:
...
+
+  # Airflow image settings
```
schnie: Annoying to maintain here for enterprise installs but cloud can be upgraded on the fly via its own list.

schnie:
@andriisoldatenko @ashb made some changes. Dropped the notion of a "distro" from the platform and added a small concept of "channels". Thinking about how we could introduce new versions for people to play with. It's pretty simple but is a start.

config/default.yaml
andriisoldatenko: @schnie want do you mean here?

andriisoldatenko: astro upgrade is using public github api and current version (tag)
and does not relate to Houston

config/default.yaml
andriisoldatenko: @schnie want do you mean here?
astro upgrade is using public github api and current version (tag)
and does not relate to Houston
https://github.com/astronomer/astro-cli/blob/f0ec475914c85ab14e0145309118ecbe68331872/version/version.go#L49

PR #181, opened 2019-11-27
Feature: Platform view-only users

petedejoy:
This PR addresses the issue summarized here: https://github.com/astronomer/issues/issues/552

We would like to, in certain Enterprise cases, limit all users from creating workspaces on the platform. Since some of our customers will have hundreds of users, it is a security concern for them to give users the ability to consume Kubernetes resources via creating one-off Airflow environments.

With this PR, all users are now given a synthetic `USER` role, with a permission `system.workspace.create` attached to it. The UX will not change by default, but folks installing Astronomer Enterprise can now move that permission onto a `SYSTEM_*` role if they would like to configure the platform such that _only_ users with system-level roles can create new workspaces.


src/lib/rbac/index.js
```diff
@@ -178,12 +178,10 @@ export async function getAuthUser(authorization) {
  * This function wraps the two calls above to append fake roleBindings
  * to the user object for any deployments that belong to workspaces where
  * the user has WORKSPACE_ADMIN role.
- * @param {Promise} promise A proimse for a user or service account.
+ * @param {User} Resolved promise for a user or service account.
```
ashb:
```suggestion
 * @param {User} The user or service account object.
```

petedejoy:
Yeah I was thinking that as well- I didn't see a test in there for `addDeploymentRoleBindings` either, is that just because we're planning on deprecating it with deployment-level RBAC?

Also, I confirmed in Orbit that the feature works, but we probably need to couple this PR with some additional error-handling in the UI. I think this works to merge as an MVP for customers, but will plan on leaving the issue open until I can get a better UX implemented in Orbit

image

petedejoy:
Added some unit tests for our new `addUserRoleBinding` function. Happy to do the same for the existing `addDeploymentRoleBindings` function if you all think it'd be worth it!

config/default.yaml
```diff
@@ -245,6 +245,12 @@ roles:
       ? system.workspace.suspend
       ? system.workspace.extendTrial
       ? system.airflow.admin
+      ? system.workspace.create
+    
+  USER:
+    name: User
+    permissions:
+
```
schnie: @petedejoy don't we want normal users to be able to create workspaces by default?

petedejoy: Ah yeah sorry, changed this to test and forgot to revert back.

config/default.yaml
```diff
@@ -245,6 +245,11 @@ roles:
       ? system.workspace.suspend
       ? system.workspace.extendTrial
       ? system.airflow.admin
+    
+  USER:
```
schnie: @petedejoy can we bump this to the bottom of the list here so it doesn't break up the scoped roles.

database/datamodel.graphql
```diff
@@ -2,6 +2,7 @@ enum Role {
   SYSTEM_ADMIN
   SYSTEM_EDITOR
   SYSTEM_VIEWER
+  USER
```
schnie: and same here

config/default.yaml
petedejoy: Whoops, done!

config/default.yaml
```diff
@@ -245,11 +245,10 @@ roles:
       ? system.workspace.suspend
       ? system.workspace.extendTrial
       ? system.airflow.admin
-
+
```
schnie:
```suggestion

```

config/default.yaml
```diff
@@ -245,11 +245,9 @@ roles:
       ? system.workspace.suspend
       ? system.workspace.extendTrial
       ? system.airflow.admin
```
schnie:
```suggestion
      ? system.airflow.admin

```

PR #182, opened 2019-11-28
Automate Airflow Chart Upgrades

andriisoldatenko:
fix https://github.com/astronomer/issues/issues/577


src/scripts/deployment-upgrade.js
```diff
@@ -0,0 +1,45 @@
...
+async function deploymentUpgrade() {
+  log.info("Starting registry cleanup");
+  const deployments = await prisma.deployments({}, `{ releaseName, version }`);
+  if (deployments.length === 0) {
+    log.info("There is no deployments to delete :(");
+    return;
+  }
+  for (const deployment of deployments) {
+    if (deployment.version !== argv["airflowChartVersion"]) {
+      await prisma.upgradeDeployment({
```
andriisoldatenko: Hi @schnie @ashb 
i have some troubles during testing this script, specifically to run `upgradeDeployment`   
in simple script…. do i need to import it like this `import upgradeDeployment from "../resolvers/mutation/upgrade-deployment";`?

andriisoldatenko: i know that `upgradeDeployment` is mutation, and requires context with commander and prisma...

src/scripts/deployment-upgrade.js
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  for (const deployment of deployments) {
+    if (deployment.version !== argv["airflowChartVersion"]) {
+      const updatedDeployment = await prisma.mutation.updateDeployment(
+        {
+          where: { id: deployment.deploymentUuid },
+          data: { version: argv["airflowChartVersion"] }
+        },
+        `{ releaseNam, version }`
```
schnie:
```suggestion
        `{ releaseName, version }`
```

src/scripts/deployment-upgrade.js
```diff
@@ -0,0 +1,61 @@
...
+import { DEPLOYMENT_AIRFLOW } from "constants";
+
+/*
+ * Upgrade deployments
+ */
+async function deploymentUpgrade() {
+  log.info("Starting automatic deployment upgrade");
+  const deployments = await prisma.deployments({}, `{ releaseName, version }`);
+  if (deployments.length === 0) {
+    log.info("There is no deployments to delete :(");
```
schnie:
```suggestion
    log.info("There are no deployments to upgrade");
```

src/scripts/deployment-upgrade.js
```diff
@@ -0,0 +1,61 @@
...
+import { generateHelmValues } from "deployments/config";
+import yargs from "yargs";
+import { DEPLOYMENT_AIRFLOW } from "constants";
+
+/*
+ * Upgrade deployments
+ */
+async function deploymentUpgrade() {
+  log.info("Starting automatic deployment upgrade");
+  const deployments = await prisma.deployments({}, `{ releaseName, version }`);
```
schnie: @andriisoldatenko I think we want this query to look for all deployments where the `version` in the db is less than the version in this config - https://github.com/astronomer/houston-api/blob/master/config/default.yaml#L319-L322.

We may want to keep the query the same, then filter down if the version on the deployment record is less than the version in this ^ config. Could use the `semver` package that we used in this PR - https://github.com/astronomer/houston-api/pull/180/files.

src/scripts/deployment-upgrade.js
schnie: Yea I was thinking that we could call the mutation directly too, but maybe this is fine.

src/scripts/deployment-upgrade.js
andriisoldatenko: @schnie 
pls correct me if i’m wrong, we are going to respect version in houston config rather than deployment-upgrade.js --airflow-chart-version  flag?

schnie: @andriisoldatenko Yes, I think so because we also need to keep track of what version the platform should use on new deployments. This will keep it in one place.

andriisoldatenko:
@schnie PTAL
practical results:
1) current version in DB:
```
postgres=# select "releaseName","airflowVersion", "version" from  "houston$default"."Deployment";
      releaseName      | airflowVersion | version
-----------------------+----------------+---------
 advanced-quark-6540   |                | 0.10.2
 empty-opposition-6214 |                | 0.10.2
(2 rows)

```
2) Run upgrade script:
`npx babel-node src/scripts/deployment-upgrade.js`

3) 
```
postgres=# select "releaseName","airflowVersion", "version" from  "houston$default"."Deployment";
      releaseName      | airflowVersion | version
-----------------------+----------------+---------
 advanced-quark-6540   |                | 0.10.3
 empty-opposition-6214 |                | 0.10.3
(2 rows)
```

schnie:
Looks good, what about the `helm list` output? Did you run this in a live environment to make sure the `helm upgrade` through commander works as intended?

andriisoldatenko:
Trying right now.

andriisoldatenko:
@schnie practically LGTM.

andriisoldatenko:
@ashb @schnie PTAL

src/scripts/deployment-upgrade.js
```diff
@@ -0,0 +1,59 @@
...
+  for (const deployment of deployments) {
+    if (semver.lt(deployment.version, airflowChartVersion)) {
+      const releaseName = deployment.releaseName;
+      const updatedDeployment = await prisma
+        .updateDeployment({
+          where: { releaseName },
+          data: { version: airflowChartVersion }
+        })
+        .$fragment(
+          `{ id releaseName version extraAu airflowVersion alertEmails workspace { id } }`
```
schnie: @andriisoldatenko what all do we need here. I'm guessing we don't really need `releaseName version extraAu airflowVersion alertEmails`.

andriisoldatenko: thanks!

PR #183, opened 2019-12-02
Add public signups disabled error to oauth route

petedejoy:
This PR addresses the issue summarized here: https://github.com/astronomer/issues/issues/614

This is a problem we run into very often with Enterprise installs, but has fallen through the cracks in the past due to how trivial of a problem it is to resolve. We can also use this new handling to bubble the error up to Orbit in a more attractive way.

Note: I suspect there's a more elegant way of handling the double if statement I have in my code :)


petedejoy:
Hmm, realizing here we'll need to handle exceptions for the first user logging into the platform, which will require a prisma query. Not ready for review just yet.

petedejoy:
@schnie this is tested and ready for review- we've had folks sink far too many hours into trying to find the root cause of this pesky `500 Internal Server Error`.

src/lib/errors/errors.js
```diff
@@ -8,7 +8,7 @@ import {
 export class PublicSignupsDisabledError extends Error {
   message =
     this.message ||
-    "Public signups are disabled, a valid inviteToken is required";
+    "Public signups are disabled, a valid inviteToken is required to login to the platform. Public signups can be enabled via helm.";
```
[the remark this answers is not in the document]
petedejoy: Done!

src/routes/v1/oauth/callback/post.js
```diff
@@ -78,8 +81,18 @@ export default async function(req, res, next) {
...
+    config.get("publicSignups") == false &&
+    !firstUser &&
+    !state.inviteToken
+  ) {
+    throw new PublicSignupsDisabledError();
+  }
+
   // If we just created the user, also create and connect the oauth cred.
-  if (!user) {
+  if (!user && config.get("publicSignups") == true) {
```
schnie:
```suggestion
  if (!user && config.get("publicSignups") === true) {
```

src/routes/v1/oauth/callback/post.js
```diff
@@ -78,8 +81,18 @@ export default async function(req, res, next) {
     });
   }
 
+  // If the user does not exist and public signups are disabled, throw an error
+  if (
+    !user &&
+    config.get("publicSignups") == false &&
```
schnie: @petedejoy how about assigning `config.get("publicSignups")` to a variable right under `firstUser`, then saying `!publicSignups` here and then `&& publicSignups` below that.

petedejoy: Good call- done!

src/routes/v1/oauth/callback/post.js
```diff
@@ -78,8 +82,13 @@ export default async function(req, res, next) {
...
   }
 
+  // If the user does not exist and public signups are disabled, throw an error
+  if (!user && !publicSignups && !firstUser && !state.inviteToken) {
+    throw new PublicSignupsDisabledError();
+  }
+
   // If we just created the user, also create and connect the oauth cred.
-  if (!user) {
+  if (!user && publicSignups) {
```
schnie: @petedejoy is this right? If we have `publicSignups` disabled, this would never right right?

petedejoy: Yeah- does this run for invited users too? Do we ever need it to run if public signups are disabled?

petedejoy: Yeah- does this run for invited users too? Do we ever need it to run if public signups are disabled?

petedejoy: Ahh wait I see- since the function above this will throw the error in the cases we want to catch, it doesn't actually matter whether publicSignups are disabled or not for this function.

PR #184, opened 2019-12-03
Clean up expired trial deployments

andriisoldatenko:
fix https://github.com/astronomer/issues/issues/586


andriisoldatenko:
need practical testing before merging.

src/schema.graphql
```diff
@@ -130,6 +130,7 @@ type Workspace {
   createdAt: DateTime!
   updatedAt: DateTime!
   trialEndsAt: DateTime
+  isSuspended: Boolean
```
ashb: If we are soft-deleting them we don't need this new field.

ashb: Wait I see. But when is this field getting set to true?

andriisoldatenko: you are right, i can't find, let's refactor using `trialEndsAt`

andriisoldatenko:
@ashb @schnie  practically LGTM
```
2019-12-04T15:46:18 INFO Starting soft deletion of expired deployments
2019-12-04T15:46:29 INFO Updating expired deployment advanced-quark-6540
2019-12-04T15:46:35 INFO Updating expired deployment empty-opposition-6214
2019-12-04T15:46:36 INFO Soft deletion of expired deployments has been finished!
Waiting for the debugger to disconnect...
```

```
postgres=# select "releaseName", "deletedAt" from  "houston$default"."Deployment";
      releaseName      |        deletedAt
-----------------------+-------------------------
 advanced-quark-6540   | 2019-12-04 13:46:33.337
 empty-opposition-6214 | 2019-12-04 13:46:35.385
(2 rows)
```

src/scripts/expire-deployments/index.js
```diff
@@ -0,0 +1,48 @@
+/*
+ * Script soft delete expired deployments
```
```suggestion
 * Script to soft delete expired deployments
```

src/schema.graphql
schnie: Agreed - we shouldn't need this. If we have gone past the `trialEndsAt` date, soft-delete. This will start grace-period before the cleanup job comes by and hard-deletes the rest of the deployment data.

```suggestion
```

PR #185, opened 2019-12-04
Hotfix: Verify Email Mutation

petedejoy:
This PR addresses an issue that @paolaperaza and I ran into yesterday in which we couldn't manually verify a new Cloud user's email address using the existing `verifyEmail` mutation.

Upon investigation, we realized that this mutation was only designed to work with oauth. It changed `verified` in the `Emails` table to `true`, but did not cover changing the user status to `active` in the `Users` table. Since the user, for whatever reason, landed on a blank chrome tab when clicking the link in the `Confirm your Email` message she received, we had no way of giving her access to the app.

@schnie - I spun up some fresh unit tests to cover the additions here, but would appreciate if you could take a look to confirm they're good. I tested this and it seems to fix the issue.


src/resolvers/mutation/verify-email/index.js
```diff
@@ -8,11 +11,32 @@
...
   // Update email to be verified.
-  const where = { address: email };
-  const data = { verified: true };
-  await ctx.db.mutation.updateEmail({ where, data });
+  const where = { id: user.id };
+  const data = {
+    emails: {
+      update: { where: { address: email }, data: { verified: true } }
+    },
+    status: "active"
```
schnie: @petedejoy do we have a `constant` or "active" that we could import and use here?

petedejoy: Ahh yeah I think I used it in the unit test. Good catch!

PR #186, opened 2019-12-06
Extend capabilities + add authUserCapabilites

samblackk:
Extends `workspaceCapabilities` and `deploymentCapabilities` to handle more granular permissions server-side. Also adds `authUserCapabilities` to handle `sysAdmin` permissions. 

Ready for merge.

### DIFF
--- a/src/resolvers/types/auth-user/index.js
+++ b/src/resolvers/types/auth-user/index.js
@@ -1,4 +1,5 @@
 import fragment from "./fragment";
+import { hasPermission } from "rbac";
 import { createAuthJWT, setJWTCookie } from "jwt";
 import { addFragmentToInfo } from "graphql-binding";
 import { USER_STATUS_ACTIVE } from "constants";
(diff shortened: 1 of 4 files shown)


src/resolvers/types/auth-user/index.js
```diff
@@ -43,5 +44,28 @@ export function isAdmin() {
...
+ * on a particular deployment.
+ * @param {Object} parent The result of the parent resolver.
+ * @param {Object} args The graphql arguments.
+ * @param {Object} ctx The graphql context.
+ * @return {AuthUserCapabilities} Map of boolean capabilities.
+ */
+export function authUserCapabilities(parent, args, ctx) {
+  const permissions = [
+    {
+      key: "canSysAdmin",
```
petedejoy: I'm wondering if `isSysAdmin` may be better in this case.

src/resolvers/types/workspace/index.js
```diff
@@ -38,23 +38,64 @@ export function deploymentCount(parent) {
...
-  const canUpdateBilling = hasPermission(
-    ctx.user,
-    "workspace.billing.update",
-    ENTITY_WORKSPACE.toLowerCase(),
-    parent.id
-  );
+  const permissions = [
+    {
+      key: "canUpdateBilling",
+      value: "workspace.iam.update"
```
petedejoy: Looks like this one (`canUpdateBilling`) and the one below (`canUpdateIAM`) have their permissions flipped.

src/resolvers/types/auth-user/index.js
schnie: @samblackk whacha think here?

samblackk: Let's keep it `canSysAdmin` for consistency. Keeps things easy to look up in the UI and anywhere else it may be used.

samblackk:
@schnie Can we get this released to datarouter today, please?

PR #187, opened 2019-12-09
improve prisma helth check

andriisoldatenko:
fix issue: 
```
2019-12-09T19:23:21 ERROR request to http://astronomer-prisma:4466/houston failed, reason: connect ECONNREFUSED 10.97.57.136:4466
```


bin/entrypoint
```diff
@@ -11,10 +11,15 @@ update-ca-certificates
...
-echo "Waiting for host: ${HOST} ${PORT}"
-while ! nc -w 1 -z "${HOST}" "${PORT}"; do
-    sleep 1
-done
+# Use prisma unofficial health check https://www.prisma.io/forum/t/health-check-url/5344
+for i in {1..300}; do
+    echo "Waiting for host: ${HOST} ${PORT}"
+    sleep 1;
+    if [ $(curl -LI "${PRISMA__ENDPOINT}/managment" -o /dev/null -w '%{http_code}\n' -s) == "200" ];
+       then exit 0;
```
ashb: I don't think you want exit here -- that would exit the container.

Also can you put the `then` on the previous line please

andriisoldatenko:
close bc it doesn't fix problem.

for history
```
set -e

# In production, we mount the registry TLS certificate
# into /usr/local/share/ca-certificates. This completes
# the certificate installation so we can trust the registry.
echo "Running update-ca-certificates"
update-ca-certificates

HOST=$(echo "${PRISMA__ENDPOINT}" | awk -F// '{print $2}' | awk -F: '{print $1}')
PORT=$(echo "${PRISMA__ENDPOINT}" | awk -F: '{print $3}' | awk -F/ '{print $1}')

# Use prisma unofficial health check https://www.prisma.io/forum/t/health-check-url/5344
for i in {1..300}; do
    echo "Waiting for host: ${HOST} ${PORT}"
    if [ $(curl -LI "${PRISMA__ENDPOINT}/managment" -o /dev/null -w '%{http_code}\n' -s) == "200" ];
       echo "Received 200 OK response from: ${HOST} ${PORT}"
       # Run the original command
       then exec "$@";
    fi;
done;
echo "Timeout after 300 seconds for waiting 200 OK from ${PRISMA__ENDPOINT}/managment"
exit 1
```

bin/entrypoint
ashb: You wan't `break` here.

PR #188, opened 2019-12-10
add kubectl

andriisoldatenko:
related PR https://github.com/astronomer/helm.astronomer.io/pull/264

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,12 +24,16 @@ RUN apk update \
 		python \
 		git \
 	&& apk add --no-cache \
+		curl \
 		bash \
 		netcat-openbsd \
 		nodejs \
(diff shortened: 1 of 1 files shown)


ashb:
What do we need this for?

andriisoldatenko:
@ashb pls check related PR

schnie:
LGTM, @ashb we need `kubectl` so we can run that quick command to check watch prisma status before running the job. Trying to avoid adding another image, at least while we have all these scans/rebuilds/refactorings going on.

It does kind of leak scope, which bothers me, but we avoid another image. What do you think?

schnie:
Also, if we do merge this we'll need to update our prod image in astronomer/astronomer too (until we get our new pipelines).

PR #190, opened 2019-12-11
Hotfix: Exclude cleanup pods in metrics

samblackk:
Attached to https://github.com/astronomer/issues/issues/651 - excludes `airflow-cleanup-pods` in the `deploymentStatus` metric subscription.


schnie:
@samblackk does this hide it from deployment health status as well as the list(s) of running pods?

PR #191, opened 2019-12-12
Feature/analytics

petedejoy:
This PR implements optional usage of the `analytics-node` package by [Segment](https://segment.com). It addresses the issue outlined [here](https://github.com/astronomer/issues/issues/616); the implementation of New Orbit ("Norbit) by @samblackk called for a re-evaluation of how we're tracking events in our Cloud product.

This PR only contains what I've deemed as "Core Events" for downstream marketing/product/CS processes. I'll work on getting the "Additional Events" roped in tomorrow, but figured I'd open this up for feedback as I had to do quite a bit of wrangling with unit tests. You can find a list of all of the events we wish to track here: https://docs.google.com/spreadsheets/d/1_5cdsL4EtKk09r-wyxAJwthgzUY6Jq7gMSA0xXX0Sxg/edit#gid=639423297


config/test.yaml
```diff
@@ -89,3 +89,8 @@ deployments:
 # Stripe configuration.
 stripe:
   enabled: true
+
+# Analytics configuration
+analytics:
+  enabled: false
+  writeKey: "test"
```
schnie:
```suggestion
  writeKey: "test"

```

src/routes/v1/registry/events/post.js
```diff
@@ -90,6 +93,15 @@ export default async function(req, res) {
         },
         rawConfig: JSON.stringify(generateHelmValues(updatedDeployment))
       });
+
+      // Run the analytics track event
+      track(req.body.user.id, "Deployed Code", {
```
petedejoy: @schnie is this right? Looks like that's how it comes down when running the unit test, but not quite sure how to test without enabling commander and actually making a deploy 🤔

schnie: @petedejoy its showing there because you added it to the body of the request in your test. In a real world scenario, the `user` object on the `request` is set by our auth middleware here: https://github.com/astronomer/houston-api/blob/master/src/lib/authentication/index.js#L23.

petedejoy: Ahh got it. So I want to add the user object directly to the request on my unit test I guess- let me give that a shot.

src/routes/v1/registry/events/post.js
```diff
@@ -90,6 +93,17 @@ export default async function(req, res) {
         },
         rawConfig: JSON.stringify(generateHelmValues(updatedDeployment))
       });
+
+      // Run the analytics track event
+      // TODO: User ID below is written to get around unit test. We'll need to add some test
```
schnie: @petedejoy is this TODO still true?

petedejoy: Nope- good catch!

src/resolvers/mutation/update-workspace/index.js
```diff
@@ -8,7 +8,7 @@ import { addFragmentToInfo } from "graphql-binding";
  * @param {Object} ctx The graphql context.
  * @return {User} The updated User.
  */
-export default function updateWorkspace(parent, args, ctx, info) {
+export default async function updateWorkspace(parent, args, ctx, info) {
```
schnie: @petedejoy do we need this `async`?

PR #192, opened 2019-12-13
Docs for developing with telepresence

verdverm:


README.md
```diff
@@ -54,6 +54,67 @@ Houston is written in ES6 and beyond, and it's currently built with [Babel](http
...
+
+You will need to make some edits:
+
+#### Edit the database conection string to remove the "svc.cluster.local"
+
+```bash
+kubectl get secret -n tester astronomer-bootstrap -o yaml
+
+echo -n "base64-string" | base64 -d   # or -D if on mac
+echo -n "postgres://postgres:<password>@<postgres-service>.<namespace>:5432"
```
schnie: @verdverm should we add this to spit out the encoded version to paste into the next command?

```suggestion
echo -n "postgres://postgres:<password>@<postgres-service>.<namespace>:5432" | base64
```

verdverm: yup, i meant to have that there

schnie: Can you merge that change in? I can't click merge on your branch.

verdverm:
@schnie done, needs your review before merge

README.md
```diff
@@ -54,6 +54,86 @@ Houston is written in ES6 and beyond, and it's currently built with [Babel](http
...
+```
+
+Any argument will cause the cluster config dir to be mounted.
+This will also set the NODE_ENV to the argument value.
+You can set the NODE_ENV yourself to select a different, local config value
+if you do not want to use the cluster versions.
+Additionally, both methods support ENV_VAR overrides in wither mode.
+
+
 ## Live testing in Kubernetes
```
schnie:
```suggestion
## Testing Airflow Chart Upgrades in Kubernetes
```

schnie: Renaming this to be more accurate now that we have telepresence docs.

scripts/volumes.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
schnie: @verdverm we could probably put this in `bin` directory with our other shell script.

PR #197, opened 2019-12-20
Add support for deploying static configs to all deployments 

schnie:
This PR allows us to generate and deploy upgrades to all airflow deployments on _every_ platform upgrade. This lets us apply static updates to `.Values.astronomer.houston.config.deployments.helm` and roll them out immediately.

If there are no actual changes in the generated values, helm won't touch the kubernetes objects making it an idempotent process.

I'm not sure how this scales to hundreds of deployments, especially if we're deploying many times a day. If it becomes an issue we can turn the autoupgrader off and only enable when we know that we want to roll a change out to the deployments, then disable again.

Having this controlled programatically feels a little bit weird, given the rest of our cloud configuration stack, but this feels like a decent way to deal with this problem for now. Maybe we branch out from this for cloud some day, but this feels reasonable for most clusters for now.

What do you guys think?

For reference, here's the helm configuration for this `Job` https://github.com/astronomer/helm.astronomer.io/blob/master/charts/astronomer/templates/houston/houston-upgrade-deployments-job.yaml.

Closes https://github.com/astronomer/issues/issues/682


schnie:
@ashb yep exactly.

PR #200, opened 2020-01-02
Add support for hiding service accounts after some time

schnie:
This PR adds support for hiding API Keys after x minutes. This is to be backward compatible with old Houston and Orbit functionality.

Closes https://github.com/astronomer/issues/issues/696


schnie:
Going to go ahead and merge in for an RC release today.

PR #201, opened 2020-01-07
Unified Tracking

petedejoy:
This PR complements the [Analytics PR](https://github.com/astronomer/houston-api/pull/191). It allows us to auto-invite users to Cloud instances via a global service token via the following changes:
- The `inviteUser` mutation now returns the invite `id`- before it returned nothing. ATC then uses that `id` as the `userId` in its `analytics.identify` call.
- When we create the user, we check if they have an invite. If so, we use their invite `id` as their user `id` when creating the database record.  This way, the users we've already identified from ATC will have all subsequent events mapped to their user record.
- If a user does not have an invite, we create the user id automatically via Prisma.


src/lib/users/index.unit.test.js
```diff
@@ -166,6 +170,9 @@ describe("userExports.validateInviteToken", () => {
   });
 
   test("throws if token is not found", async () => {
+    jest.spyOn(prismaExports.prisma, "inviteToken").mockReturnValue({
```
petedejoy: @schnie this fixed the issue I was seeing with all three unit tests (this one and the two immediately below it), but I am not so sure why 🤔

src/lib/users/index.js
```diff
@@ -55,21 +61,38 @@ export async function createUser(opts) {
...
-    fullName,
-    emails: {
-      create: {
-        address: email,
-        primary: true,
-        verified: !!(!emailConfirmation || haveInvite),
-        token: emailToken
+  // Create our base mutation. Use the invite id if a user has an invite
+  // else, use the default random prisma id (for analytics tracking)
+  const mutation = haveInvite
```
schnie: @petedejoy since most of this is duplicated, except for `id: inviteId`. Could we set up an object with all the common fields and then `if (haveInvite) mutation.id = inviteId;`

src/lib/users/index.unit.test.js
```diff
@@ -29,6 +29,7 @@ describe("userExports.createUser", () => {
   jest
     .spyOn(prismaExports.prisma, "usersConnection")
     .mockImplementation(usersConnection);
+  // Set by each test case
```
schnie: @petedejoy is this an old comment?

src/lib/users/index.js
petedejoy: @schnie - just fixed this up, can you take a look and confirm it's what you're thinking?

src/lib/users/index.js
```diff
@@ -37,6 +37,12 @@ export async function createUser(opts) {
     throw new PublicSignupsDisabledError();
   }
 
+  // Return the invite ID from the database to use in the
+  // user record
+  const inviteId = haveInvite
```
schnie: @petedejoy what do you think about refactoring this logic a bit and combining with the `if` statement we have below. They are essentially running the same check in two places, so we could condense.

petedejoy: Ah good catch! For whatever reason I was thinking we had to keep them separate because of the order of operations here- just pushed up a commit that does away with `inviteId` entirely and pulls the logic down to the conditional below. Check it out when you have a sec.

PR #204, opened 2020-01-09
Add check for customerStripeId in expire-deployments script

schnie:
This PR adds an additional check for expiring deployments. We need to also only return deployments that do not have a stripe customer id associated.

It also adds support for a `--dry-run` flag to only print the deployments it would soft-delete, instead of actually going through with it, and more logging.

I'm thinking that we may want to enable this `--dry-run` flag when we initially deploy and then watch it run a few times, before enabling it.

I'll follow up with a helm PR to make this a configuration.


schnie:
Also added `--dry-run` flag to cleanup job.

PR #205, opened 2020-01-09
Add `system.workspace.delete` permission

petedejoy:
This PR addresses the issue raised in https://github.com/astronomer/issues/issues/685.

We now have a `system.workspace.delete` permission on the `SYSTEM_ADMIN` role that is allowed to run the `deleteWorkspace` mutation.


paolaperaza:
Thanks for this, @petedejoy + @schnie !

PR #207, opened 2020-01-10
fix issues when registry/commander rebooting

andriisoldatenko:
>Hitting a situation on v0.10.3 -> v0.11.0 upgrade where the DockerImages migration fails and leaves the migration table locked, because the registry pod is rebooting, which it needs to query. Also seems like the upgrader helm hook fails trying to make calls to commander for a similar reason, commander is / could be rebooting during it.


bin/entrypoint
```diff
@@ -8,14 +8,22 @@ set -e
...
+}
 
-echo "Waiting for host: ${HOST} ${PORT}"
-while ! nc -w 1 -z "${HOST}" "${PORT}"; do
-    sleep 1
-done
-echo "Received response from: ${HOST} ${PORT}"
+PRISMA_HOST=$(echo "${PRISMA__ENDPOINT}" | awk -F// '{print $2}' | awk -F: '{print $1}')
+PRISMA_PORT=$(echo "${PRISMA__ENDPOINT}" | awk -F: '{print $3}' | awk -F/ '{print $1}')
+REGISTRY_PORT=5000
```
schnie: @andriisoldatenko should we make this an env var as well? We do pass it into houston as a config here https://github.com/astronomer/houston-api/blob/master/config/default.yaml

bin/entrypoint
```diff
@@ -8,14 +8,22 @@ set -e
...
+REGISTRY_HOST="${HELM__RELEASE_NAME}-registry.${HELM__RELEASE_NAMESPACE}"
```
schnie: @andriisoldatenko what do you think about passing in $REGISTRY_HOST as a single env var, instead of the two HELM ones?

andriisoldatenko: yeah make sense

bin/entrypoint
andriisoldatenko: fixed.

PR #208, opened 2020-01-13
implement migration created date for docker images

andriisoldatenko:
> Ran into an issue with showing the currentTag to the users. The Houston query returns the latest DockerImage records tag which works fine for all future pushes. The problem is that our migration that queries the registry and backfills the DockerImage table gives them all essentially the same createdAt date and the order that the registry API returns them does not appear to be sorted by any date either. (edited) 
So I had pushed 3 times and it was showing cli-2 as my latest after the upgrade and migration ran.
We could just kill the migration and only show for future pushes as an easy fix. Or we could pull some additional metadata about the tags and also migrate the createdAt field into DockerImages

**practical testing:**

```
2020-01-13T14:39:04 INFO Found 1 deployments to migrate: mathematical-parallax-0827
2020-01-13T14:39:04 DEBUG Migrating mathematical-parallax-0827...
2020-01-13T14:39:04 DEBUG Requesting docker tags for mathematical-parallax-0827 at http://localhost:5001/v2/mathematical-parallax-0827/airflow/tags/list
(node:14742) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
2020-01-13T14:39:04 INFO Fetching manifest from http://localhost:5001/v2/mathematical-parallax-0827/airflow/manifests/cli-1
2020-01-13T14:39:04 DEBUG Fectching image config from http://localhost:5001/v2/mathematical-parallax-0827/airflow/blobs/sha256:5e1ce6debcf243c4ff842586d508aaae1590c774b98209f12eab05b030b8ebc8
Batch 2 run: 1 migrations

```

```
postgres=# select "name","createdAt" from "houston$default"."DockerImage" ;
                   name                   |        createdAt
------------------------------------------+-------------------------
 mathematical-parallax-0827/airflow:cli-1 | 2020-01-13 11:28:15.965
(1 row)

```


database/datamodel.graphql
```diff
@@ -141,5 +141,5 @@ type DockerImage {
   tag: String!
   digest: String!
   deployment: Deployment! @relation(name: "DeploymentDockerImages" onDelete: CASCADE)
-  createdAt: DateTime! @createdAt
+  createdAt: DateTime!
```
ashb: If we go this way (rather than the data import approach we're taking a look at) we should add a comment saying why we aren't using the `@createdAt` decorator

src/routes/v1/registry/events/post.js
```diff
@@ -146,6 +148,7 @@ export async function extractImageMetadata(ev) {
   return {
     labels: imageMetadata.config.Labels,
     env: imageMetadata.config.Env,
-    digest: manifest.config.digest
+    digest: manifest.config.digest,
+    created: imageMetadata.created
```
ashb: This return value isn't used is it?

andriisoldatenko: https://github.com/astronomer/houston-api/blob/0495279ef6f5e0ad162a8f1f4c34bd3134b0fb87/src/migrations/20191101143903_migrate_initial_docker_image.js#L94-L106

ashb: Ah gotcha, I was only looking in the event handler.

PR #210, opened 2020-01-15
Add --canary flag to cron scripts

schnie:
This PR makes it possible to pass a --canary flag to any of our cron scripts to only execute them on deployments in the database that have the canary boolean set to true in the database.


src/scripts/expire-deployments/index.js
```diff
@@ -22,17 +22,27 @@ async function expireDeployments() {
...
+      workspace: {
+        stripeCustomerId: null, // And where the workspace does not yet have a stripe customer id
+        trialEndsAt_lte: expireDate // And where the workpace trial end date has been exceeded
+      }
+    }
+  };
+
+  if (argv["canary"]) {
+    log.info(`Limiting search to canary deployments`);
+    query.where = { ...query.where, canary: true };
```
ashb: Any reason not just `query.where.canary = true`?

```suggestion
    query.where.canary = true;
```

schnie: Nope, copy/pasted.

schnie:
@sjmiller609 yes it is worth it. Could you or @andriisoldatenko knock that out? I'll be out for a little bit and we need to roll this out ASAP.

PR #213, opened 2020-01-20
Generate fernetKey in Houston, rather than letting Helm generate it

schnie:
This PR addresses the manual patch from last night where we had to go add the `fernetKey` to the airflow chart values. We are currently not generating this in Houston, so Helm creates it here - https://github.com/astronomer/helm.astronomer.io/blob/master/charts/airflow/templates/secrets/fernetkey-secret.yaml#L23.

This means when Commander does the `helm delete` + `helm install`, we loose this value and generate a new one, breaking the existing airflow database until we restore the old key from velero backups.


src/resolvers/mutation/create-deployment/index.js
```diff
@@ -78,16 +78,21 @@ export default async function createDeployment(parent, args, ctx, info) {
...
-  const elasticsearchPassword = crypto.randomBytes(16).toString("hex");
+  const elasticsearchPassword = generatePassword({ length: 32, numbers: true });
   const hashedElasticsearchPassword = await bcrypt.hash(
     elasticsearchPassword,
     10
   );
 
+  // Generate a random fernetKey and base64 encode it for this deployment.
+  const fernetKey = new Buffer(
+    generatePassword({ length: 32, numbers: true })
```
andriisoldatenko: @schnie few notes:
```
In [1]: from cryptography.fernet import Fernet
   ...: fernet_key= Fernet.generate_key()
   ...: print(fernet_key.decode())
S70k0vV4DjuxuMM0wnUcYXccIKEIyboORAItUtBIFBw=
len(fernet_key.decode())
```

but in nodejs:
```
> var generator = require('generate-password');
undefined
> generator({ length: 32, numbers: true })
Thrown:
TypeError: generator is not a function
> generator.generate({ length: 32, numbers: true });
'fAY3MKxqJGaRW831c6UmZFnjQHSFzFGx'
> generator.generate({ length: 32, numbers: true }).lenght;
undefined
> generator.generate({ length: 32, numbers: true }).length;
32
>
```

schnie: @andriisoldatenko check out the import at the top. I'm renaming the import as `generatePassword`, so its essentially doing what you are doing there. And the `new Buffer().toString()` takes care of base64 encoding it.

andriisoldatenko: @schnie well done!
```
> new Buffer(generator.generate({ length: 32, numbers: true })).toString("base64");
'WHFIeG5xQmlUN2tsOThlaEh2Rm1HcDhCbDJMR1hZZmU='

```

PR #214, opened 2020-01-20
Return the deployment config column in prisma queries

schnie:
This requests the `config` column from the db so the `upgradedDeployment` will contain the user-defined overrides, like resources, executor, etc.

This fixes why we saw reduced pgbouncer limits and lower requests on some components when we ran  the `upgrade-deployments --canary`


schnie:
These look like the key areas that would have been missed when generating new helm values, with out the presence of the `config` column.
https://github.com/astronomer/houston-api/blob/master/src/lib/deployments/config/index.js#L134
https://github.com/astronomer/houston-api/blob/master/src/lib/deployments/config/index.js#L155
https://github.com/astronomer/houston-api/blob/master/src/lib/deployments/config/index.js#L500

PR #216, opened 2020-01-22
Add support for containerd image pull authentication

ianstanton:
- Add fix for failed image pull for airflow deployments when using kubernetes nodes running `containerd`
- `containerd` expects a 401 response with a "WWW-Authenticate" header to inform it whether to send a message with a basic or bearer token (more information here: containerd/containerd#3291 (comment))
- This is required for IBM Kubernetes support
- Testing on EKS, GKE, AKS to be sure it doesn't affect functionality
- Related to https://github.com/astronomer/issues/issues/726


schnie:
Looking at the spec here: https://tools.ietf.org/html/rfc6750#section-3, it looks like we may want to also specify some additional information such as `realm`, `error` and `error_description`.

> All challenges defined by this specification MUST use the auth-scheme
   value "Bearer".  This scheme MUST be followed by one or more
   auth-param values.

Looking at the next section, I'm also wondering if we need to review all the errors here to ensure we're sending the most appropriate code.

I'm also wondering about that `JSON.stringify` array of errors in `sendError`. I'm not sure why we need that.

ianstanton:
Tested on EKS :heavy_check_mark:

vparekh94:
- Deployed platform on GKE (with logging turned off for resource considerations)
- Was able to create deployments
- Was able to push code successfully
- Was able to switch executors successfully

@ianstanton @schnie anything else you would do to test something like this?

ianstanton:
Thanks @vparekh94! Testing on AKS this morning.

ianstanton:
- Deployed platform on AKS (with logging turned off for resource considerations) :heavy_check_mark: 
- Was able to create deployments :heavy_check_mark: 
- Was able to push code successfully (multiple times) :heavy_check_mark: 
- Edited deployment Configuration :heavy_check_mark: 
- Was able to switch executors successfully :heavy_check_mark:

src/routes/v1/registry/authorization/handler.js
```diff
@@ -44,12 +44,10 @@ export default async function(req, res) {
...
-    return sendError(
-      res,
-      REGISTRY_CODES.UNAUTHORIZED,
-      "No authorization credentials specified",
-      403
-    );
+    return res
+      .status(401)
+      .set("WWW-Authenticate", "Bearer")
+      .end();
```
ashb: Is there a reason we can't keep the sendError format (of the JSON response) but also include the WWW-Authenticate header?

So something like this, where we add an extra (optional) headers argument to `sendError`:

```js
    return sendError(
      res,
      REGISTRY_CODES.UNAUTHORIZED,
      "No authorization credentials specified",
      401,
      { "WWW-Authenticate": "Bearer" }
    );
```

schnie: @ianstanton what do you think?

ianstanton: I can test this out

verdverm:
We ought to be able to test containerd support on GKE by creating a node pool that uses that runtime. GKE has 3 runtimes:

1. docker (default)
2. containerd
3. containerd + gvisor

src/routes/v1/registry/authorization/handler.js
ashb: It _should_ be okay, from the linked containerd issue:

```
curl -Ssv https://docker.pkg.github.com/v2/csantanapr/docker/knative-samples_helloworld-go/manifests/latest
> GET /v2/csantanapr/docker/knative-samples_helloworld-go/manifests/latest HTTP/1.1
> Host: docker.pkg.github.com
> User-Agent: curl/7.54.0
> Accept: */*
>
< HTTP/1.1 401 Unauthorized
< WWW-Authenticate: Basic realm="GitHub Docker Registry"
< Server: GitHub.com
* Connection #0 to host docker.pkg.github.com left intact
{"errors":[{"code":"UNAUTHORIZED","message":"GitHub Docker Registry needs login"}]}
```

ianstanton: Tested and this works as well. Updated with your suggestion.

src/routes/v1/registry/authorization/handler.js
```diff
@@ -48,7 +48,8 @@ export default async function(req, res) {
       res,
       REGISTRY_CODES.UNAUTHORIZED,
       "No authorization credentials specified",
-      403
+      401,
+      { "WWW-Authenticate": "Bearer" }
```
ashb: This argument doesn't do anything :)

```
export function sendError(res, code, message, http_err) {
```

needs to be updated to take and handle

```
export function sendError(res, code, message, http_err, headers) {
```

ianstanton: @ashb Thanks for the heads up. Would something like this work, or do we need to handle for when `headers` is null?

```
export function sendError(res, code, message, http_err, headers) {
  return res
    .status(http_err || 401)
    .set("Content-Type", "application/json")
    .set(headers)
    .end(
      JSON.stringify({
        errors: [
          {
            code,
            message,
            details: []
          }
        ]
      })
    );
}
```

schnie: @ianstanton does `.set()` allow you to pass an object of key/values like that?

ashb: Yes, https://expressjs.com/en/api.html#res.set.

I would probably do `.set(headers || {})` which should work I think.

ashb: https://github.com/expressjs/express/blob/4.x/lib/response.js#L749-L774

ashb:
Looks like you have some tests to update @ianstanton

ianstanton:
@ashb is it safe to update the expected status code to `401` in the failed tests? I feel that this test in particular is no longer a valid test if we do so:

https://github.com/astronomer/houston-api/blob/9d44b1545823aed05b1a7262f6367b8a78e3109e/src/routes/v1/registry/authorization/index.unit.test.js#L370-L376

PR #218, opened 2020-01-24
Shorten service acount email address for FAB

schnie:
This is a quick PR to shorten the email addresses of our service accounts in the JWT passed to airflow. 

FAB limits email addresses to `varchar(64)` so longer `baseDomain` values generate service account emails that exceed this, including our cloud and Truata recently.

FAB DDL - https://github.com/dpgaspar/Flask-AppBuilder/blob/98b1be8b3390cd592dc20f215062e55d27e08eec/bin/sqlite_upgrade_1.3.sql#L7

I think this PR is also backward compatible since we key on the userid, not the email in our security manager - https://github.com/astronomer/astronomer-fab-securitymanager/blob/master/src/astronomer/flask_appbuilder/security.py#L163.

This means if a service account request come with a new email in the JWT, it will be updated, and not cause any issues.


schnie:
Old - ck5mmd97q00ke0739yj0azwes@service-accounts.gcp0001.us-east4.astronomer.io = 73
New - ck5mmd97q00ke0739yj0azwes@sa.gcp0001.us-east4.astronomer.io = 59

schnie:
FAB having varchar(64) for emails doesn't seem right though. @kaxil @ashb do you guys think we should PR to FAB or something?

kaxil:
> FAB having varchar(64) for emails doesn't seem right though. @kaxil @ashb do you guys think we should PR to FAB or something?

We should definitely PR in FAB. I am on holiday for the next week but happy to PR if you guys haven't already done it by that time

kaxil:
What would happen for the cases where user_id and email are already saved. Will this try to overwrite that email id?

schnie:
@kaxil my understanding after looking at our security manager is that we key on the user id, so if the email changes, it'd just be updated in the airflow db.

kaxil:
> @kaxil my understanding after looking at our security manager is that we key on the user id, so if the email changes, it'd just be updated in the airflow db.

We should test this out on staging because I feel if it does not overwrite, it would fail with user_id already exists or unique key violation. If it does overwrite we are fine.

schnie:
The relevant part of the security manager is here https://github.com/astronomer/astronomer-fab-securitymanager/blob/master/src/astronomer/flask_appbuilder/security.py#L163.

This part of Houston sets `sub` to be the `userId` (32 char id) - https://github.com/astronomer/houston-api/blob/master/src/routes/v1/authorization/handler.js#L76. 

I'm thinking that since the userId wouldn't be changing so that `else` block would be hit and the user would be updated. It doesn't look like it updates the email field though. :thinking:

schnie:
The other factor here is that this hasn't been deployed to any enterprise customers yet, and it would have never worked for cloud, so we kind of got lucky that there should be no service-account users out there in the world yet.

schnie:
Worth noting that actual users who's emails are greater than 64 chars would trigger this same issue. I feel like we need to fix FAB ASAP to address this. We'll hit this next I'm sure.

PR #219, opened 2020-01-30
Card Mutation & Query: Add fallback for company

samblackk:
Fix https://github.com/astronomer/astro-ui/issues/306

There is a bug where there isn't a fallback for adding the `company` field properly during `addCard()` mutations. This fixes that up and allows backwards compatibility. 

Ready for merge.


ryanahamilton:
@schnie mind reviewing/merging this one?

schnie:
@samblackk @ryanahamilton what exactly happened that requires this change?

samblackk:
@schnie In current orbit we got around this by sending the fallback from Orbit. In `astro-ui` we're not sending the fallback as it should happen in Houston.

schnie:
@samblackk so this is to make Houston work for Orbit and Astro UI, but after Orbit is dead, we can just use the new path?

samblackk:
Yep!

schnie:
Since we're moving forward with Astro UI, I'm wondering if we should just support Astro UI here, since they'll ship together.

ryanahamilton:
@schnie this update _is_ what is needed in order to support Astro UI. It just doesn't break with the current Orbit because this additional fallback logic is handled by the client, but should be moved (by this PR) to the the API. We aren't going out of the way to support the old Orbit in any way.

PR #220, opened 2020-02-05
Replace orbit with astro-ui

schnie:
This PR removes the word "orbit" from the codebase. No functional differences here.


src/lib/users/index.js
```diff
@@ -82,7 +82,7 @@ export async function createUser(opts) {
   if (emailToken != null) {
     sendEmail(email, "confirm-email", {
       token: emailToken,
-      orbitUrl: orbit(),
+      UIUrl: ui(),
```
ashb: We should possibly call this fn `uiUrl()`

schnie:
Incoming PR for astronomer/astronomer

src/lib/users/index.js
schnie: Debated this internally as well haha. I noticed Polidea went the same route with `DAGID`, by capitalizing the first acronym here https://github.com/PolideaInternal/airflow/pull/653/files#diff-fe030a7c1568b3decf599edf399be7f4R784

PR #221, opened 2020-02-05
Change date modifier from 0 to 7

petedejoy:
This PR should cause the cleanup script to look for deployments that are 1 week past their trial end date. This is intended to allow Cloud users to input payment information and continue usage for 1 week after their trial ends, as they are paywalled in the app and prompted to do so.


src/scripts/expire-deployments/index.js
```diff
@@ -98,7 +98,7 @@ const argv = yargs
   })
   .option("date-modifier", {
     alias: "m",
-    default: 0,
+    default: 7,
     description:
       "Add or subtract days from the expiration date, which defaults to now",
```
```suggestion
      "Add or subtract days from the expiration date, which defaults to 7 days from the expiration date",
```

src/scripts/expire-deployments/index.js
```diff
@@ -98,9 +98,9 @@ const argv = yargs
   })
   .option("date-modifier", {
     alias: "m",
-    default: 0,
+    default: 7,
     description:
-      "Add or subtract days from the expiration date, which defaults to now",
+      "Add or subtract days from the expiration date, which defaults to 7 days from the expiration date",
```
```suggestion
      "Add or subtract days from the expiration date",
```

PR #222, opened 2020-02-06
Update alert URLs with new routing

ryanahamilton:
Addresses [#757](https://github.com/astronomer/issues/issues/757).

Updates alert emails with the new Astro UI route patterns. 

These links, before _and_ after this PR are broken (as I outlined in the issue) because the value for `alert.labels.workspace` isn't coming through. This needs to be addressed separately.


schnie:
Documented the second issue here: https://github.com/astronomer/issues/issues/764

PR #225, opened 2020-02-10
Check for platform updates periodically in houston

andriisoldatenko:
fix https://github.com/astronomer/issues/issues/750
related pr https://github.com/astronomer/updates.astronomer.io/pull/1


src/scripts/check-platform-updates/index.js
```diff
@@ -0,0 +1,78 @@
...
+        });
+      }
+    }
+  }
+}
+
+const argv = yargs
+  .option("update-service-url", {
+    alias: "url",
+    default: "https://updates.astronomer.io/certified-airflow",
```
schnie: @andriisoldatenko @ashb, how should we differentiate the paths here? `/astronomer-platform` and `/certified-airflow`? Should we use `.json` on them?

@andriisoldatenko can you also come up with the corresponding file for the platform versions here? We can backfill it with 0.11.x versions to start off.

andriisoldatenko: @schnie nice catch! 
@ashb are you going to add 2 sub-paths? or how can i differentiate it?

src/scripts/check-platform-updates/index.js
```diff
@@ -0,0 +1,80 @@
...
+import "dotenv/config";
+import log from "logger";
+import { prisma } from "generated/client";
+import request from "request-promise-native";
+import { size } from "lodash";
+import yargs from "yargs";
+
+exports.getPlatformReleases = getPlatformReleases;
+
+async function getPlatformReleases(uri) {
```
ashb:
```suggestion
export async function getPlatformReleases(uri) {
```

instead of `exports.getPlatformReleases`

src/scripts/check-platform-updates/index.js
ashb: Yes, I was thinking we'd have `https://updates.astronomer.io/certified-airflow` and `https://updates.astronomer.io/astronomer-platform`.

We can add .json if we want. WDYT?

src/scripts/check-platform-updates/index.js
andriisoldatenko: done.

src/scripts/check-platform-updates/index.js
andriisoldatenko: i don't think `.json` is a good idea, better to use more canonical URIs
sgtm `https://updates.astronomer.io/certified-airflow`

andriisoldatenko: @schnie do i need to load all in 1 table on one table per sub-path?

schnie: We only need to store the `astronomer-platform` response in Houston.

src/scripts/check-platform-updates/index.js
ashb: Did it help the export issue?

database/datamodel.graphql
```diff
@@ -144,3 +144,13 @@ type DockerImage {
   deployment: Deployment! @relation(name: "DeploymentDockerImages")
   createdAt: DateTime! @createdAt
 }
+
+type PlatformVersion {
+  id: ID! @id
+  version: String @unique
+  url: String
+  level: String
```
ashb: My example in the python plugin didn't have it, but I also allowed a `description` field. See the schema in the updates repo.

src/resolvers/query/platform-versions/index.js
```diff
@@ -0,0 +1,13 @@
...
+
+/*
+ * Get list of available platform versions.
+ * @param {Object} parent The result of the parent resolver.
+ * @param {Object} args The graphql arguments.
+ * @param {Object} ctx The graphql context.
+ * @return {[]PlatformVersion} List of PlatformVersions.
+ */
+export default async function platformVersions(parent, args, ctx, info) {
+  return ctx.db.query.platformVersions({}, addFragmentToInfo(info, fragment));
```
ashb: We shouldn't have `addFragmentToInfo` here. That forces fields to be returned (which we sometimes need on some types for other resovlers) but in this case we should leave it entirely up to the callers control

```suggestion
  return ctx.db.query.platformVersions({}, info);
```

And then we can delete the new fragment.js

src/schema.graphql
```diff
@@ -228,6 +228,17 @@ type DeploymentCapabilities {
...
 }
 
+type PlatformVersion {
+  version: String!
+  url: String
+  level: String
+  releaseDate: DateTime!
+}
+
+type PlatformVersions {
```
ashb: This type is (rightly) unused and can be removed.

src/scripts/check-platform-updates/index.js
```diff
@@ -0,0 +1,75 @@
...
+  return versions;
+}
+
+export async function updatePlatformVersions(uri) {
+  log.info("Starting update platform versions");
+  // Build the deployment query.
+  const newReleases = await getPlatformReleases(uri);
+
+  // Find the platform versions from db
+  const releases = await prisma.platformVersions({});
```
ashb: Prisma has built-in support for upserts:https://www.prisma.io/docs/1.34/prisma-client/basic-data-access/writing-data-JAVASCRIPT-rsc6/#creating-or-updating-objects-upserts so we can do this without having to fetch all the rows:

```js
await prisma.upsertPlatformVersion({
    where: { version: newRelease.version },
    create: newRelease,
    update: newRelease
)
```

(Using `newRelease` un-audited like that might be a bad idea, and listing the fields explicitly is probably better)

src/scripts/check-platform-updates/index.js
```diff
@@ -0,0 +1,75 @@
...
+      }
+    }
+  }
+}
+
+const argv = yargs
+  .option("update-service-url", {
+    alias: "url",
+    default: "https://updates.astronomer.io/astronomer-platform",
+    description: "Filter soft deleted deployments older than",
```
ashb: Update please :)

database/datamodel.graphql
andriisoldatenko: done

src/scripts/check-platform-updates/index.js
andriisoldatenko: yes

src/schema.graphql
andriisoldatenko: removed.

andriisoldatenko:
@ashb @schnie practical tests:
create `test.json`
```
{
  "version": 5,
  "available_releases": [
    {
      "version": "0.0.1",
      "url": "https://astronomer.io/cea/release-notes/{new_version}.html",
      "level": "bug_fix",
      "release_date": "2018-11-13T20:20:39+00:00"
    },
    {
      "version": "0.0.2",
      "url": "https://astronomer.io/cea/release-notes/{new_version}.html",
      "level": "bug_fix123",
      "release_date": "2018-11-13T20:20:39+00:00"
    }
  ]
}
```

```
run python -m http.server 8000
```

```
$ houston-api git:(check-for-platform-updates-periodically-in-houston) ✗ npx babel-node /Users/andrii/workspace/src/github.com/astronomer/houston-api/src/scripts/check-platform-updates/index.js --url http://0.0.0.0:8000/test.json
2020-02-11T00:36:44 INFO Starting update platform versions
2020-02-11T00:36:44 DEBUG Fetching platform updates from http://0.0.0.0:8000/test.json
2020-02-11T00:36:44 INFO Checking possible new release version 0.0.1
2020-02-11T00:36:44 INFO Checking possible new release version 0.0.2
(
```

```
postgres=# select * from "houston$default"."PlatformVersion";
            id             | version |                            url                             |   level    |        created
At        |        updatedAt        |     releaseDate     | description
---------------------------+---------+------------------------------------------------------------+------------+---------------
----------+-------------------------+---------------------+-------------
 ck6h1cvuf004e0782yo0xn9p7 | 0.0.1   | https://astronomer.io/cea/release-notes/{new_version}.html | bug_fix    | 2020-02-10 22:
36:21.638 | 2020-02-10 22:36:44.836 | 2018-11-13 20:20:39 |
 ck6h1cvwq004l078224bq3xlg | 0.0.2   | https://astronomer.io/cea/release-notes/{new_version}.html | bug_fix123 | 2020-02-10 22:
36:21.768 | 2020-02-10 22:36:44.884 | 2018-11-13 20:20:39 |
(2 rows)
```

src/scripts/check-platform-updates/index.js
```diff
@@ -0,0 +1,82 @@
...
+      },
+      update: {
+        url: newRelease.url,
+        releaseDate: newRelease.release_date,
+        level: newRelease.level,
+        description: newRelease.description
+      }
+    });
+  }
+}
```
andriisoldatenko: @schnie @ashb wdyt?

andriisoldatenko: https://github.com/astronomer/houston-api/pull/225/files#diff-d7446c3444648f9aa3d0da7b453505faR45-R65

src/scripts/check-platform-updates/index.js
```diff
@@ -0,0 +1,82 @@
...
+import { prisma } from "generated/client";
+import request from "request-promise-native";
+import { size } from "lodash";
+import yargs from "yargs";
+
+export async function getPlatformReleases(uri) {
+  let versions;
+  log.debug(`Fetching platform updates from ${uri}`);
+  try {
+    const response = await request({
```
ashb: Please add as a request parameter the current platform version _ (It's not used for anything currently, but will be useful for seeing what versions we have out there)

src/scripts/check-platform-updates/index.js
ashb: :+1:

src/resolvers/query/platform-versions/index.js
ashb: @andriisoldatenko ^^

andriisoldatenko: 👍

src/scripts/check-platform-updates/index.js
andriisoldatenko: fixed.

andriisoldatenko: fixed.

src/resolvers/query/platform-versions/index.js
```diff
@@ -0,0 +1,10 @@
+/*
+ * Get list of available platform versions.
+ * @param {Object} parent The result of the parent resolver.
+ * @param {Object} args The graphql arguments.
+ * @param {Object} ctx The graphql context.
+ * @return {[]PlatformVersion} List of PlatformVersions.
+ */
+export default async function platformVersions(parent, args, ctx, info) {
+  return ctx.db.query.platformVersions({}, info);
```
ashb: This is going to return all versions, even possibly older ones, or more than one to the UI, which puts the effort of filtering this on the front end (which I don't think it knows what the current platform version is) -- so I think we should

1. Do some filtering here to not include "past" releases (if we are on 0.11.0 we don't care about 0.10.2 from the DB.)
2. Only limit to the single-most interesting release. (I.e. if 0.11.1 and 0.12.0 are out we _maybe_ don't care about 0.11.1 anymore.)

2 isn't quite so clear cut an answer because of update paths and "release streams" where people are more willing to upgrade from 0.11.0 to 0.11.1 than 0.11.0 to 0.12.0.

src/schema.graphql
```diff
@@ -289,6 +297,8 @@ type Query {
 
   self: AuthUser @auth
 
+  platformVersions: [PlatformVersion]
```
ashb: We should possibly make this `@auth` -- otherwise there's a bit of an information disclosure for anon-users on the network, and what version is available (thus what security fixes you haven't applied) seems like sensitive information.

src/resolvers/query/platform-versions/index.js
andriisoldatenko: @ashb what about original requirement?
>We are adding a new service to expose when we release new versions of the platform and airflow. We need to add the ability for Houston to ping this service periodically and determine if there is an update available and log it somewhere where it can be exposed to users in Orbit.
cc @schnie

src/schema.graphql
andriisoldatenko: @schnie wdyt?

src/resolvers/query/platform-versions/index.js
ashb: Take this scenaraio:

- Installed version is 0.11.0
- Update check is performed and 0.11.1 is returned, added to db
- Notice is shown to user
- They upgrade

Point 1 above is about this case -- we have nothing to delete (or otherwise hide) records from the DB, so this notice would still be shown even though they've updated to this version.

Also the updates document will document _every_ version that is available, I wasn't thinking that we'd remove "older" releases from that document.

andriisoldatenko: cc @samblackk do you have thoughts about frontend part?

src/schema.graphql
schnie: Yea, I think that makes sense! Maybe even limit to System level users.

andriisoldatenko: done.

src/resolvers/query/platform-versions/index.js
ashb: For reference here is how I handled this for Airflow frontend https://github.com/astronomer/astronomer-airflow-version-check/pull/1

samblackk: @andriisoldatenko I agree with @ashb that we should filter as much as we can in houston before passing to the frontend. Only the current and "newest" version should come into orbit so the only comparison we need to do is verify if those two values are different, and if so, display the alert. I'm not sure on preference for minor releases. Perhaps that could be configurable? Allow the sys admin to decide if they want to be notified for only major releases or all of them.

ashb: We could filter out to only show newer releases, so if the query only possible upgrades - no version comparison needed on front end then

ashb: We could filter out to only show newer releases, so if the query only possible upgrades - no version comparison needed on front end then

andriisoldatenko:
```
query platformVersions {
     platformVersions {
      version
      url
      level
      releaseDate
    }
}

andrii  10 minutes ago
{
  "data": {
    "platformVersions": [
      {
        "version": "0.0.2",
        "url": "https://astronomer.io/cea/release-notes/{new_version}.html",
        "level": "bug_fix123",
        "releaseDate": "2018-11-13T20:20:39.000Z"
      }
    ]
  }
}

in case of HELM__RELEASE_VERSION=“0.0.1” and we have 2 version in database (0.0.1, 0.0.2)
```

src/schema.graphql
```diff
@@ -289,6 +297,8 @@ type Query {
 
   self: AuthUser @auth
 
+  platformVersions: [PlatformVersion] @auth(permissions: ["system.users.get"])
```
schnie: @andriisoldatenko we should make a new permission here and define it in `default.yaml` - `system.updates.get`.

I think it'd be fine if all `SYSTEM_*` roles had this permission by default, so this could probably be cascaded through those 3 roles.

andriisoldatenko: done

database/datamodel.graphql
```diff
@@ -145,13 +145,13 @@ type DockerImage {
...
 
-type PlatformVersion {
+type PlatformRelease {
   id: ID! @id
   version: String @unique
   url: String
   description: String
   level: String
-  releaseDate: DateTime!
+  releaseDate: DateTime
```
ashb: The airflow plugin has this as non-null (and uses "now" if the update document doesn't have it.)

Perhaps we should change the update document to require this field?

andriisoldatenko: in your example json file it's empty 🤔

src/resolvers/query/platform-releases/index.unit.test.js
```diff
@@ -0,0 +1,38 @@
...
+    // Construct db object for context.
+    const db = {
+      query: {
+        platformReleases
+      }
+    };
+
+    // Run the graphql mutation.
+    const res = await graphql(schema, query, null, { db });
+    expect(res.errors).toBeUndefined();
```
ashb:
```suggestion
    expect(res.errors).toBeUndefined();
    expect(rest.data).toHaveProperty("updateAvailable", null);
```

(That should test that it exists and is explicitly null which is what we want.)

src/resolvers/query/platform-releases/index.unit.test.js
```diff
@@ -0,0 +1,38 @@
...
+const query = `
+  query updateAvailable {
+    updateAvailable {
+      version
+    }
+  }
+`;
+
+describe("updateAvailable", () => {
+  test("typical request is successful", async () => {
```
ashb:
```suggestion
  test("when no updates available", async () => {
```

ashb: And then lets also add a new "test" block that checks when there is an update available (by making the mock platformReleases return something appropriate)

src/resolvers/query/platform-releases/index.js
```diff
@@ -0,0 +1,40 @@
...
+ * @return {PlatformRelease} PlatformRelease.
+ */
+export default async function updateAvailable(parent, args, ctx, info) {
+  // Get current helm release version
+  const currentVersion = config.get("helm.releaseVersion");
+
+  // Get all available versions from database
+  const platformReleases = await ctx.db.query.platformReleases(
+    {},
+    addFragmentToInfo(info, fragment)
```
ashb:
```suggestion
    info
```

No fragment needed/wanted here.

src/resolvers/query/platform-releases/index.unit.test.js
andriisoldatenko: LOL
image

src/resolvers/query/platform-releases/index.js
ashb: Actually, correction: we maybe do, but the only field we need to explicitly ensure is `version`.

src/resolvers/query/platform-releases/index.unit.test.js
andriisoldatenko: fixed.

database/datamodel.graphql
ashb: Yeah -- I'm saying the example is maybe wrong and we should change that to not allow it to be empty there or here. WDYT?

src/resolvers/query/platform-releases/index.unit.test.js
```diff
@@ -0,0 +1,72 @@
...
+    ];
+
+    // Construct db object for context.
+    const db = {
+      query: {
+        platformReleases: jest.fn().mockReturnValue(releases)
+      }
+    };
+
+    config.get = jest
```
ashb: I wonder if this will leak out and affect other tests?

If they pass I guess it's fine. What I did elsewhere was:

```
  describe("in singleNamespace mode", () => {
    beforeAll(() => (config.helm.singleNamespace = true));
    afterAll(() => (config.helm.singleNamespace = false));
    test("Doesn't specify any limits", () => {
      expect(limitRange()).toEqual({});
    });
  });
```

The other option (since we are happy hard-coding this value) is to just put this value directly in to config/test.yaml which is already loaded.

src/resolvers/query/platform-releases/index.unit.test.js
schnie: :grimacing:

src/resolvers/query/platform-releases/index.js
```diff
@@ -0,0 +1,40 @@
...
+  );
+
+  if (size(platformReleases) === 0) {
+    log.info("There are no platform release in database yet :(");
```
schnie:
```suggestion
    log.info("There are no platform releases in the database");
```

src/resolvers/query/platform-releases/index.js
```diff
@@ -0,0 +1,40 @@
+import fragment from "./fragment";
+import log from "logger";
+import { addFragmentToInfo } from "graphql-binding";
+import config from "config";
+import { size, first } from "lodash";
+import semver from "semver";
+/*
+ * Get list of available platform versions.
```
schnie:
```suggestion
 * Check if there is a new platform release available.
```

PR #226, opened 2020-02-11
Add conditional dependency of analytics client

petedejoy:


src/lib/analytics/index.js
```diff
@@ -5,18 +5,18 @@ const Analytics = require("analytics-node");
 const { enabled, writeKey } = config.get("analytics");
 
 // Create the analytics-node client
-const client = new Analytics(writeKey);
+const client = enabled ? new Analytics(writeKey) : null;
```
schnie: @petedejoy should we maybe guard here with `enabled && writeKey ? ...`

petedejoy: Hmm, we could, but then folks would need to track down a write key if they wanted to do local dev with analytics disabled. What do you think?

PR #230, opened 2020-02-13
Fix jest test

samblackk:
Hopefully pass the test that's failing?


schnie:
Hmm, how did the recent change affect this?

samblackk:
I don't know :( 

All I did was change: 
```
  track(ctx.user.id, "Logged In", {
    email: args.identity
  });
```

to: 

```
  track(user.id, "Logged In", {
    email: args.identity
  });
```

because `ctx` was undefined and causing the mutation to stop with an error.

samblackk:
Still working on getting this test to pass. It seems to fail differently locally than in drone.

src/resolvers/mutation/create-token/index.js
```diff
@@ -67,9 +67,11 @@ export default async function createToken(parent, args, ctx) {
   }
 
   // Run the analytics track event
-  track(user.id, "Logged In", {
-    email: args.identity
-  });
+  if (ctx && ctx.user) {
```
schnie: @samblackk what if we dropped the if statement and used `user.id` instead of the `ctx` one? `user` should be guaranteed here since it would have thrown an error earlier in the function otherwise.

schnie:
Fixed in https://github.com/astronomer/houston-api/pull/231 instead.

PR #233, opened 2020-02-15
Add support for Azure DB PostgreSQL - houston

ianstanton:
fix issue https://github.com/astronomer/issues/issues/102
- Add function `cleanCreator` to strip away `@dbserver` from username
- Add function `getAzureDbServer` to get `@dbserver` from username
- Add `@dbserver` to `metadataConnection` and `resultBackendConnection`
  - These values will be passed to the airflow chart and used in `pgbouncer-config` https://github.com/astronomer/airflow-chart/pull/32
- Add unit tests for the above functions
- Required for use with Azure DB for PostgreSQL

### DIFF
--- a/src/lib/deployments/database/index.js
+++ b/src/lib/deployments/database/index.js
@@ -234,8 +234,6 @@ export function cleanCreator(creator) {
 // Get Azure DB Server name from username. Necessary for Azure DB PostgreSQL users.
 export function getAzureDbServer(user) {
   const dbServer = user.match(/@.*/g);
-  if (Array.isArray(dbServer))
-    return dbServer[0];
-  else
-    return "";
+  if (Array.isArray(dbServer)) return dbServer[0];
+  else return "";
 }
--- a/src/lib/deployments/database/index.js
+++ b/src/lib/deployments/database/index.js
@@ -45,6 +45,7 @@ export async function createDatabaseForDeployment(deployment) {
     length: 32,
     numbers: true
   });
+  const azureDbServerName = getAzureDbServer(parsedConn.user);
 
   log.info(`Creating database ${dbName}`);
 
@@ -88,7 +89,8 @@ export async function createDatabaseForDeployment(deployment) {
     pass: airflowPassword,
     host: parsedConn.host,
     port: parsedConn.port,
-    db: dbName
+    db: dbName,
+    azureDbServer: azureDbServerName
   };
 
   // Construt connection details for celery schema.
@@ -97,7 +99,8 @@ export async function createDatabaseForDeployment(deployment) {
     pass: celeryPassword,
     host: parsedConn.host,
     port: parsedConn.port,
-    db: dbName
+    db: dbName,
+    azureDbServer: azureDbServerName
   });
 
   // Return both urls.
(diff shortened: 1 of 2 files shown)


ianstanton:
cc @vishwas-astro

vparekh94:
@ianstanton so what would the user need to do to use this if they were deploying the platform manually (without automation)? Can ya dump that here - I can get it added to the install guide (or just add it yourself if its quick)

src/lib/deployments/database/index.js
```diff
@@ -32,11 +32,14 @@ export async function createDatabaseForDeployment(deployment) {
   const rootConn = createConnection(parsedConn);
 
   // Generate some data.
+  const dbServerName = getDbServer(parsedConn.user);
   const dbName = generateDatabaseName(deployment.releaseName);
   const airflowSchemaName = "airflow";
   const celerySchemaName = "celery";
   const airflowUserName = generateAirflowUsername(deployment.releaseName);
+  const airflowConnUserName = airflowUserName + dbServerName;
```
andriisoldatenko: // add explanation here why we do so.

src/lib/deployments/database/index.unit.test.js
```diff
@@ -0,0 +1,27 @@
...
+    expect(cleaned).toBe("user");
+  });
+
+  test("test if getDbServer function gets azure dbserver name from azure db username", async () => {
+    const creator = "user@azure-db";
+    const azureDbServer = databaseExports.getDbServer(creator);
+    expect(azureDbServer).toBe("@azure-db");
+  });
+
+  test("test if getDbServer function returns empty string for normal user", async () => {
```
andriisoldatenko: add more negative tests...

src/lib/deployments/database/index.js
```diff
@@ -219,3 +225,15 @@ export function parseConnection(conn) {
   if (isString(conn)) return parse(conn);
   return conn;
 }
+
+// Strip '@dbserver' from username. Necessary for Azure DB PostgreSQL users.
+export function cleanCreator(creator) {
+  return creator.replace(/@.*/g, "");
```
sjmiller609: What do you think about adding a log at level "info" right here? Just because replacing the creator might confusing to an operator debugging the connection.

src/lib/deployments/database/index.js
```diff
@@ -219,3 +225,15 @@ export function parseConnection(conn) {
...
+}
+
+// Get DB Server name from username. Necessary for Azure DB PostgreSQL users.
+export function getDbServer(user) {
+  const dbServer = user.match(/@.*/g);
+  if (Array.isArray(dbServer)) return dbServer[0];
```
sjmiller609: How about also logging the 'dbServer' parsed here, before returning? I don't think it's necessary to log if it is not parsed (i.e. in the 'else' block)

ianstanton: Added!

src/lib/deployments/database/index.js
ianstanton: Added!

ianstanton:
Closing until testing is complete

ianstanton:
Closing until we do further performance testing with Azure DB PostgreSQL

PR #235, opened 2020-02-17
InviteToken Source Addition

petedejoy:
We now have the enum `InviteSource` being stored in the `InviteToken` table. This allows us to determine whether an invite is system-level or workspace-level. It also fixes the issue we were experiencing with the Unified Tracking project.


database/datamodel.graphql
```diff
@@ -76,6 +81,7 @@ type InviteToken {
   id: ID! @id
   email: String!
   token: String! @unique
+  source: InviteSource
```
schnie: @petedejoy could make this `!`

src/routes/v1/oauth/callback/post.js
```diff
@@ -65,6 +65,7 @@ export default async function(req, res, next) {
   );
 
   // Set the userId, either the existing, or the newly created one.
+  console.log(state);
```
schnie: @petedejoy

src/lib/users/index.js
```diff
@@ -77,7 +79,10 @@ export async function createUser(opts) {
   if (haveInvite) {
     // Return the invite ID from the database to use in the
     // user record if the user has an invite
-    mutation.id = await prisma.inviteToken({ token: inviteTokens }).id();
+    const inviteToken =
+      find(inviteTokens, { source: INVITE_SOURCE_SYSTEM }) ||
+      head(inviteTokens);
```
schnie:
```suggestion
      head(inviteTokens);

```

src/lib/users/index.js
```diff
@@ -77,7 +79,11 @@ export async function createUser(opts) {
...
+
+    mutation.id = inviteToken.id;
```
schnie: @petedejoy maybe a comment on each of these lines describing whats going on. 

This logic is getting more complicated, so might be worth it to say it in english so the next person knows why.

src/routes/v1/oauth/callback/post.js
petedejoy: NOOOO I WAS DOING SO WELL

src/lib/users/index.js
petedejoy: Good call- fixing that up now

petedejoy: Added some- let me know if you think they're overly verbose.

schnie:
@petedejoy looking great. Only thing left might be to add a test or two to make sure we're handling that logic correctly in the post route.

Maybe a way to check that we correctly select a system token if we do pass it one and another to test that it falls back correctly or something, probably mocking the function that returns the array of tokens.

petedejoy:
@schnie- makes sense! do you mean in the `createUsers` function? We build the array and set the token there, haven't changed the post route at all I don't think.

schnie:
Yes, my bad. I'm thinking some more tests on the `createUsers` function. I was thinking you could mock up `generateRoleBindings` to add some different combinations of rolebindings and test that this logic behaves as you expect.

petedejoy:
@schnie we should be good to go here?

PR #236, opened 2020-02-18
Add more robust tracking for logged in events

petedejoy:
This PR addresses the issue outlined here: https://github.com/astronomer/issues/issues/797. I have tested this and confirmed it works for both local and oauth logins.


src/routes/v1/oauth/callback/post.js
```diff
@@ -80,6 +81,12 @@ export default async function(req, res, next) {
       where: { id: userId },
       data: { fullName, avatarUrl }
     });
+    // Run the analytics track call for a log in event
```
schnie:
```suggestion

    // Run the analytics track call for a log in event
```

PR #237, opened 2020-02-19
Analytics Group

petedejoy:
This PR introduces the analytics.js `group` call into our tracking schema. We can use this to create "companies" in downstream tools that are associated with each workspace on Cloud. These companies have `trialEndsAt` passed as a custom attribute (and updated whenever a sysadmin extends that date) so that our non-techincal sales and CS folks have a nice UI to explore users in workspaces and the important data about that workspace.


src/resolvers/mutation/workspace-add-user/index.js
```diff
@@ -75,6 +79,12 @@ export default async function workspaceAddUser(parent, args, ctx, info) {
       token,
       workspaceLabel: res.workspace.label
     });
+    // Run the group event to bucket user into workspace
+    // Note, we can use the inviteId here because it will be the same as the
+    // userId once the user accepts the invite
+    console.log(res.id);
+    console.log(workspaceUuid);
```
schnie: @petedejoy

src/resolvers/mutation/workspace-add-user/index.js
```diff
@@ -66,7 +70,7 @@ export default async function workspaceAddUser(parent, args, ctx, info) {
           source: INVITE_SOURCE_WORKSPACE
         }
       },
-      `{workspace { label } }`
+      `{ id, workspace { id, label } }`
```
schnie: @petedejoy do we need these changes?

petedejoy: We need the first one (so we get the inviteToken ID), but we can remove the workspace ID return.

src/resolvers/mutation/workspace-add-user/index.js
petedejoy: Shit sorry, somehow missed this review 🤦‍♂

PR #238, opened 2020-03-03
Always update pods when secrets updated

sjmiller609:
Running on dev cloud.

with regard to https://github.com/astronomer/issues/issues/843

Root cause: https://github.com/astronomer/issues/issues/843#issuecomment-593980900

Intended change summary: https://github.com/astronomer/issues/issues/843#issuecomment-594004118

dev test results:
- helm get values does not include airflowPodAnnotations after create
- helm get values includes the annotation after updating
- helm get values includes a different value for the annotation after updating the existing secret's value

If we publish airflow chart with the patch that makes use of 'airflowPodAnnotations' then we can acceptance test this on dev.


src/resolvers/mutation/update-deployment/index.js
```diff
@@ -85,7 +86,16 @@ export default async function updateDeployment(parent, args, ctx, info) {
     });
 
     // Map the user input env vars to a format that the helm chart expects.
-    const envs = mapCustomEnvironmentVariables(updatedDeployment, args.env);
```
schnie: @sjmiller609 can this be `envs` still? Feels more descriptive than `values`.

src/resolvers/mutation/update-deployment/index.js
```diff
@@ -85,7 +86,16 @@ export default async function updateDeployment(parent, args, ctx, info) {
...
+    const values = mapCustomEnvironmentVariables(updatedDeployment, args.env);
+    // Add an annotation to Airflow pods to inform pods to restart when
+    // secrets have been changed
+    let buf = Buffer.from(args.env);
```
schnie: @sjmiller609 these can be `const` too right? We aren't reassigning anywhere below.

src/resolvers/mutation/update-deployment/index.js
```diff
@@ -85,7 +86,16 @@ export default async function updateDeployment(parent, args, ctx, info) {
...
+    let hash = crypto
+      .createHash("sha512")
+      .update(buf)
+      .digest("hex");
+    // This annotation is a sha512 hash of the user-provided Airflow secrets
```
schnie:
```suggestion

    // This annotation is a sha512 hash of the user-provided Airflow secrets
```

src/resolvers/mutation/update-deployment/index.js
```diff
@@ -85,7 +86,16 @@ export default async function updateDeployment(parent, args, ctx, info) {
...
+    values["airflowExtraAnnotations"] = {"checksum/airflow-secrets": hash};
```
schnie:
```suggestion
    values.airflowPodAnnotations = { "checksum/airflow-environment": hash };
```

src/resolvers/mutation/update-deployment/index.js
sjmiller609: envs is not accurate because it now includes other kinds of helm values - airflowExtraAnnotations. This variable is actually a dict of extra, user-provided helm values.

src/resolvers/mutation/update-deployment/index.js
sjmiller609: yes

sjmiller609:
An image with this code is running on dev cloud right now

src/resolvers/mutation/update-deployment/index.js
```diff
@@ -85,7 +86,16 @@ export default async function updateDeployment(parent, args, ctx, info) {
...
     // Map the user input env vars to a format that the helm chart expects.
-    const envs = mapCustomEnvironmentVariables(updatedDeployment, args.env);
+    const values = mapCustomEnvironmentVariables(updatedDeployment, args.env);
+    // Add an annotation to Airflow pods to inform pods to restart when
+    // secrets have been changed
+    const buf = Buffer.from(args.env);
+    const hash = crypto
+      .createHash("sha512")
+      .update(buf)
+      .digest("hex");
```
sjmiller609: testing in dev revealed that the hash is not updated when the value of a secret is updated, so sec

src/resolvers/mutation/update-deployment/index.js
schnie: This was just a new line for spacing and consistency with the rest of the project.

PR #239, opened 2020-03-06
Fix wait-for-deployment

sjmiller609:
with regard to https://github.com/astronomer/issues/issues/917

TLDR boolean type casting an empty array is True in javascript. The int value 0 is False however.

I already committed this change directly to the release branch.


src/resolvers/subscription/deployment-status/index.js
```diff
@@ -51,7 +51,9 @@ export async function subscribe(parent, args, { pubsub }) {
     const res = await Promise.resolve(getMetric(releaseName));
     publish({
       deploymentStatus: {
-        result: res && res.result ? res.result[0].value[1] : 0
+        result: res &&
+                res.result &&
+                res.result.length ? res.result[0].value[1] : 0
```
schnie: Should this be `res.result.length > 0`?

sjmiller609: no, that is redundant.

sjmiller609:
not sure about the merge conflict I just tried to resolve

sjmiller609:
ok looks good now

sjmiller609:
got a lint error, fixing.

sjmiller609:
I switched up the CI to run lint and unit testing without waiting for build first. but I did not put the build downstream of testing, because testing is sufficiently long enough for us to benefit from running them in parallel. Also, it's probably better to know security issues sooner rather than later.

andriisoldatenko:
@ashb @schnie PTAL

package.json
```diff
@@ -9,6 +9,7 @@
     "test": "prisma generate && jest",
     "coverage": "prisma generate && jest --coverage",
     "lint": "prisma generate && eslint src --ignore-pattern src/lib/generated --ext .js --ext .graphql",
+    "fix-lint": "prisma generate && eslint src --ignore-pattern src/lib/generated --ext .js --ext .graphql --fix",
```
andriisoldatenko, line 12: i think you can remove it:
`npm run lint -- --fix` should work.

sjmiller609: ty!

PR #242, opened 2020-03-10
Add email addresses to track call body

petedejoy:
This PR gives us a user's email address on the `properties` object of each track call. This allows us to use the email in streaming slack messages and loads the email into all of our Snowflake tables along with the userId.

### DIFF
--- a/src/lib/analytics/index.js
+++ b/src/lib/analytics/index.js
@@ -1,3 +1,4 @@
+import { prisma } from "generated/client";
 import config from "config";
 const Analytics = require("analytics-node");
 
@@ -17,6 +18,8 @@ export function identify(userId, traits) {
 
 export function track(userId, event, properties) {
   if (client) {
+    const email = usersQuery(userId);
+    properties.email = email;
     client.track({ userId, event, properties });
   }
 }
@@ -25,4 +28,9 @@ export function group(userId, groupId, traits) {
   if (client) {
     client.group({ userId, groupId, traits });
   }
+}
+
+export async function usersQuery(id) {
+  const email = await prisma.user({ id }).email;
+  return email;
 }


schnie:
Hmm, so this will add an extra database call to every `track` event. That does feel a little hefty for analytics.. :thinking:

schnie:
@petedejoy can we get the email from the user object on the context?

petedejoy:
Was thinking about that- we can, but we don’t use ctx to get the user id in all cases. Check out the routes/registry/post track call for an example there- using ctx wouldn’t work.


Pete DeJoy
Product Manager, Astronomer Enterprise
Astronomer | Cloud-Native and Enterprise-Grade Airflow
Mobile: +1.914.426.3051
Calendar: www.calendly.com/astropete

schnie:
@petedejoy do we want to keep pushing this to snowflake or push it into our data warehouse where our DAGs feed. If we did that, we'd
A) Consolidate our Data Warehouse to a single place
B) Could join this analytics data with our normal Houston data, so you'd be able to get emails without triggering another db query for each call in Houston.

petedejoy:
I'm pretty neutral on where we warehouse everything, but we do need emails on some of these calls- for example, we want to see which user is doing what in these slack channels to inform CS behavior. There's probably a more lightweight way of doing this- maybe we can just use ctx where it's already available and then run a query only where absolutely necessary (ie. in the registry post route).

ryw:
Unless someone strongly objects, we will use Snowflake for our data warehouse.

Both @mrrobby and I are working on Astronomer Cloud data pipelines into Snowflake actively:

* he's working on hourly billing data fact tables, so we can see how our billing model builds in real time - right now the billing queries are so deeply nested that it's hard to verify if everything is correct
* he'll also be bringing in work to roll workspaces into customers, involving pipedrive, etc.
* I'm starting today to build a DAG to cycle through all Airflow metadata databases to generate hourly data for # of DAGs, DAG runs, task_instances

All this will be going into Snowflake

ryw:
@petedejoy we shouldn't have to run a houston query w/ every track to get email address. Let's do this another way. Let's do a call to talk it through... track events should only require a userId

schnie:
Sounds good to me. Having it all in one place is the most important factor for me, so we can keep our analytics footprint minimal in Houston.

mrrobby:
Thanks @ryw 

I'll be working on a new schema and loading format as per our upcoming prioritization, so please let me know what data changes are needed so I can keep that the integration layer. The current overview board of milestones can be seen here 

https://app.asana.com/0/1160577638478545/board

You can check the current status and also click into the individual projects from there. I don't have it in the list to add anything to the houston DAGs but if needed immediately, I can look at working that into the external tables setup.

PR #244, opened 2020-03-12
Deployment Subscription: Increase poll interval & kill timeout

samblackk:
Attached to https://github.com/astronomer/issues/issues/940

* 1 minute poll interval
* Kills subscription after 24hrs (matches auth token expire)

This PR should take care of the `/houston-down` we've experience as well as decrease latency on the server caused by too many simultaneous requests from the subscription.


src/resolvers/subscription/deployment-status/index.js
```diff
@@ -54,7 +55,11 @@ export async function subscribe(parent, args, { pubsub }) {
         result: res && res.result ? res.result[0].value[1] : 0
       }
     });
-  }, pubsub);
+  },
+  pubsub,
+  interval,
+  3600000 * 24 // kill after 24 hours
```
schnie: @samblackk should this be a config too?

samblackk: It's not in the config for the other 2 subscriptions. I can come back later and refactor all of them to have this configurable from the config.

PR #246, opened 2020-03-19
Delete system admin rolebinding

ianstanton:
- Adds mutation that allows for removing system admin rolebinding from users
- This is only allowed if there is more than one system admin in existence
- Related to astronomer/issues#780 and astronomer/issues#768
- Example:
```
mutation DeleteAdmin {
  deleteSystemRoleBinding(
    userId: "ck7ypw1my00650987h58i5s87"
    role: SYSTEM_ADMIN
  ) {
    id
  }
}
```


src/resolvers/mutation/delete-system-rolebinding/index.js
```diff
@@ -0,0 +1,50 @@
...
+  if (!startsWith(args.role, "SYSTEM")) throw new InvalidRoleError();
+
+  // Check if the RoleBinding already exists. If not, throw error.
+  const where = { role, user: { id: userId } };
+  const exists = await ctx.db.exists.RoleBinding(where);
+  if (!exists) throw new MissingRoleBindingError();
+
+  // Query for all SYSTEM_ADMIN RoleBindings across all users.
+  const systemAdmins = await ctx.db.query.roleBindings(
+    { where: { role } },
```
schnie: @ianstanton this mutation would be used to also remove a `SYSTEM_EDITOR` or `SYSTEM_VIEWER` too right? I think we only care to do this check if the request is to remove a `SYSTEM_ADMIN`. WDYT?

src/resolvers/mutation/delete-system-rolebinding/index.js
```diff
@@ -0,0 +1,50 @@
...
+  );
+
+  // Throw error if there are less than 2 users with role SYSTEM_ADMIN
+  if (systemAdmins.length < 2) throw new NoSystemAdminError();
+
+  // Query for SYSTEM_ADMIN RoleBinding for the given user.
+  const roleBindings = await ctx.db.query.roleBindings(
+    { where: { role, user: { id: userId } } },
+    "{ role, id }"
+  );
```
schnie: @ianstanton could this and the `exists` query above possibly be combined? Maybe we could move this query up to replace it. Then we could throw the `MissingRoleBindingError` if `size(roleBindings) === 0`, and we could grab the `id` up front, rather than running a second query down here to grab it.

src/resolvers/mutation/delete-system-rolebinding/index.js
```diff
@@ -0,0 +1,50 @@
...
+
+  // Delete RoleBinding.
+  return ctx.db.mutation.deleteRoleBinding(
+    {
+      where: {
+        id: roleBindings[0].id
```
schnie: nitpick but I always like to use lodash's `first()` and `get()` to access stuff like this, like `get(first(roleBindings), "id")`, just for a little more protection against nulls and empties, etc.

ianstanton: Added

src/resolvers/mutation/delete-system-rolebinding/index.js
ianstanton: Good catch! Updated this.

src/resolvers/mutation/delete-system-rolebinding/index.js
ianstanton: Updated

PR #247, opened 2020-03-23
Fix Azure AD user sign up

ianstanton:
- We've found that some Azure AD accounts do not include an `email` attribute.
- When these users attempt to sign up with an invitation, Houston gets caught up on `tokenSet.claims.email.toLowerCase()` and the user receives an `Internal Server Error` message.
- These accounts do include `preferred_username`, which is in the form of an email address.
- `preferred_username` will be used if `email` does not exist.
- Related to astronomer/issues#614


src/routes/v1/oauth/callback/post.js
```diff
@@ -36,12 +36,16 @@ export default async function(req, res, next) {
     nonce: null
   });
 
+  const userEmail = tokenSet.claims.email
+    ? tokenSet.claims.email.toLowerCase()
+    : tokenSet.claims.preferred_username
+    ? tokenSet.claims.preferred_username.toLowerCase()
```
petedejoy: Almost certain this is correct, but I've never used a double ternary before so I'll need to defer to @schnie for validation.

schnie: Should be good

schnie: Huge nitpick but this could be `const email` and down below where we are assigning `userData` it could be shortened to `email`.

src/routes/v1/oauth/callback/post.js
```diff
@@ -36,12 +36,16 @@ export default async function(req, res, next) {
...
+    : null;
+
   // Grab user data
   // Some IDPs don't return useful info, so fall back to the claims if we don't have it
   const userData = merge(await provider.userinfo(tokenSet.access_token), {
-    email:
-      tokenSet.claims.email.toLowerCase() ||
-      tokenSet.claims.preferred_username.toLowerCase(),
+    email: userEmail,
```
schnie: If we do the previous change, this line could be `email,`

ianstanton: Updated

ianstanton:
@schnie Anything else we need here? Is this good to merge?

PR #248, opened 2020-03-23
Handle image metadata response from Azure Blob registry storage

ianstanton:
- Add fix for image metadata pull with registry backed by Azure Blob Storage
- If using Azure Blob registry storage back end, we receive a `307` status code when we fetch image metadata
- We need to follow the link in `headers.location` to get the expected image metadata response.
- Related to astronomer/issues#942
- Related to https://github.com/Azure/acr/issues/217


src/routes/v1/registry/events/post.js
```diff
@@ -146,11 +150,25 @@ export async function extractImageMetadata(ev) {
...
-    await client.get(configURL, {
-      headers: { Accept: manifest.config.mediaType }
-    })
-  ).body;
+  let imageMetadata = {};
+  const metadataResponse = await client.get(configURL, {
+    headers: {
+      Accept: manifest.config.mediaType,
+      Authorization: `Bearer ${dockerJWT}`,
+      "User-Agent": `houston/${version()}`
```
schnie: @ianstanton `User-Agent` could probably be moved to the common headers in the client configuration since it's fine to send to all 3 possible requests. Could also make `Authorization` shared too, but we'd have to override with `null` or something on the follow up redirect request.

ianstanton: Updated

schnie:
@ianstanton looks like we got a conflict with `master` now. Might be because of the last PR we merged earlier.

ianstanton:
@schnie anything else we need here?

schnie:
Nope, just forgot to come back and merge!

PR #252, opened 2020-03-27
Remove baseDomain from service account emails

schnie:
This PR shortens the service account emails for FAB to a standard length below the current 64char limit in FAB.

Closes https://github.com/astronomer/issues/issues/1029


schnie:
@ashb anything we need to worry about when this rolls out? All service account emails will get changed on the next access.

ashb:
I think it should update the email on the row. It might create a second user record in Airflow (but that won't matter).

We should test using an SA before and after

PR #254, opened 2020-03-30
Catch specific error so we can redirect back to UI for better error UX

ryanahamilton:
Part of the solution for astronomer/issues#1053. 

Previously this error stranded users on a Houston URL, with the following error:

image

Now, it catches this specific error and redirects the user back to the Astro UI login screen with an `error=noPubSignups` param. This gives the UI the opportunity to display a friendlier error message (astronomer/astro-ui#458).


src/routes/v1/oauth/callback/post.js
```diff
@@ -64,15 +64,27 @@ export default async function(req, res, next) {
...
+          fullName,
+          email,
+          inviteToken: state.inviteToken,
+          active: true // OAuth users are active immediately
+        });
+  } catch (e) {
+    // Confirm that error is a noPubSignupsError and send user back to UI login
+    // where an error message can be shown
+    const noPubSignupsError = new PublicSignupsDisabledError().message;
+    if (e.message === noPubSignupsError) {
```
schnie: @ryanahamilton instead of using the message string to match, you might be able to check if its an `instanceof PublicSignupsDisbaledError`.

Looks like we're doing something similar here: https://github.com/astronomer/houston-api/blob/e523e22f9da843043ef73b31226eb4779630f3d9/src/routes/v1/registry/events/post.js#L62.

ryanahamilton: Updated. That's much cleaner—thanks!

schnie:
@ashb see any issues with this? Instead of just throwing up a Server Error, we redirect to an astro-ui page to deal with it cleaner.

src/routes/v1/oauth/callback/post.js
```diff
@@ -64,15 +64,26 @@ export default async function(req, res, next) {
...
+      : await _createUser({
+          fullName,
+          email,
+          inviteToken: state.inviteToken,
+          active: true // OAuth users are active immediately
+        });
+  } catch (e) {
+    // Confirm that error is a noPubSignupsError and send user back to UI login
+    // where an error message can be shown
+    if (e instanceof PublicSignupsDisabledError) {
```
ashb: Perhaps the fix here is to catch any error and pass it on to the UI with the "Code" attribute. For example

https://github.com/astronomer/houston-api/blob/9887a9c18f4ce2da90f473b97dc6d2b6db811876/src/lib/errors/errors.js#L18-L28

I think this is available as either `e.code` or `e.extensions.code` https://github.com/astronomer/houston-api/blob/9887a9c18f4ce2da90f473b97dc6d2b6db811876/src/resolvers/mutation/workspace-add-user/index.unit.test.js#L97

ashb:
```suggestion
    if (e instanceof ApolloError) {
```

schnie: True, making this more generic would be nice.

ryanahamilton: I can update this to an `ApolloError` easy enough. There isn't a way to easily pass this error code back to the UI (and redirect) is there? i.e. do we still have to identify the error with the URL param?

schnie: Hmm, I'm not sure of any other way to relay that information, other than through a querystring or specific route, or something.

@ashb this error is out of band from a typical astro-ui request, so we have to deal with it a little different.

ashb: Yeah, the best we could do is pass the `code` up to the UI.

Would also have to make more of our errors subclass ApolloError.

ryanahamilton: Updated. I definitely prefer sending the code instead of an arbitrary string in the url param.

PR #256, opened 2020-03-31
Remove sensitive response from getSecret

ianstanton:
- `res.response` can include sensitive information in houston logs when `getSecret` is called.
- Removes `res.response` if `getSecret` is called.
- Related to astronomer/issues#858


schnie:
Could we at least show the success code if we scrub the output?

ianstanton:
> Could we at least show the success code if we scrub the output?

Good call, added this. An example log will look like:
 `INFO Response from #getSecret: {"result":{"success":true}}`

PR #260, opened 2020-04-14
Add type associations to schema

ryanahamilton:
This will allow us to be more efficient on the client side where multiple queries could be combined and also let us factor out fields like `deploymentCount` in the future.


src/schema.graphql
```diff
@@ -129,6 +129,8 @@ type Workspace {
   properties: JSON
   label: String
   users: [User]
+  deployments: [Deployment]
+  serviceAccounts: [ServiceAccount]
```
schnie, line 133: @ryanahamilton do these work without defining the backing resolvers?

ryanahamilton: Yeah, they appear to Just Work™ in my local testing of the new query I added in astronomer/astro-ui#484.

ryanahamilton: I take that back… it isn't scoping to only "active".

ryanahamilton: Updated with scoped resolver and tests.

PR #265, opened 2020-04-20
Feature: Missing SMTP Notice

samblackk:
Attached to: https://github.com/astronomer/issues/issues/727

Add's support to show SMTP config issues client-side. 
Ready for merge.


schnie:
@samblackk could we make Houston return a boolean instead of the actual SMTP string? This string will contain username and password of the mail server and that would get pushed down to every user of the platform, exposing those creds to anyone snooping around.

src/resolvers/query/app-config/index.js
```diff
@@ -4,10 +4,10 @@ import config from "config";
  * Get information about how app is configured for the cluster.
  */
 export default async function appConfig() {
-  const vars = {
+  return {
     version: config.get("helm.releaseVersion"),
-    baseDomain: config.get("helm.baseDomain")
+    baseDomain: config.get("helm.baseDomain"),
+    smtpEnabled: config.get("email.enabled"),
+    smtpUrl: config.get("email.smtpUrl")
```
schnie, line 11:
```suggestion
    smtpUrlExists: !!config.get("email.smtpUrl")
```
Maybe something like this? Maybe there's a better name.

samblackk: Sure thing

ryanahamilton:
Whats the motivation for two fields (`smtpEnabled`, `smtpExists`) instead of just one (e.g. `smtpConfigured`) given [the client-side usage](https://github.com/astronomer/astro-ui/pull/495) only uses the two fields in tandem?
```js
const noSMTP = !config.smtpEnabled || !config.smtpExists;
```

schnie:
Yea that makes sense.

samblackk:
@ryanahamilton @schnie Suggestion applied. Ready to be seen again.

schnie:
@ryw @andriisoldatenko looks like trivy scan is failing. Did we fix this anywhere else?

PR #268, opened 2020-04-23
Add Deployment singleton query resolver

samblackk:
Attached to https://github.com/astronomer/issues/issues/1067
Required change for https://github.com/astronomer/astro-ui/pull/511

This PR creates a singleton query for `workspaceDeployment` so we don't have to fetch the entire array of the deployments when we just need 1. 

Ready for review.


src/schema.graphql
```diff
@@ -328,7 +328,12 @@ type Query {
 
   workspaceDeployments(
     workspaceUuid: Uuid!
-    releaseName: String
+    releaseName: String # backwards compatibility, move to use workspaceDeployment to query 1 dep
+  ): [Deployment] @auth(permissions: ["workspace.deployments.get", "system.deployments.get"], op: OR)
+
+  workspaceDeployment(
```
ryanahamilton, line 334: Wondering if this (and the underlying resolvers) should just be named `deployment()` as we're essentially querying that table directly?

samblackk: We're using ones without the `workspace` prefix for "sys admin only" queries. ei. `deployments` will return all the deps on the platform to someone with sys admin role bindings. Same for users, service accounts, etc.

ryanahamilton: gotcha, that makes sense. thanks!

src/resolvers/query/workspace-deployment/index.js
```diff
@@ -0,0 +1,23 @@
...
+
+/*
+ * Get 1 deployment for a workspace.
+ * @param {Object} parent The result of the parent resolver.
+ * @param {Object} args The graphql arguments.
+ * @param {Object} ctx The graphql context.
+ * @return {[]Deployment} List of Deployments.
+ */
+export default async function workspaceDeployment(parent, args, ctx, info) {
+  return await ctx.db.query.deployments(
```
schnie, line 12: @ryanahamilton is there a singleton query here? Would we want to return an object instead of an array for singleton queries?

ryanahamilton: Yeah, we should return a single object instead of an array.

src/schema.graphql
```diff
@@ -328,7 +328,12 @@ type Query {
...
+    workspaceUuid: Uuid!
+    releaseName: String!
   ): [Deployment] @auth(permissions: ["workspace.deployments.get", "system.deployments.get"], op: OR)
```
ryanahamilton, line 337: Related to @schnie's comment about returning an object instead of an array:

```suggestion
  ): Deployment @auth(permissions: ["workspace.deployments.get", "system.deployments.get"], op: OR)
```

samblackk:
Good catch. Updated!

src/resolvers/query/workspace-deployment/index.js
```diff
@@ -0,0 +1,20 @@
+import fragment from "./fragment";
+import { addFragmentToInfo } from "graphql-binding";
+
+/*
+ * Get 1 deployment for a workspace.
+ * @param {Object} parent The result of the parent resolver.
+ * @param {Object} args The graphql arguments.
+ * @param {Object} ctx The graphql context.
+ * @return {[]Deployment} List of Deployments.
```
ryanahamilton, line 9:
```suggestion
 * @return {Deployment} The Deployment.
```

src/schema.graphql
```diff
@@ -328,9 +328,14 @@ type Query {
 
   workspaceDeployments(
     workspaceUuid: Uuid!
-    releaseName: String
+    releaseName: String # backwards compatibility, move to use workspaceDeployment to query 1 dep
   ): [Deployment] @auth(permissions: ["workspace.deployments.get", "system.deployments.get"], op: OR)
 
+  workspaceDeployment(
+    workspaceUuid: Uuid!
```
ryanahamilton, line 335: Is `workspaceUuid` needed now that your last commit removed its usage from the args in the underlying resolver?

samblackk: It's required for RBAC checking.

ryanahamilton:
@samblackk I was testing this once more and noticed that the response shape is slightly different than other singleton queries. It's nested in `workspaceDeployment`, requiring the query to look like this:

```js
if (data) {
  const deployment = data.workspaceDeployment;
  if (deployment) return deployment;
}
```
instead of the expected:
```js
if (data) {
  const { deployment } = data;
  if (deployment) return deployment;
}
```

Current response:

```json
{"data":{"workspaceDeployment":{"id":"ck8itcru5007p0746twdkyjaj","label":"Up Top 1","description":"asdwe","type":"airflow","releaseName":"spherical-nova-3509","deployInfo":{"current":null,"__typename":"DeployInfo"},"workspace":{"id":"ck8itch7s004q0746r56l69ff","label":"Numero Uno","__typename":"Workspace"},"urls":[{"type":"airflow","url":"https://deployments.local.astronomer.io/spherical-nova-3509/airflow","__typename":"DeploymentUrl"},{"type":"flower","url":"https://deployments.local.astronomer.io/spherical-nova-3509/flower","__typename":"DeploymentUrl"}],"createdAt":"2020-04-02T13:47:16.587Z","updatedAt":"2020-05-04T21:43:07.216Z","config":{"executor":"CeleryExecutor"},"env":[],"properties":{"component_version":"0.0.0","alert_emails":["ew@ss.com"]},"deploymentCapabilities":{"canUpdateDeployment":true,"canDeleteDeployment":true,"canCreateServiceAccount":true,"canUpdateServiceAccount":true,"canDeleteServiceAccount":true,"__typename":"DeploymentCapabilities"},"__typename":"Deployment"}}}
```

Expected response:

```json
{"data":{"id":"ck8itcru5007p0746twdkyjaj","label":"Up Top 1","description":"asdwe","type":"airflow","releaseName":"spherical-nova-3509","deployInfo":{"current":null,"__typename":"DeployInfo"},"workspace":{"id":"ck8itch7s004q0746r56l69ff","label":"Numero Uno","__typename":"Workspace"},"urls":[{"type":"airflow","url":"https://deployments.local.astronomer.io/spherical-nova-3509/airflow","__typename":"DeploymentUrl"},{"type":"flower","url":"https://deployments.local.astronomer.io/spherical-nova-3509/flower","__typename":"DeploymentUrl"}],"createdAt":"2020-04-02T13:47:16.587Z","updatedAt":"2020-05-04T21:43:07.216Z","config":{"executor":"CeleryExecutor"},"env":[],"properties":{"component_version":"0.0.0","alert_emails":["ew@ss.com"]},"deploymentCapabilities":{"canUpdateDeployment":true,"canDeleteDeployment":true,"canCreateServiceAccount":true,"canUpdateServiceAccount":true,"canDeleteServiceAccount":true,"__typename":"DeploymentCapabilities"},"__typename":"Deployment"}}

```

Is this something we can/should change?

samblackk:
@ryanahamilton I think this is a side effect of how I need to query the deployments (and is correct for our graph). We should change the client side code to: 

```
if (data) {
  const { workspaceDeployment } = data;
  if (workspaceDeployment) return workspaceDeployment;
}
```

ryanahamilton:
@samblackk ah, yep, that's correct. Thanks for the clarification. I've tested this all looks good!

PR #272, opened 2020-05-04
Env variable schema

ryanahamilton:
Needed to support astronomer/astro-ui#523 for the following issues:

- astronomer/issues#1065
- astronomer/issues#1168
- astronomer/issues#222

Remaining:

- [x] Ensure I can update a secret variable to non-secret ONLY if I set a new value

TODO:
- [x] remove new table;
- [x] integrate with commander;
- [x] add unit tests to `deployment-variables/index.js`
- [x] add hide-from-ui annotation.


ashb:
We could store the secret vs not as a JSON (encoded-string) in the annotation of the secret -- that is a fairly common pattern in kube resources, and means we wouldn't need to keep the "secret or not" in our DB.

schnie:
> We could store the secret vs not as a JSON (encoded-string) in the annotation of the secret -- that is a fairly common pattern in kube resources, and means we wouldn't need to keep the "secret or not" in our DB.

A lot, if not all Kubernetes resources have immutable `metadata`. I'm not sure if this is possible or not here.

ashb:
> > We could store the secret vs not as a JSON (encoded-string) in the annotation of the secret -- that is a fairly common pattern in kube resources, and means we wouldn't need to keep the "secret or not" in our DB.
> 
> A lot, if not all Kubernetes resources have immutable `metadata`. I'm not sure if this is possible or not here, or maybe that's just labels.

It appears that annotations for secrets (plus pods and configmaps, not that it's relevant) is editable.

So it is an option if we want to keep this in one place/system.

For example we could have something like this:

```yaml
metadata:
  annotations:
    astronomer.io/hide-from-ui: '["PASSWORD_1", "APIKEY"]'
```

src/resolvers/query/deployment-variables/index.js
```diff
@@ -0,0 +1,52 @@
...
+/*
+ * Get list of deployment environment variables
+ * @param {Object} parent The result of the parent resolver.
+ * @param {Object} args The graphql arguments.
+ * @param {Object} ctx The graphql context.
+ * @return [EnvironmentVariablePayload] The variable payload.
+ */
+export default async function deploymentVariables(parent, args, ctx) {
+  // Get variable keys and metadata from Postgres
+  const deployments = await ctx.db.query.deployments(
```
schnie, line 16: Is there a singleton `deployment` query so we don't have to access `[0]` down below?

src/resolvers/mutation/update-deployment-variables/index.js
```diff
@@ -0,0 +1,97 @@
...
+
+  // Send variable values to commander
+  await ctx.commander.request("setSecret", {
+    release_name: releaseName,
+    namespace: generateNamespace(releaseName),
+    secret: {
+      name: generateEnvironmentSecretName(releaseName),
+      data: arrayOfKeyValueToObject(toKeyValue)
+    }
+  });
```
schnie, line 74: After we call `setSecret` here, we'll need to call `updateDeployment` and pass in the updated list of environment variables to the airflow chart under the [secret section](https://github.com/astronomer/airflow-chart/blob/master/values.yaml#L104-L108). This is what tells the Airflow pods to mount the specific keys inside the freshly updated secret. This same process is done in the [updateDeployment resolver](https://github.com/astronomer/houston-api/blob/master/src/resolvers/mutation/update-deployment/index.js#L78-L111).

I was going to suggest that we pull this code somewhere under `lib/deployments` to share between the two resolvers, but I'm wondering if we should just straight up move it from there to this new resolver, since the old one won't be used to update env vars anymore.

If anyone is using `updateDeployment` to set env vars by consuming the API themselves, this would break their stuff though. I wonder if Apollo metrics would have any insights here. :thinking: Did we fully set that up?

src/resolvers/mutation/update-deployment-variables/index.js
```diff
@@ -0,0 +1,97 @@
...
+ * @param {Object} parent The result of the parent resolver.
+ * @param {Object} args The graphql arguments.
+ * @param {Object} ctx The graphql context.
+ * @return [EnvironmentVariablePayload] The variable payload.
+ */
+export default async function updateDeploymentVariables(parent, args, ctx) {
+  let { deploymentUuid, releaseName, payload } = args;
+
+  // Get Deployment Variables
+  let currentVariables = await ctx.db.query.environmentVariables({
```
schnie, line 19: Nitpicking but could we make these `const` for consistency?

schnie:
Interesting @ashb. Can you also add new keys to an existing secret, as opposed to updating the value of an existing key?

@andriisoldatenko we'd probably need to update something in Commander to support this (depending on a few things), but I do like that it keeps this data in one place.

database/datamodel.graphql
```diff
@@ -117,6 +117,15 @@ type Workspace {
   trialEndsAt: DateTime
 }
 
+type EnvironmentVariable {
```
andriisoldatenko, line 120: @schnie wdyt if we move it to type in graphql instead of new table?

schnie: Yea, if we go with consolidating storage into the secret, then we can back this resolver with a commander call to get the secrets.

schnie:
Discussed further in slack - https://astronomerteam.slack.com/archives/C02NZELT2/p1588788110112500.

The cleanest approach, assuming this works in practice, is to store a list of env var keys that the user has marked as "secret" in an annotation on the kube secret.

We'll need to adjust Houston and Commander to account for this in Commanders `getSecret` and `setSecret` methods. `getSecret` should pass this annotation data back to Houston and `setSecret` should allow Houston to pass this info to Commander.

This can keep the general "flow" of our query and mutation similar to the rest of them for now. We can circle back and address the "distributed transaction" problem via a new issue.

schnie:
Kubernetes max secret size is 1MB, so maybe we `JSON.stringify()` the object we'd send to commander, base64 encode it, and validate that the final size of that is under 1MB.

schnie:
In addition, as discussed in slack we'll be limited to 262,144 chars for the _list_ of secret env vars. Probably out of range, but worth noting here.

schnie:
As also decided in Slack, Commander's protos (the interface) for these procedures will allow Houston to specify arbitrary key/values for a new `annotations` field. This will keep the logic inside of Houston about what annotations to send Commander.

For now, all we care about is sending/updating the one annotation:
`astronomer.io/hide-from-ui: '["SECRET_ENV_VAR_1", "SECRET_ENV_VAR_2"]'`

sjmiller609:
Regarding my initial comment, I think that if the request fails and bubbles up to the UI when the kube secret *-env fails to update, and this is considered the source of truth (it is fetched before next update, responding with failure if fail to fetch), then it is fine.

sjmiller609:
@ashb not trying to throw a wrench in things, but what is the difference in storing the env vars in annotations of the secret vs in the secret's data? I do think using the secret data as source of truth is fine so long as that is considered the source of truth.

ashb:
We can't store the "visibility" flag in the secret data as easily - we'd have to store a separate key I think.

```
data:
  APIKEY: X
  APIKEY__is_secret: true
```

Vs annotations:

```
metadata:
  annotations:
    astronomer.io/hidden-from-ui: '["APIKEY"]'
data:
  APIKEY: X
```

sjmiller609:
ah gotcha, thank you. I was thinking it was somehow related to my comment and I was worried I sent you down a rabbit hole.

andriisoldatenko:
@ryanahamilton 
is it possible to add tests to be on same level of code coverage?

src/schema.graphql
```diff
@@ -215,13 +216,31 @@ type DeployInfo {
...
+}
+
 # Deployment properties are only to support the legacy API.
 # Properties have been moved to top-level fields on Deployment internally.
 type Deployment {
   id: ID!
   uuid: Uuid @alias(field: "id")
   config: JSON
-  env: JSON
+  env: JSON @deprecated # backwards capability
```
andriisoldatenko, line 242: @schnie @ashb 
do we need to change schema? with new secret annotations?
i mean do we want to store it in separate virtual type or store in Json in Deployment record?

schnie: Are you suggesting that we keep the schema the same? I think we wanted to have a more strongly typed system in place, rather than the raw `JSON`. The backend for this new resolver is the kube secret, rather than the db since we wanted to consolidate all state to one place.

src/resolvers/query/deployment-variables/index.js
andriisoldatenko: fixed.

andriisoldatenko:
practically LGTM:

@schnie @samblackk @ryanahamilton PTAL!
cc @ashb @adam2k 

image

```
apiVersion: v1
data:
  "0": dHJ1ZQ==
  "1": dHJ1ZQ==
  AIRFLOW__CLI__ENDPOINT_URL: dHJ1ZQ==
  AIRFLOW__CORE__AIRFLOW_HOME: dHJ1ZQ==
  AIRFLOW__CORE__BASE_LOG_FOLDER: dHJ1ZQ==
  AIRFLOW__CORE__COLORED_LOG_FORMAT: dHJ1ZQ==
  AIRFLOW__CORE__DAGS_FOLDER: dHJ1ZQ==
  AIRFLOW__CORE__REMOTE_BASE_LOG_FOLDER: dHJ1ZQ==
  AIRFLOW__CORE__REMOTE_LOG_CONN_ID: dHJ1ZQ==
  AIRFLOW__KUBERNETES__DAGS_VOLUME_CLAIM: dHJ1ZQ==
  AIRFLOW__KUBERNETES__GIT_SUBPATH: ZmFsc2U=
  AIRFLOW__SCHEDULER__CHILD_PROCESS_LOG_DIRECTORY: MTIzNDU2Nzg5MA==
  AIRFLOW__SCHEDULER__DAG_DIR_LIST_INTERVAL: MTI=
  AIRFLOW__WEBSERVER__WEB_SERVER_SSL_CERT: dHJ1ZQ==
kind: Secret
metadata:
  annotations:
    astronomer.io/hide-from-ui: '["AIRFLOW__KUBERNETES__GIT_SUBPATH","AIRFLOW__SCHEDULER__CHILD_PROCESS_LOG_DIRECTORY"]'
  creationTimestamp: "2020-06-08T12:18:06Z"
  labels:
    component: airflow-secret
    release: ""
    tier: airflow
  name: telescopic-frequency-6298-env
  namespace: astronomer-telescopic-frequency-6298
  resourceVersion: "188667"
  selfLink: /api/v1/namespaces/astronomer-telescopic-frequency-6298/secrets/telescopic-frequency-6298-env
  uid: 3b9f08cc-4294-418c-868c-5bbb972c088a
type: Opaque
(base) ➜  astro-cli git:(master)

```

src/resolvers/mutation/update-deployment-variables/index.js
andriisoldatenko: @schnie it's a good point wdyt if i add update-env-variables to:
https://github.com/astronomer/houston-api/blob/2b45d7f16249cf3751b4eccfe5a31f3a04017a77/src/resolvers/mutation/update-deployment/index.js#L101 ?

src/resolvers/mutation/update-deployment-variables/index.js
andriisoldatenko: fixed.

database/datamodel.graphql
andriisoldatenko: done.

src/resolvers/mutation/update-deployment-variables/index.js
```diff
@@ -0,0 +1,133 @@
...
+  // Send variable values to commander
+  await ctx.commander.request("setSecret", {
+    release_name: releaseName,
+    namespace: generateNamespace(releaseName),
+    secret: {
+      name: generateEnvironmentSecretName(releaseName),
+      data: arrayOfKeyValueToObject(updatedVariables),
+      annotations: { "astronomer.io/hide-from-ui": JSON.stringify(annotations) }
+    }
+  });
```
adam2k, line 65:
```Suggestion
const data = arrayOfKeyValueToObject(updatedVariables);
const secretAnnotations = { "astronomer.io/hide-from-ui": JSON.stringify(annotations) };
const secret = {
    release_name: releaseName,
    namespace, // Only do this if you can you use the namespace generated above.
    secret: {
      name, // Only do this if you can you use the name generated above.
      data,
      annotations: secretAnnotations // because annotations is already defined above
    }
  }
await ctx.commander.request("setSecret", secret);
```

src/resolvers/query/deployment-variables/index.js
```diff
@@ -0,0 +1,61 @@
...
+ * Get list of deployment environment variables
+ * @param {Object} parent The result of the parent resolver.
+ * @param {Object} args The graphql arguments.
+ * @param {Object} ctx The graphql context.
+ * @return [EnvironmentVariablePayload] The variable payload.
+ */
+export default async function deploymentVariables(parent, args, ctx) {
+  // Get variable keys and metadata from Postgres
+  const { id } = await ctx.db.query.deployment(
+    { where: { releaseName: args.releaseName } },
```
adam2k, line 19: Can you deconstruct `args.releaseName` at the beginning of the function to `const { releaseName } = args;` and then swap all instances of `args.releaseName` to `releaseName`?

and in this case `{ releaseName: args.releaseName }`  could be `{ releaseName }`.

src/resolvers/query/deployment-variables/index.js
```diff
@@ -0,0 +1,61 @@
...
+    isSecret: includes(secretKeys, variable.key)
+  }));
+
+  // Build EnvironmentVariablePayload
+  const result = {};
+  result.releaseName = args.releaseName;
+  result.deploymentUuid = id;
+
+  // Order variables alphabetically
+  result.environmentVariables = orderBy(vars, ["key"], ["asc"]);
```
adam2k, line 57:
```Suggestion
  const environmentVariables = orderBy(vars, ["key"], ["asc"]);
  
  return {
    deploymentUuid: id,
    environmentVariables,
    releaseName // This will work if you use the suggestion about destructing the args at the top.
  };
```

src/resolvers/mutation/update-deployment-variables/index.js
```diff
@@ -0,0 +1,133 @@
...
+ * @param {Object} ctx The graphql context.
+ * @return [EnvironmentVariablePayload] The variable payload.
+ */
+export default async function updateDeploymentVariables(parent, args, ctx) {
+  let { deploymentUuid, releaseName, payload } = args;
+
+  // Get Deployment Variables
+  const response = await ctx.commander.request("getSecret", {
+    namespace: generateNamespace(args.releaseName),
+    name: generateEnvironmentSecretName(args.releaseName)
```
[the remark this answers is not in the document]
andriisoldatenko: @adam2k can you please explain why suggested change better than existing one? 
Also for consistency if we are going to change this part we must change in whole code base?

src/resolvers/mutation/update-deployment-variables/index.js
```diff
@@ -0,0 +1,133 @@
...
+  });
+
+  // Get current env variables from commander response
+  const currentVariables = objectToArrayOfKeyValue(
+    get(response, "secret.data", {})
+  );
+
+  // Build payload array for commander
+  const newVariables = payload.map(variable => ({
```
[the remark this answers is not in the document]
andriisoldatenko: @adam2k what's the benefit to moving to another method?

src/resolvers/mutation/update-deployment-variables/index.js
```diff
@@ -0,0 +1,133 @@
...
+  // Create array of secret keys as k8s annotations
+  const annotations = map(
+    filter(updatedVariables, { isSecret: true }),
+    i => i.key
+  );
+
+  // Send variable values to commander
+  await ctx.commander.request("setSecret", {
+    release_name: releaseName,
+    namespace: generateNamespace(releaseName),
```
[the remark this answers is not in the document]
andriisoldatenko: good catch we can doo it only once.

src/resolvers/mutation/update-deployment-variables/index.unit.test.js
```diff
@@ -0,0 +1,107 @@
...
+      {
+        key: "KEY_2",
+        value: "VAL_2",
+        isSecret: false
+      }
+    ];
+    expect(await mergeEnvVariables(currentVars, newVars)).toEqual(mergedVars);
+  });
+
+  test("merge old env variable KEY_1 (secret) with new KEY_2 (non secret) with same value", async () => {
```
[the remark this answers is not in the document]
andriisoldatenko: fixed.

src/resolvers/mutation/update-deployment-variables/index.unit.test.js
```diff
@@ -0,0 +1,107 @@
...
+      {
+        key: "KEY_2",
+        value: "VAL_2",
+        isSecret: true
+      }
+    ];
+    expect(await mergeEnvVariables(currentVars, newVars)).toEqual(mergedVars);
+  });
+
+  test("merge old env variable KEY_1 (secret) with new KEY_2 (non secret)", async () => {
```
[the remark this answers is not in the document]
andriisoldatenko: fixed.

src/resolvers/mutation/update-deployment-variables/index.unit.test.js
```diff
@@ -0,0 +1,107 @@
...
+      {
+        key: "KEY_2",
+        value: "VAL_2",
+        isSecret: false
+      }
+    ];
+    expect(await mergeEnvVariables(currentVars, newVars)).toEqual(mergedVars);
+  });
+
+  test("merge old env variable KEY_1 (non secret) with new KEY_2 (secret)", async () => {
```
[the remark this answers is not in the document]
andriisoldatenko: fixed.

src/resolvers/mutation/update-deployment-variables/index.unit.test.js
```diff
@@ -0,0 +1,107 @@
+import { mergeEnvVariables } from "./index";
+
+describe("mergeEnvVariables", () => {
+  test("merge old env variable KEY_1 (non secret) with new KEY_2 (non secret)", async () => {
```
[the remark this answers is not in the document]
andriisoldatenko: fixed.

src/resolvers/mutation/update-deployment-variables/index.js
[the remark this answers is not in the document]
adam2k: I was talking to @kushalmalani about this too.  We'll probably need to agree as a group what is best. I was thinking we could change the files as we go (either adding new features or refactoring code).  IMHO it's easier to read when the objects are implicitly defined in JavaScript and when there aren't nested method calls.

src/resolvers/mutation/update-deployment-variables/index.js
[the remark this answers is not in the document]
adam2k: It could make it so that you don't need to comment on line 37.  If the method was named something like the comment then it's less to read and easier to understand, so instead of: `// Build payload array for commander` and then the lines of code you could simply have something like `const newVariables = buildPayloadForCommander(payload)`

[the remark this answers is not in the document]
adam2k: This also makes unit testing easier for us.  I know that this particular block can be tested in context, but just to get into that habit is mostly helpful.

andriisoldatenko:
@ashb @schnie @adam2k PTAL!

src/resolvers/mutation/update-deployment-variables/index.js
```diff
@@ -0,0 +1,133 @@
...
+  };
+
+  // Get the deployment first.
+  const deployment = await ctx.db.query.deployment(
+    { where: { id: args.deploymentUuid } },
+    queryFragment
+  );
+
+  // Update the deployment, passing in our custom env vars.
+  await ctx.commander.request("updateDeployment", {
```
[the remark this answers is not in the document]
andriisoldatenko: for consistency i'll leave as is. In future we can refactor in all places.

src/resolvers/mutation/update-deployment-variables/index.js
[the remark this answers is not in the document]
andriisoldatenko: i think it's so simple code to extract to function. will wait for @schnie opinion on it.

src/resolvers/query/deployment-variables/index.js
```diff
@@ -0,0 +1,63 @@
...
+  const { releaseName } = args;
+
+  // Get variable keys and metadata from Postgres
+  const { id } = await ctx.db.query.deployment(
+    { where: { releaseName } },
+    "{ id }"
+  );
+
+  const namespace = generateNamespace(releaseName);
+  const secretName = generateEnvironmentSecretName(args.releaseName);
```
schnie, line 26:
```suggestion
  const secretName = generateEnvironmentSecretName(releaseName);
```

src/schema.graphql
schnie: @andriisoldatenko can/should we remove the `env` field as well as the old backend resolver here?

src/schema.graphql
```diff
@@ -215,13 +216,29 @@ type DeployInfo {
...
+type EnvironmentVariable {
+  releaseName: String!
+  createdAt: DateTime!
+  updatedAt: DateTime!
+  key: String,
+  value: String,
+  isSecret: Boolean,
+}
+
+type DeploymentVariablesPayload {
```
schnie, line 228: As we type more of our API, maybe we should think about some standard naming conventions. For this one, I'm wondering if `EnvironmentVariablesResponse` would work. `EnvironmentVariables` keeps it related to the wrapped type - `EnvironmentVariable`, and `Response` makes it clear that this is a type that bundles up data for an operation response.

If we have a similar situation soon, we can reuse this pattern. This would also line up with the names of our Commander request/response protos.

Stumbled upon this doc too - https://www.apollographql.com/docs/apollo-server/schema/schema/#structuring-mutation-responses. Good doc in general - cc @andriisoldatenko @adam2k

andriisoldatenko: @ryanahamilton cc please update UI accordingly .

andriisoldatenko:
practical test in cluster:
**astro-ui:**
image

**k8s secret env:**
```
 k get secrets nuclear-hemisphere-5721-env -n astronomer-nuclear-hemisphere-5721 -o yaml
apiVersion: v1
data:
  TEST: dGVzdA==
  TEST_SECRET: dGVzdA==
kind: Secret
metadata:
  annotations:
    astronomer.io/hide-from-ui: '["TEST_SECRET"]'
  creationTimestamp: "2020-06-09T15:59:15Z"
  labels:
    component: airflow-secret
    release: ""
    tier: airflow
  name: nuclear-hemisphere-5721-env
  namespace: astronomer-nuclear-hemisphere-5721
  resourceVersion: "15813"
  selfLink: /api/v1/namespaces/astronomer-nuclear-hemisphere-5721/secrets/nuclear-hemisphere-5721-env
  uid: 0016ae67-986a-46bd-a20a-5730d30fbf6a
type: Opaque
```

andriisoldatenko:
last commit fix the issue:
```
![image](https://user-images.githubusercontent.com/1374633/84203269-db18f280-aab1-11ea-8b21-1126f8688696.png)
/usr/local/airflow $ set | grep -i TEST
AIRFLOW__CORE__AIRFLOW_HOME='/usr/local/airflow/test'
TEST='test'
TEST_SECRET='test'
```
@schnie PTAL

andriisoldatenko:
practically works well in KinD together with [commander](https://github.com/astronomer/commander/pull/30)/and [astro-ui](https://github.com/astronomer/astro-ui/pull/523)

src/resolvers/mutation/update-deployment-variables/index.unit.test.js
```diff
@@ -0,0 +1,260 @@
...
+describe("updateDeploymentVariables", () => {
+  test("typical request is successful", async () => {
+    // Create some deployment vars.
+    const id = casual.uuid;
+    const releaseName = generateReleaseName();
+
+    const deployment = jest.fn().mockReturnValue({
+      releaseName,
+      workspace: {
+        id: casual.uuid,
```
adam2k, line 38: Could this be `id,` and use the value of `const id` abover or does it break the test?

src/resolvers/mutation/update-deployment-variables/index.unit.test.js
```diff
@@ -0,0 +1,260 @@
...
+      value
+      isSecret
+    }
+  }
+}`;
+
+describe("updateDeploymentVariables", () => {
+  test("typical request is successful", async () => {
+    // Create some deployment vars.
+    const id = casual.uuid;
```
adam2k, line 32: Can all of these `const`s be converted to `let`s, pulled up into the describe block scope and reinitialized in a `beforeEach(() ...` jest method? It seems like at least `id`, `releaseName`, `commander` and `user` are all the same in each test.  If you can move all of the vars it might help to move some of into the describe block scope.

src/resolvers/mutation/update-deployment-variables/index.unit.test.js
```diff
@@ -0,0 +1,260 @@
...
+    const releaseName = generateReleaseName();
+
+    const deployment = jest.fn().mockReturnValue({
+      releaseName,
+      workspace: {
+        id: casual.uuid,
+        stripeCustomerId: casual.uuid,
```
adam2k, line 39: Same as the comment above.  Can this use the `const id` or

src/resolvers/mutation/update-deployment-variables/index.js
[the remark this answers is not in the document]
adam2k: no problem.  This is minor and only a suggestion.

andriisoldatenko:
@schnie @ashb PTAL!

PR #274, opened 2020-05-06
allow annotations to be applied to service account

andriisoldatenko:
fix https://github.com/astronomer/issues/issues/851


src/lib/deployments/config/index.js
```diff
@@ -61,7 +61,8 @@ export function generateHelmValues(deployment, values = {}) {
     elasticsearch(deployment), // Apply the elasticsearch connection details.
     defaultAirflowTag(deployment), // Apply the default airflow tag we want to deploy.
     platform(deployment), // Apply astronomer platform specific values.
-    deploymentOverrides(deployment) // The deployment level config.
+    deploymentOverrides(deployment), // The deployment level config.
+    serviceAccountAnnotations(deployment) // The deployment level config for service account annotation.
```
schnie, line 65: Can we move this new line up right above the `platform` section? `deploymentOverrides` needs to come last so that we actually _override_ anything.

andriisoldatenko: fixed.

andriisoldatenko:
```
limits:
    -
        type: Pod
        max:
            cpu: 3500m
            memory: 13440Mi
    -
        type: Container
        default:
            cpu: 100m
            memory: 384Mi
        defaultRequest:
            cpu: 100m
            memory: 384Mi
quotas:
    pods: 14
    requests.cpu: 6200m
    requests.memory: 23808Mi
    limits.cpu: 6200m
    limits.memory: 23808Mi
defaultAirflowTag: 1.10.4-alpine3.10-onbuild
serviceAccountAnnotations:
    eks.amazonaws.com/role-arn: 'arn:aws:iam::123456789012:role/my-role'
labels:
    platform: deadly-terrier
    workspace: ck9x7thjp00340753ofc428fp
platform:
    release: deadly-terrier
    workspace: ck9x7thjp00340753ofc428fp
executor: CeleryExecutor
cloudRole: 'arn:aws:iam::123456789012:role/my-role'
```
practically looks good, except i need to somehow to remove last line...

src/lib/deployments/config/index.js
```diff
@@ -593,6 +594,25 @@ export function deploymentOverrides(deployment) {
...
 }
 
+/*
+ * Generate the service account annotations.
+ * @param {Object} deployment A deployment object.
+ * @return {Object} Properly formatted config, pulled from deployment.
+ */
+export function serviceAccountAnnotations(deployment) {
+  // Get the deployment config.
+  const cloudRole = get(deployment, "config.cloudRole", {});
```
ashb, line 604: Where is this coming from?

src/lib/deployments/config/index.js
```diff
@@ -593,6 +594,25 @@ export function deploymentOverrides(deployment) {
...
+
+  // Get service account annotation key
+  const serviceAccountAnnotationKey = config.get(
+    "deployments.serviceAccountAnnotationKey"
```
ashb, line 608: Why do we need this config? Can't we just pass a dictionary of annotations to apply?

andriisoldatenko: https://github.com/astronomer/houston-api/blob/9f367c7b0471fa114b573a9477537610d01594d7/config/default.yaml#L537

src/lib/deployments/config/index.js
andriisoldatenko:
```
{"workspaceUuid":"ck9x7thjp00340753ofc428fp","releaseName":"","type":"airflow","config":{"executor":"CeleryExecutor", "cloudRole":"arn:aws:iam::123456789012:role/my-role"},"label":"c1","description":"","properties":{}}
```

query:
```
mutation createDeployment($workspaceUuid: Uuid!, $releaseName: String!, $type: String!, $label: String!, $description: String, $version: String, $config: JSON, $env: JSON, $properties: JSON) {
  createDeployment(workspaceUuid: $workspaceUuid, releaseName: $releaseName, type: $type, label: $label, description: $description, version: $version, config: $config, env: $env, properties: $properties) {
    id
    config
    env
    urls {
      type
      url
      __typename
    }
    properties
    description
    label
    releaseName
    status
    type
    version
    workspace {
      id
      label
      __typename
    }
    createdAt
    updatedAt
    __typename
  }
}
```

src/lib/deployments/config/index.js
ashb: I mean I don't get what this config is _doing_.

I would have expected the config to instead by something like:

```
 serviceAccountAnnotations:
    eks.amazonaws.com/role-arn: arn:aws:iam::123456789012:role/my-role
```

That, or we just store the annotations we want directly in the config property like this:

```
{
  "workspaceUuid": "ck9x7thjp00340753ofc428fp",
  "releaseName": "",
  "type": "airflow",
  "config": {
    "executor": "CeleryExecutor",
    "serviceAccountAnnotations": {
      "eks.amazonaws.com/role-arn": "arn:aws:iam::123456789012:role/my-role"
    }
  },
  "label": "c1",
  "description": "",
  "properties": {}
}
```

I don't get the purpose of the indrection/extra layer.

ashb: (If this has been discussed before feel free to point me at where)

config/default.yaml
```diff
@@ -532,3 +532,6 @@ deployments:
     # the KubernetesPodOperator and KubernetesExecutor. This is what signals to apply the
     # required RoleBindings.
     allowPodLaunching: true
+
+  # Deployment Service account annotation key, depends on cloud provider. Default value for AWS
+  serviceAccountAnnotationKey: eks.amazonaws.com/role-arn
```
ashb, line 537:
```suggestion
  # Annotation name to map the `cloud-role` from creating a deployment to on the Deployment Service account This is the correct value for AWS
  cloudRoleServiceAccountAnnotationName: eks.amazonaws.com/role-arn
```

Slight change of name to make this a bit clearer to understand just based on the config file.

src/lib/deployments/config/index.js
```diff
@@ -593,6 +594,25 @@ export function deploymentOverrides(deployment) {
...
+  );
+
+  return {
+    serviceAccountAnnotations: { [serviceAccountAnnotationKey]: cloudRole }
```
ashb, line 612: We should `return {}` if cloudRole is not set.

src/lib/deployments/config/index.js
andriisoldatenko: the idea is to not hardcode `eks.amazonaws.com/role-arn` in astro-cli and do smth like it:
```
astro deployment create --cloud-role=arn:aws:iam::123456789012:role/my-role
```

also in another cloud provider client can configure `cloudRoleServiceAccountAnnotationName` and all will work.

src/lib/deployments/config/index.js
andriisoldatenko: fixed.

src/lib/deployments/config/index.js
ashb: Ah gotcha.

src/lib/deployments/config/index.js
```diff
@@ -593,6 +594,27 @@ export function deploymentOverrides(deployment) {
...
 }
 
+/*
+ * Generate the service account annotations.
+ * @param {Object} deployment A deployment object.
+ * @return {Object} Properly formatted config, pulled from deployment.
+ */
+export function serviceAccountAnnotations(deployment) {
+  // Get the deployment config.
+  const cloudRole = get(deployment, "config.cloudRole");
```
schnie, line 604: Hmm, I think this is why it's showing up in the output of the helm config. These `config` properties are mapped straight into the helm input.

Maybe we should flip this around a little bit. 

When the user calls the mutation to create or update a deployment and specifies the `cloudRole` flag, maybe we should do this work and map the `{ [cloudRoleServiceAccountAnnotationName]: cloudRole }` part into the users `config` object in the database.

It's essentially moving this logic to the resolver and then we won't need the extra function here and can rely on the `deploymentOverrides` to pass this info through.

andriisoldatenko:
@schnie PTAL.

schnie:
This change introduces some different behavior on `args.config`.

Currently, this blob of JSON just gets passed straight through to the mutation. This new feature is changing that to also contain a special flag that does not directly get passed through into the mutation or helm - `args.config.cloudRole`.

I feel like we may want to make this flag a top-level on the API requests, rather than adding more stuff to the already untyped `config` JSON blob.

@ashb @andriisoldatenko what do you think?

andriisoldatenko:
@schnie yeah sounds good to me, TBH i don't like this magic `config` of JSON type.
+1 from y side.

src/resolvers/mutation/create-deployment/index.js
```diff
@@ -108,12 +108,29 @@ export default async function createDeployment(parent, args, ctx, info) {
     ...args.properties
   };
 
+  let serviceAccountAnnotations;
```
schnie, line 111:
```suggestion
  const serviceAccountAnnotations = {};
```

This way you can keep it a constant, but by making it an object, we can still append keys down below.

src/resolvers/mutation/create-deployment/index.js
```diff
@@ -141,7 +158,7 @@ export default async function createDeployment(parent, args, ctx, info) {
     description: args.description,
     releaseName,
     createdAt: deployment.createdAt,
-    config: args.config
+    config: deploymentConfig
```
schnie, line 161: Hmm, this object will now contain information about the users cloud environment.

@petedejoy can we drop the tracking of `config` to segment?

schnie: But I guess only enterprise users would be using this particular feature, and they wouldn't have tracking enabled.

Still feels a little weird. If there's no issue, I'd say we drop `config` from this call.

andriisoldatenko: i can do just `args.config` and ignore role?

src/resolvers/mutation/create-deployment/index.js
andriisoldatenko: fixed.

andriisoldatenko:
@schnie PTAL

src/resolvers/mutation/create-deployment/index.js
```diff
@@ -108,12 +108,28 @@ export default async function createDeployment(parent, args, ctx, info) {
...
+    const serviceAccountAnnotationKey = config.get(
+      "deployments.cloudRoleServiceAccountAnnotationName"
+    );
+    serviceAccountAnnotations[serviceAccountAnnotationKey] = args.cloudRole;
+  }
+
+  const deploymentConfig = merge(
+    {},
+    args.config || generateDefaultDeploymentConfig(),
+    { serviceAccountAnnotations }
```
schnie, line 124: @andriisoldatenko should this be enclosed in {} or just be bare? This already is an object.

src/resolvers/mutation/update-deployment/index.js
```diff
@@ -49,10 +49,24 @@ export default async function updateDeployment(parent, args, ctx, info) {
...
+  // Generate the service account annotations.
+  if (args.cloudRole) {
+    const serviceAccountAnnotationKey = config.get(
+      "deployments.cloudRoleServiceAccountAnnotationName"
+    );
+    serviceAccountAnnotations[serviceAccountAnnotationKey] = args.cloudRole;
+  }
+
+  const deploymentConfig = merge({}, args.config, {
+    serviceAccountAnnotations
```
schnie, line 63: Same as above for here too.

andriisoldatenko: @schnie 
i tested practically and without `{}` i see in config something like it:
`{"config":  {"eks.amazonaws.com/role-arn":"test"}}`

src/resolvers/mutation/create-deployment/index.js
```diff
@@ -108,12 +108,28 @@ export default async function createDeployment(parent, args, ctx, info) {
     ...args.properties
   };
 
+  const serviceAccountAnnotations = {};
```
andriisoldatenko, line 111: @schnie  ☝️

config/default.yaml
```diff
@@ -532,3 +532,6 @@ deployments:
     # the KubernetesPodOperator and KubernetesExecutor. This is what signals to apply the
     # required RoleBindings.
     allowPodLaunching: true
+
+  # Annotation name to map the `cloud-role` from creating a deployment to on the Deployment Service account This is the correct value for AWS
+  cloudRoleServiceAccountAnnotationName: eks.amazonaws.com/role-arn
```
schnie, line 537: @andriisoldatenko having EKS baked in as a default feels a little off. Could we default this to null / `~` and _also_ check for the existence of this value down below in the if statements. 

eg: `if (args.cloudRole && serviceAccountAnnotationKey)`

Also what do you think about making this config named `serviceAccountAnnotationKey` to match the usage in the code?

andriisoldatenko: fixed in c3bb906b7e922f57744dd864152a2d8cf214ed74

andriisoldatenko: @schnie PTAL

config/default.yaml
```diff
@@ -532,3 +532,6 @@ deployments:
     # the KubernetesPodOperator and KubernetesExecutor. This is what signals to apply the
     # required RoleBindings.
     allowPodLaunching: true
+
+  # Annotation name to map the `cloud-role` from creating a deployment to on the Deployment Service account This is the correct value for AWS
+  serviceAccountAnnotationKey: ~
```
schnie, line 537:
```suggestion
  # Annotation name to map the `cloud-role` from creating a deployment to on the Deployment Service account
  # AWS - eks.amazonaws.com/role-arn
  # GCP - iam.gke.io/gcp-service-account
  serviceAccountAnnotationKey: ~
```

config/default.yaml
```diff
@@ -532,3 +532,8 @@ deployments:
     # the KubernetesPodOperator and KubernetesExecutor. This is what signals to apply the
     # required RoleBindings.
     allowPodLaunching: true
+
+  # Annotation name to map the `cloud-role` from creating a deployment to on the Deployment Service account
```
schnie, line 536:
```suggestion
  # Annotation name to map the `cloud-role` from creating a deployment to on the Deployment Service account.
```

andriisoldatenko:
@schnie are we good to go?

src/resolvers/mutation/update-deployment/index.js
```diff
@@ -49,10 +49,28 @@ export default async function updateDeployment(parent, args, ctx, info) {
...
+  );
+
   // Munge the args together to resemble the createDeployment mutation.
   // Once we fix the updateDeployment schema to match, we can skip this.
   const mungedArgs = merge({}, updatablePayload, {
-    config: args.config,
+    config: deploymentConfig,
     env: args.env,
     properties: get(args, "payload.properties", {})
   });
```
schnie, line 76: @andriisoldatenko @vishwas-astro what would happen if a user creates the deployment with a cloudRole, but then updates the deployment later via the UI?

I'm thinking that the service account annotations would get wiped out. What can we do to ensure we persist this through an update? Or should we make this on create only and make it immutable? Even then, I think this would get wiped out.

vishwas-astro: I thought the above code is updating serviceAccountAnnotations only if it has any value in it. If UI calls Update Deployment with empty string wont it skip updating it?

schnie: It's because we treat this `args.config` as all or nothing. It just gets completely replaced in this update, so we loose some precision here.

andriisoldatenko: @schnie 
stop we decided to not add it to UI, you can update only using cli. 
i get your point, but as we discussed earlier you proposed to store it in DB, we can try to merge during update existing config with `args.config` or we need new column in this case.

andriisoldatenko: wdyt about this approach?:
```
const deploymentConfig = merge(
    {},
    args.config,
    // Get current serviceAccountAnnotations
    get(deployment.config, "serviceAccountAnnotations"),
    // Store serviceAccount annotations only we have value in it
    !isEmpty(serviceAccountAnnotations) && { serviceAccountAnnotations }
  );
```

andriisoldatenko: added tests
```
(base) ➜  houston-api git:(allow-annotations-to-be-applied-to-pods-via-helm-config-values) ✗ npx jest src/resolvers/mutation/update-deployment/index.unit.test.js
 PASS  src/resolvers/mutation/update-deployment/index.unit.test.js
  updateDeployment
    ✓ typical request is successful (16ms)
    ✓ typical request is successful with cloudRole (5ms)
    ✓ we are not overriding original serviceAccountAnnotations during update without cloudRole (4ms)
    ✓ we are overriding original serviceAccountAnnotations during update with cloudRole exists in args (4ms)

Test Suites: 1 passed, 1 total
Tests:       4 passed, 4 total
Snapshots:   0 total
Time:        2.789s, estimated 3s
Ran all test suites matching /src\/resolvers\/mutation\/update-deployment\/index.unit.test.js/i.
(
```

schnie: If you create a deployment via the CLI with a cloud role, then go adjust a slider in the UI, to call this mutation, does the service account persist in the `config` column?

Maybe this works fine because the client gets a full copy of `config` and resends it back up.

andriisoldatenko: i did some small fixes and more practical tests here results:
```
astro deployment create new-deployment-name2223 --executor=local --cloud-role=test
astro deployment update cka6y923301c90j53rlxe4obl description=example --cloud-role=test-update
```

```
astronomer_houston=# select config from houston$default."Deployment";
                                                config
-------------------------------------------------------------------------------------------------------
 {"executor":"LocalExecutor","serviceAccountAnnotations":{"eks.amazonaws.com/role-arn":"test-update"}}
(1 row)
```

`kubectl get serviceaccounts -n astronomer-earthbound-photon-6643 -o yaml`:

```
- apiVersion: v1
  kind: ServiceAccount
  metadata:
    annotations:
      eks.amazonaws.com/role-arn: test
```

bugfix here b25d8b51c3e3682ec56a326102a3273a06271ac5

andriisoldatenko: @schnie if we need to allow astro cli to update cloudRole i need to add sync=True
https://github.com/astronomer/houston-api/blob/master/src/resolvers/mutation/update-deployment/index.js#L77
please let me know!

schnie: Oh yes, we'll need to add that to CLI

PR #276, opened 2020-05-11
Workspace singleton Resolver

ryanahamilton:
Adds a singleton query for `workspace` so we don't have to fetch the entire array of the workspaces when we just need 1.

I wasn't 100% sure on the appropriate role naming we should use, e.g. `workspace.config.get`, `system.workspace.get`.

Will be following this up with UI PR to utilize this update.


src/resolvers/query/workspace/index.js
```diff
@@ -0,0 +1,38 @@
...
+
+  // Get a list of ids of workspaces that this user has access to.
+  const workspaceIds = ctx.user.roleBindings.map(rb =>
+    rb.workspace ? rb.workspace.id : null
+  );
+
+  // Throw an error if the workspaceUuid is not in the list of ids.
+  if (!hasSystemPerm && !workspaceIds.includes(workspaceUuid)) {
+    throw new PermissionError();
+  }
```
schnie, line 28: @ryanahamilton I think these lines can be omitted. They should be covered by your usage of the `@auth` directive.

Essentially this block here https://github.com/astronomer/houston-api/blob/master/src/lib/directives/auth/index.js#L70-L74

ryanahamilton: So that's what I thought as well… the reason I added this was because the tests I wrote weren't throwing any permissions errors (but I was getting the proper permissions error from the Playground). Maybe I didn't set up my tests properly?

schnie: Oh right. I don't think the auth resolver is even being used in those tests. The tests here are essentially assuming that you have permission to call this resolver and can focus on the resolver logic.

ryanahamilton: Well that explains a lot! I was mimicking the tests off of [some other queries checking for permissions](https://github.com/astronomer/houston-api/blob/master/src/resolvers/query/workspace-service-accounts/index.unit.test.js#L76). While looking at that test, I discovered that it was yielding a false-positive. The test is just checking for error to exist, not a specific error correlating to permissions. A "Cannot read property 'query' of undefined" response is causing to pass.

ryanahamilton: Updated by removing the redundant permission checks and the related test.

schnie: In graphql-land we can typically use the `@auth` directive, which uses the lower-level functions that you had before. Typically we use the lower-level functions directly when we're doing something that isn't an access check on the workspace/deployment id argument. I think we use them in the non-graphql routes too.

PR #277, opened 2020-05-13
Add support for custom claims fallback and ignoring provider.userInfo

ianstanton:
- Add ability to ignore `provider.userInfo` in `userData` request. This allows for compatibility with some auth providers such as ADFS.
- Add optional `claimsFallback` values for `email` and `name`. If standard fields such as `tokenSet.claims.email` and `tokenSet.claims.name` do not exist, this allows to fall back to a field that is present, such as `tokenSet.claims.upn`.
- Example:
```
astronomer:
  houston:
    config:
      auth:
        openidConnect:
          exampleProvider:
            displayName: "Example Provider"
            enabled: true
            fetchUserInfo: false
            claimsMapping:
              email: upn
              name: upn
            clientId: ~
            discoveryUrl: ~
```
- Related to astronomer/issues#1293


config/default.yaml
```diff
@@ -137,6 +137,11 @@ auth:
   openidConnect:
     # Clock skew for slightly out-of-sync clocks
     clockTolerance: 0
+    # Disable userInfo request for providers such as ADFS
```
ashb, line 140: This should be one level deeper when we read it --i.e. under `auth0`/`okta`/`google`. So having these as an example provider is probably the way to go.

src/routes/v1/oauth/callback/post.js
```diff
@@ -41,15 +43,23 @@ export default async function(req, res, next) {
...
     : null;
 
   // Grab user data
   // Some IDPs don't return useful info, so fall back to the claims if we don't have it
-  const userData = merge(await provider.userinfo(tokenSet.access_token), {
-    email,
-    sub: tokenSet.claims.sub,
-    name: tokenSet.claims.name || tokenSet.claims.unique_name
-  });
+  const userData = noUserInfoRequest ?
```
ashb, line 52: This is a little bit hard to read, I'd do it using `let` as follows:

```js
  let userData = {
      email: email,
      sub: tokenSet.claims.sub,
      name: tokenSet.claims.name || tokenSet.claims.unique_name || tokenSet.claims.`${nameFallback}`.toLowerCase()
  };

  if (!noUserInfoRequest) {
    userData = merge(provider.userinfo(token.access_token), userData);
  }
```

Additionally "double negative" (i.e. `!` plus the `no` in the name) should be avoided, so change it to `if (fetchUserInfo)` etc.

We need to work out how to get that config setting set on the provider.

ashb: Look in src/lib/oauth/config/index.js for the `_getIssuer` function. We set `issuer.metadata.authUrlParams` -- we'd want to set these new config fields on the metadata there too.

ashb: And then refer to it as `provider.issuer.metadata.fetchUserInfo` for instance.

config/default.yaml
```diff
@@ -156,12 +157,24 @@ auth:
...
       clientId: ~
       discoveryUrl: https://accounts.google.com
 
     okta:
       enabled: false
+      fetchUserInfo: true
+      clientId: ~
+      discoveryUrl: ~
+
+    exampleProvider:
```
ashb, line 171:
```suggestion
    exampleProvider:
      name: "Example Provider"
```

config/default.yaml
```diff
@@ -165,6 +165,17 @@ auth:
...
       discoveryUrl: ~
 
+    exampleProvider:
+      displayName: "Example Provider"
+      enabled: false
+      # Disable userInfo request for providers such as ADFS
+      fetchUserInfo: false
+      claimsFallback:
+        email: []
+        name: []
```
ashb, line 175: Since we are only allowing a single one (that's fine) this should be a string, not an array.
```suggestion
        email: upn
        # name: []
```

config/default.yaml
```diff
@@ -165,6 +164,17 @@ auth:
       clientId: ~
       discoveryUrl: ~
 
+    exampleProvider:
```
schnie, line 167: Should we have this `exampleProvider` commented out or something, maybe with an additional explanation comment? I guess it kind of feels weird to have this block of actually unused config in here.

ianstanton:
Looking into failed circle-ci test. Moving back into draft until resolved.

ianstanton:
> Looking good, should we add some tests on this stuff?

@schnie Yep, I'll need a hand with this. What test scenarios should be covered?

ianstanton:
> Looking into failed circle-ci test. Moving back into draft until resolved.

Re-ran this and passed checks

src/routes/v1/oauth/callback/post.js
```diff
@@ -41,15 +44,24 @@ export default async function(req, res, next) {
     ? tokenSet.claims.email.toLowerCase()
     : tokenSet.claims.preferred_username
     ? tokenSet.claims.preferred_username.toLowerCase()
+    : get(tokenSet.claims, claimsFallback.email)
+    ? get(tokenSet.claims, claimsFallback.email).toLowerCase()
```
schnie, line 48: @ianstanton we should probably break this triple ternary up so its easier to reason about.

config/default.yaml
ashb: Could do. (It has `enabled: false` so it won't be used.)

src/routes/v1/oauth/callback/post.js
ashb:
```javascript
function getFirstMatchingClaim(claims, ...names) {
  for (let claim of names) {
    const val = get(claims, name);
    if (val !== undefined) return val;
  }
}
```

And then we dan do this as

```suggestion
    const email = _.lowerCase(getFirstMatchingClaim(tokenSet.claims, 'email', 'preferred_username', ...claimsFallback.email))
```

This also supports a list of fallback claims.

src/routes/v1/oauth/callback/index.unit.test.js
```diff
@@ -1,48 +1,273 @@
...
+    const state = `{"provider":"example","integration":"example","origin":"http://houston.local.astronomer.io:8871/v1/oauth/callback"}`;
+    const res = await request(app)
+      .post("/")
+      .send({
+        id_token: idToken,
+        expires_in: expiresIn,
+        state
       });
-    expect(res.statusCode).toBe(200);
+    expect(res.statusCode).toBe(500);
```
ashb, line 271: It seems odd that we should ever expect a 500 error!

ianstanton: Hmm this is what we receive from the test if we use fallback values that don't exist in the claim. Does this mean there is an issue elsewhere? What error code would you expect to see here?

src/routes/v1/oauth/callback/post.js
```diff
@@ -146,5 +140,28 @@ export default async function(req, res, next) {
...
+  let claimsEmail = null;
+
+  if (claims.email) {
+    claimsEmail = claims.email;
+  } else if (claims.preferred_username) {
+    claimsEmail = claims.preferred_username;
+  } else if (get(claims, claimsFallback)) {
+    claimsEmail = get(claims, claimsFallback);
+  }
+  return claimsEmail.toLowerCase();
```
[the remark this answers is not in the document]
ianstanton: Should we throw an error in this case?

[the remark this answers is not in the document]
schnie: Hmm, this logic is starting to feel a little off. Is there a way we could simplify / reduce this to _always_ pull the field names from the configuration?

Right now, we try a couple hard-coded field names, then fallback to the configuration to find an alternatively named field. Could we make it so our existing providers behave the same way and pull these hard-coded values from config?

This seems like it'd reduce the complexity of our Houston logic down to just pre-canned configuration.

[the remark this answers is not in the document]
schnie: So maybe renaming `claimsFallback` to something like `claimsFieldNames` or something better that you can think of.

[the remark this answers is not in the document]
ianstanton: Are you thinking something like this? Azure AD example:
```
auth:
  openidConnect:
    microsoft:
      enabled: false
      clientId: ~
      discoveryUrl: ~
      claimsFieldNames:
        email: "preferred_username"
        name: ~
```

And our function would look like:
```
function getProviderEmail(claims, claimsFieldNames) {
  let claimsEmail = null;

  if (claims.email) {
    claimsEmail = claims.email;
  } else if (get(claims, claimsFieldNames)) {
    claimsEmail = get(claims, claimsFieldNames);
  }
  return claimsEmail && claimsEmail.toLowerCase();
}
```

andriisoldatenko:
@ianstanton to fix scan-trivy you can rebase from master.

src/routes/v1/oauth/callback/post.js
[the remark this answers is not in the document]
schnie: I was thinking it could even be simpler. I don't know if we need the function or not, but pseudo-code like:
```
const fieldNames = get(providerCfg[key], "claimsFieldNames", {});
const email = get(claims, fieldNames.email);
```

We'd have to add this new field to our other default providers through. Does that make any sense?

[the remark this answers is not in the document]
ianstanton: Not sure I understand. Do you mean to pull it out of `src/lib/oauth/config/index.js` where we have ` issuer.metadata.claimsFallback = get(providerCfg[key], "claimsFallback", {});`?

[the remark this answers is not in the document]
ianstanton: And would we hardcode the values for each known provider in our default config?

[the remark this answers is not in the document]
schnie: Yea the ones that we have hard-coded in the actual code can be shifted out into the configuration for that provider to make it part of the normal flow, so we avoid any "fallback" logic.

ianstanton:
@schnie Changes that we worked on yesterday are up.
@ashb Do you remember why we have `tokenSet.claims.unique_name` as a fallback value for `name`?
Seen here in our current code:
https://github.com/astronomer/houston-api/blob/c91b1359807f03cad0d4f995b8b1ff3af899ff2e/src/routes/v1/oauth/callback/post.js#L48-L52

ashb:
> @ashb Do you remember why we have `tokenSet.claims.unique_name` as a fallback value for `name`?
> Seen here in our current code:
> https://github.com/astronomer/houston-api/blob/c91b1359807f03cad0d4f995b8b1ff3af899ff2e/src/routes/v1/oauth/callback/post.js#L48-L52
 
Nope, sorry. I suspect it was that way when I found it and left it alone

config/default.yaml
```diff
@@ -146,22 +146,54 @@ auth:
       # if those integrations are enabled, but do not have connection
       # information.
       enabled: false
+      fetchUserInfo: true
+      claimsMapping:
+        email: email
+        name: name
```
ashb, line 152: These should be defaulted so not needed in the config I'd say.

src/routes/v1/oauth/callback/post.js
```diff
@@ -30,34 +30,44 @@ export default async function(req, res, next) {
...
-    ? tokenSet.claims.email.toLowerCase()
-    : tokenSet.claims.preferred_username
-    ? tokenSet.claims.preferred_username.toLowerCase()
-    : null;
+  const claims = tokenSet.claims;
+  let email = getClaim(claims, claimsMapping, "email");
+  const name = getClaim(claims, claimsMapping, "name");
+  const sub = getClaim(claims, claimsMapping, "sub");
+
+  console.log(name);
```
ashb, line 49: Debug print left in?
```suggestion
```

src/routes/v1/oauth/callback/post.js
```diff
@@ -30,34 +30,44 @@ export default async function(req, res, next) {
...
 
+  if (!email) {
+    // Somehow we got no email! Abort
+    res.sendStatus(400);
```
ashb, line 53: Add a log line here, and perhaps some message to the end user -- they might see this!

src/lib/oauth/config/index.js
```diff
@@ -78,6 +78,8 @@ async function _getIssuer(key, integration = "self", providerName = undefined) {
     DEFAULT_CLIENT_ARGS,
     get(providerCfg[key], "authUrlParams")
   );
+  issuer.metadata.fetchUserInfo = get(providerCfg[key], "fetchUserInfo", true);
+  issuer.metadata.claimsMapping = get(providerCfg[key], "claimsMapping", {});
```
ashb, line 82:
```suggestion
  issuer.metadata.claimsMapping = get(providerCfg[key], "claimsMapping", { email: "email", name: "name" });
```

src/routes/v1/oauth/callback/post.js
```diff
@@ -30,34 +30,44 @@ export default async function(req, res, next) {
...
 
-  const email = tokenSet.claims.email
-    ? tokenSet.claims.email.toLowerCase()
-    : tokenSet.claims.preferred_username
-    ? tokenSet.claims.preferred_username.toLowerCase()
-    : null;
+  const claims = tokenSet.claims;
+  let email = getClaim(claims, claimsMapping, "email");
+  const name = getClaim(claims, claimsMapping, "name");
+  const sub = getClaim(claims, claimsMapping, "sub");
```
ashb, line 47:
```suggestion
  const sub = claim.sub
```

Sub is a required field of OpenID connect, so this one shouldn't be needed.

src/routes/v1/oauth/callback/post.js
ashb: I think it's possible that we don't get an email until _after_ we've called the userinfo endpoint, so we need to move this error to a few lines later.

src/lib/oauth/config/index.js
schnie: @ashb I don't think we need this since the `getClaim` function will fallback to the `name`, which is the standard name for all the claims. It's covered by that second `get` function, defaulting to `name`.

src/routes/v1/oauth/callback/post.js
schnie: Only added this one, just in case someone had something funky here. If its guaranteed to not be different then we can revert it.

src/lib/oauth/config/index.js
ianstanton: Yes, I think it should be covered in `mapped` here:
```
export function getClaim(claims, mapping, name) {
  const mapped = get(mapping, name, name);
  return get(claims, mapped);
}
```
For example:
- If we call `getClaim(claims, { email: "upn", name: "name" }, "email");`, then `mapped` will have the value `upn`. 
- If we call `getClaim(claims, {}, "email");`, then `mapped` will default to the value `email`.

ashb: Ah cool

src/routes/v1/oauth/callback/post.js
```diff
@@ -134,5 +147,6 @@ export default async function(req, res, next) {
   // Respond with redirect to the UI.
   const url = `${ui()}/${state.redirect || "oauth"}?${qs}`;
   const cleanUrl = url.replace(/([^:]\/)\/+/g, "$1");
+
```
schnie, line 150: Seems like an unrelated change? Leftover from debugging?

src/routes/v1/oauth/callback/post.js
```diff
@@ -92,6 +104,7 @@ export default async function(req, res, next) {
       where: { id: userId },
       data: { fullName, avatarUrl }
     });
+    // TODO: mock the prisma.updateUser in the tests
```
schnie, line 107: Remove this change?

schnie:
Was going to merge but looks like some conflicts

ianstanton:
@schnie merge conflicts should be resolved now

andriisoldatenko:
i don't think so, we are still using it 
https://github.com/astronomer/houston-api/blob/main/package.json
cc @ianstanton

andriisoldatenko:
or nvm, i miss read, yeah we can add it to ignore.

PR #279, opened 2020-05-14
Refactor User and Invite queries with input types, Allow query by ID

ryanahamilton:
- Lookup Users and Invites by ID (the rabbit hole led to this refactor). The current inability to do so forces us to use some not ideal patterns in Astro UI to retrieve a single record. PR to refactor those patterns will follow.
- Unifies the User and Invite query variables with [Input Types](https://graphql.org/graphql-js/mutations-and-input-types/) (`input InviteSearch` and `input UserSearch`). This provides a way to require one of multiple optional variables (e.g. `userUuid`, `username`, `email`, or `fullName`). The resolvers are now more robust and have been expanded to accept any of these lookup variables.

Q: Do these schema changes have broader impact, possibly in CLI? There was only one query in use in the UI that was impacted, and I have [a PR open](https://github.com/astronomer/astro-ui/pull/548) to resolve that.

Q: Should our `workspaceUsers()` query allow singleton results when we already have a `workspaceUser()` query for that purpose? Same question for the `users()` and `workspaceInvites()`, but neither of those have singletons yet.

~There are some redundant `userQuery()` and `inviteQuery()` functions and tests utilized in these resolvers. Might make sense to abstract them, but I wasn't sure where an ideal location for them would be?~

Attached to astronomer/issues#1298


samblackk:
@ryanahamilton You can abtrasct redudant functions into https://github.com/astronomer/houston-api/blob/master/src/lib/users/index.js

ryanahamilton:
I've abstracted the duplicated function/tests into shared lib files.

I also looked through the Houston queries in astro-cli and confirmed that none of these with modified schema are in use there.

schnie:
LGTM

PR #284, opened 2020-05-14
Fix Error handling for `createUser` mutation

petedejoy:
This PR fixes the error handling issue described here: https://github.com/astronomer/issues/issues/1173

Because the `hasErrors` export defined in the `throwPrismaError` function was looking for an object structure that would be thrown by a prisma error, it caused Houston to throw a `cannot read property errors of undefined` on the `createUser` route instead of the intended Houston error.

This fix allows the UI to properly queue off of our Houston `InviteTokenEmailError` so that we can properly tell a user that they are trying to login with an email address that is _different_ than the one that is associated with a given signup token.

@ashb assigning you for review since I believe you set up the `errors/prisma.js` stuff.


src/lib/users/index.js
```diff
@@ -69,10 +69,10 @@ export async function createUser(opts) {
         address: email,
         primary: true,
         verified: !!(!emailConfirmation || haveInvite),
-        token: emailToken
-      }
+        token: emailToken,
+      },
     },
-    roleBindings
+    roleBindings,
```
ashb: Did we change our style rules to have trailing commas now? :heart_eyes:

petedejoy: No, just have that as the default applied to all of my personal projects and accidentally set it as a global earlier this week :). Reverted back.

ashb: That said, it appears the only changes in this file are formating related? My normal stance is "no formatting only changes to file, unless that is the _only_ thing that PR is doing"

ashb: Ahhh :sob:

ryanahamilton: We need to standardize these rules across our repos. (We require them in astro-ui, but not here.)

petedejoy: Yeah +1 on that- it's quite challenging to manage a bunch of projects in vscode with different styling rules (or maybe I'm just doing it wrong).

PR #285, opened 2020-05-18
Run correct airflow migrations when user pushes code

andriisoldatenko:
fix https://github.com/astronomer/issues/issues/1213


src/routes/v1/registry/events/post.js
```diff
@@ -94,12 +99,16 @@ export default async function(req, res) {
...
+          data: { config: updatedConfig, airflowVersion }
         })
         .$fragment(
           `{ id config label releaseName extraAu version workspace { id } }`
         );
 
+      const helmValues = merge({}, generateHelmValues(updatedDeployment), {
+        airflowVersion: airflowVersion
+      });
+
```
ashb, line 111:
```suggestion
        .$fragment(
          `{ id config label releaseName extraAu airflwoVersion version workspace { id } }`
        );

      const helmValues = generateHelmValues(updatedDeployment);
```

ashb: (We need to set defaultAirflowTag to affect the init containers, which is what this change makes it do.)

src/routes/v1/registry/events/post.js
```diff
@@ -94,12 +99,16 @@ export default async function(req, res) {
...
           where: { releaseName },
-          data: { config: updatedConfig }
+          data: { config: updatedConfig, airflowVersion }
         })
         .$fragment(
           `{ id config label releaseName extraAu version workspace { id } }`
         );
 
+      const helmValues = merge({}, generateHelmValues(updatedDeployment), {
+        airflowVersion: airflowVersion
```
schnie, line 109: @andriisoldatenko would we want `defaultAirflowTag` here?

https://github.com/astronomer/airflow-chart/blob/master/templates/_helpers.yaml#L98-L101

src/routes/v1/registry/events/post.js
andriisoldatenko: fixed.

src/routes/v1/registry/events/post.js
andriisoldatenko: fixed.

src/lib/deployments/config/index.js
```diff
@@ -629,7 +629,7 @@ export function airflowImages() {
   return config
     .get("deployments.images")
     .slice(0)
-    .sort((a, b) => semver.rcompare(b.version, a.version));
+    .sort((a, b) => semver.rcompare(a.version, b.version));
```
ashb, line 632: You've reversed this sort order. Why?

src/routes/v1/registry/events/post.js
```diff
@@ -94,12 +99,17 @@ export default async function(req, res) {
...
         .updateDeployment({
           where: { releaseName },
-          data: { config: updatedConfig }
+          data: { config: updatedConfig, airflowVersion }
         })
         .$fragment(
           `{ id config label releaseName extraAu version workspace { id } }`
         );
 
+      const helmValues = merge({}, generateHelmValues(updatedDeployment), {
```
ashb, line 108: If you return `airflowVersion` from the previous mutation, then you don't need this merge - it would all trickle up to the right place.

src/lib/deployments/config/index.js
andriisoldatenko: @ashb i cherry-picked commit https://github.com/astronomer/houston-api/pull/285/commits/2c73d07d07fea4ff5c6a3fab971358864074ef39 `Use newest airflow version for default`
cc @schnie

ashb: I would have expected us to just reverse the order in the config - I thought there was a comment somewhere (in helm template?) that would also need updating of we change this sort order

ashb: I worry what this change will do to any enterprise customers who have changed this config already

andriisoldatenko: make sense
do i need to change this merged PR?
https://github.com/astronomer/astronomer/pull/655/files

src/routes/v1/registry/events/post.js
andriisoldatenko: fixed.

src/lib/deployments/config/index.js
schnie: @andriisoldatenko should we undo this one? We still want to select the _oldest_ version by default for now.

src/routes/v1/registry/events/post.js
```diff
@@ -52,6 +52,7 @@ export default async function(req, res) {
         return res.sendStatus(200);
       }
 
+      let airflowVersion;
```
schnie, line 55: @andriisoldatenko using this version we need to select the proper tag and pass it in as `defaultAirflowTag` in the call to commander right?

We'll need to grab the actual configured image tag that corresponds to this version and pass that in.

src/lib/deployments/config/index.js
andriisoldatenko: done.

src/routes/v1/registry/events/post.js
andriisoldatenko: @ashb looks like you suggested wrong change, we have to pass `airflowVersion` to commander values, but using `defaultAirflowTag `.
cc @schnie

andriisoldatenko: nvm.

src/routes/v1/registry/events/post.js
andriisoldatenko: from what i understand it's not necessary, it's done under the hood.
https://github.com/astronomer/houston-api/blob/2dd0cdfb28b5d3d6df0b625d18784d7ee0bab8b3/src/routes/v1/registry/events/post.js#L116 and https://github.com/astronomer/houston-api/blob/2d9337e00b7dd6f3673b50450ef80eb82cee83b1/src/lib/deployments/config/index.js#L393-L392

andriisoldatenko:
practically when i have `deployments.images`:
```
deployments:
  images:
     - version: 1.10.7
       channel: stable
       tag: 1.10.7-alpine3.10-onbuild
     - version: 1.10.10
       channel: edge
       tag: 1.10.10-alpine3.10-onbuild
```

i see in logs:
```
elasticsearch:
    enabled: false
    connection:
        user: universal-corona-4805
        host: localhost
        port: 9200
defaultAirflowTag: 1.10.7-alpine3.10-onbuild
labels:
    platform: deadly-terrier
    workspace: cka5h0fkn002c0735hytvvted
platform:
    release: deadly-terrier
    workspace: cka5h0fkn002c0735hytvvted
executor: CeleryExecutor
```

src/routes/v1/registry/events/post.js
schnie: Oh, ok that looks right. So as long as we pass the updated deployment object, that contains the updated `airflowVersion`, this should work out?

ashb: It doesn't appear that we test this. We should ideally test that, as a result of calling this handler, that `generateHelmValues` is called with the correct/expected value in the `{airflowVersion}` field.

For instance these need updating https://github.com/astronomer/houston-api/blob/master/src/routes/v1/registry/events/index.unit.test.js#L91-L92

We're not currently checking the arguments passed to these, I think we should start.

(We could also possibly mock generateHelmValues so that we don't test that twice.)

ashb:
And that was from you pushing a 1.10.7-based image, right?

andriisoldatenko:
it's first test: create deployment.

src/routes/v1/registry/events/post.js
```diff
@@ -94,7 +99,7 @@ export default async function(req, res) {
       const updatedDeployment = await prisma
         .updateDeployment({
           where: { releaseName },
-          data: { config: updatedConfig }
+          data: { config: updatedConfig, airflowVersion }
         })
         .$fragment(
           `{ id config label releaseName extraAu version workspace { id } }`
```
ashb, line 105:
```suggestion
          `{ id config label releaseName extraAu airflowVersion version workspace { id } }`
```

Since this return is what gets fed into `generateHelmValues`.

andriisoldatenko: fix.

src/routes/v1/registry/events/post.js
andriisoldatenko: fixed.

andriisoldatenko:
1. Create Deployment from Astro UI
(base) ➜  ~ kubectl describe pod solar-galaxy-6053-webserver-5d5f99f99-jng9n -n astronomer-solar-galaxy-6053 | grep Image
    Image:         astronomerinc/ap-airflow:1.10.7-alpine3.10-onbuild
    Image ID:      docker.io/astronomerinc/ap-airflow@sha256:70fd9bcc6db6973f6accd0c908c716af28b3c0d00dc3bb3ff15da2ede016cee8
    Image:         astronomerinc/ap-airflow:1.10.7-alpine3.10-onbuild
    Image ID:      docker.io/astronomerinc/ap-airflow@sha256:70fd9bcc6db6973f6accd0c908c716af28b3c0d00dc3bb3ff15da2ede016cee8

(base) ➜  t_20200519_130630 cat Dockerfile
FROM astronomerinc/ap-airflow:1.10.7-alpine3.10-onbuild%

2. Bump Airflow version using Dockerfile:

(base) ➜  t_20200519_130630 cat Dockerfile
FROM astronomerinc/ap-airflow:1.10.10-alpine3.10-onbuild

(base) ➜  t_20200519_130630 astro deploy
Authenticated to local.astronomer-development.com

Select which airflow deployment you want to deploy to:
 #     LABEL     DEPLOYMENT NAME       WORKSPACE     DEPLOYMENT ID
 1     test      solar-galaxy-6053     w1            ckad1b7tv00oe0j49ib6xzkm9


astronomer_houston=# select "airflowVersion" from houston$default."Deployment" where "releaseName" = 'solar-galaxy-6053';
 airflowVersion
----------------
 1.10.10
(1 row)

PR #288, opened 2020-05-21
Add validation for env var format

adam2k:
The environment variables can only be uppercase A-Z and underscores.

In reference to: https://github.com/astronomer/issues/issues/856


src/lib/deployments/validate/environment/index.unit.test.js
```diff
@@ -8,10 +8,46 @@ describe("When valid environment variables are passed", () => {
...
+    expect(() => {
+      const badEnvs = [
+        { key: "ONE_VALID", value: ""},
+        { key: "TWO_VALID", value: ""},
+        { key: "LAST IS INVALID", value: "" }
+      ];
+      validateEnvironment(badEnvs);
+    }).toThrow();
+  });
+});
```
andriisoldatenko, line 53:
```suggestion
});

```

andriisoldatenko:
@adam2k `npm run lint` fails

andriisoldatenko:
Can you please update codecov report to 77%?

src/lib/deployments/validate/environment/index.js
```diff
@@ -35,3 +40,16 @@ export default function validateEnvironment(envs) {
     throw new InvalidDeploymentError(msg);
   }
 }
+
+function validateEnvironmentVarialbles(envs) {
```
andriisoldatenko, line 44: @adam2k can you please add comment to describe your function?
something like:
https://github.com/astronomer/houston-api/blob/d111c9f4dfb10e8eeac67a23412958a4d3e8b4b3/src/lib/deployments/validate/index.js#L9-L13

src/lib/deployments/validate/environment/index.unit.test.js
```diff
@@ -8,10 +8,46 @@ describe("When valid environment variables are passed", () => {
...
       const badEnvs = [{ key: "AIRFLOW__CORE__EXECUTOR", value: "" }];
       validateEnvironment(badEnvs);
     }).toThrow();
   });
+
+  test("it throws an error for invalid key formats", () => {
+    expect(() => {
+      const badEnvs = [{ key: "", value: "" }];
+      validateEnvironment(badEnvs);
+    }).toThrow();
```
andriisoldatenko, line 22:
```
expect(() => {
      const badEnvs = [{ key: "_", value: "test" }];
      validateEnvironment(badEnvs);
    }).toThrow();
expect(() => {
      const badEnvs = [{ key: "__", value: "test" }];
      validateEnvironment(badEnvs);
    }).toThrow();

```

src/lib/deployments/validate/environment/index.js
schnie: `validateEnvironmentVarialbles` looks like a typo :)

src/lib/deployments/validate/environment/index.unit.test.js
adam2k: Are you suggesting that the env var should not end in an underscore?  OR do you think that the case is we need at least one uppercase character so something like `_A_` is valid?

adam2k: I updated the regex and added tests.  I'm assuming starting and ending with _ is OK.

andriisoldatenko: nope i demonstrated that env var `_` like this is invalid.

src/lib/deployments/validate/environment/index.unit.test.js
```diff
@@ -5,13 +5,81 @@ describe("When valid environment variables are passed", () => {
...
+
+    expect(() => {
+      const badEnvs = [{ key: "_", value: "test" }];
```
andriisoldatenko, line 35: 👍

PR #290, opened 2020-05-23
Add tests for deployment naming

adam2k:


src/lib/deployments/naming/index.js
```diff
@@ -11,15 +11,17 @@ import Haikunator from "haikunator";
...
  */
 export function validateReleaseName(releaseName) {
+  const maxCharLen = 63;
   // Test string pattern
   const pattern = /^(?![0-9]+$)(?!.*-$)(?!-)[a-zA-Z0-9-]{1,63}$/g;
-  if (!pattern.test(releaseName)) throw new InvalideReleaseName();
+  if (!pattern.test(releaseName)) throw new InvalidReleaseName();
 
   // Test total char length
   const prefix = config.get("helm.releaseNamespace");
```
andriisoldatenko, line 20: wdyt if we fallback to default value here?
```suggestion
  const prefix = config.get("helm.releaseNamespace") || "";
```

src/lib/deployments/naming/index.js
```diff
@@ -11,15 +11,17 @@ import Haikunator from "haikunator";
...
-  if (prefix.length + releaseName.length > 63) {
+  // TODO: Figure out what this should do if it prefix is null
+  if (prefix && prefix.length + releaseName.length > maxCharLen) {
```
andriisoldatenko, line 22: and no need to check if `prefix` here

src/lib/deployments/naming/index.js
```diff
@@ -102,6 +102,14 @@ export function generateAirflowUsername(releaseName) {
...
-  return `${rel}_celery`;
+  return `${formatRelName(releaseName)}_celery`;
+}
+
+/**
+ * Convert dashes to underscores for generating release names.
+ * @param  {String} name
+ * @return {String} underscore formatted name.
+ */
+function formatRelName(name) {
```
andriisoldatenko, line 113: may be `dashesTounderscores` or something more meaningful?

adam2k: yeah, good idea.  Should this method be moved into a utils directory too?

src/lib/deployments/naming/index.js
adam2k: Do you think this will cause any downstream issues?  Should the namespace be the default k8s namespace, either `"default"` or "`astronomer`" instead of `""`?

andriisoldatenko: default/astronomer is not a good idea... may be @schnie has ideas in mind.

adam2k: I talked to @schnie and we thought it was best to add a default value into the `test.yaml`.  The releaseNamespace is set when the platform is launched via helm.  I think defaulting that prefix to anything in the code may cause more downstream issues later on.

src/lib/deployments/naming/index.js
```diff
@@ -67,41 +67,47 @@ export function generateDeploymentLabels(platform = {}) {
...
+export function generateCeleryUsername(name) {
+  return `${hyphensToUnderscores(name)}_celery`;
+}
+
+/**
+ * Convert hyphens to underscores for generating correctly formatted release names.
+ * @param  {String} name of a release.
+ * @return {String} underscore formatted name.
+ */
+function hyphensToUnderscores(name) {
```
schnie, line 111: @adam2k lodash has a `snakeCase` function that may serve the same purpose - https://lodash.com/docs/4.17.15#snakeCase.

adam2k: ah!  Good point.  Will remove this.  Thanks!

andriisoldatenko:
LGTM

PR #295, opened 2020-05-27
Add more tests for deployment/logs

adam2k:
In reference to the issue:

fix https://github.com/astronomer/issues/issues/1341


src/lib/deployments/logs/index.js
```diff
@@ -5,20 +5,20 @@ import elasticsearch from "elasticsearch";
...
  * This is a singleton instance of a log generator. It
- * infinately iterates over sample logs.
+ * infinitely iterates over sample logs.
  */
 export const generateLogMessage = (function* generateLogs() {
   const max = sampleLogs.length;
   let counter = 0;
   while (true) {
-    yield sampleLogs[++counter];
+    yield sampleLogs[counter++];
```
adam2k, line 16: `++counter` was skipping the first item in the sample file.

schnie: Whoops

PR #296, opened 2020-05-27
Issue#1339 test coverage validate

kushalmalani:
Adding unit tests for deployment property validation

fix https://github.com/astronomer/issues/issues/1339


src/lib/deployments/validate/properties/index.unit.test.js
```diff
@@ -0,0 +1,24 @@
+import validateProperties from "./index";
+import { InvalidDeploymentError } from "errors";
+
+describe("validateProperties", () => {
+  test("throws error if deployment propery is invalid", async () => {
```
vishwas-astro, line 5: property not spelt properly

schnie:
LGTM. While we're here, I think all the `properties` stuff can probably be refactored out of houston. If you trace other references to `properties` you'll see the rest of it.

This was something the original houston did and we only carried forward to support the original UI with the newer houston. We used to have a separate `deployment_properties` table where we kept 1 or 2 pieces of data about a linked `deployment` record. When we refactored, we just moved those properties to the top-level `deployment` record and dropped the extra table from the schema.

Might be worth ripping this stuff out in a future PR, so we don't have to carry it forward with all the prisma 2 stuff. Just have to make sure the UI and CLI aren't still referencing anything.

schnie:
Saw this comment recently, which got me thinking about it.
https://github.com/astronomer/houston-api/blob/03b7d3c43def9421bf32a57c7264aff93957d654/src/lib/deployments/config/index.js#L466-L488

kushalmalani:
@schnie Sounds good. I can file a ticket so we can keep track of this and a future PR should take care of this.

https://github.com/astronomer/issues/issues/1346. Feel free to edit the description if needed.

PR #297, opened 2020-05-27
Update file name

adam2k:
```diff
rename from src/resolvers/mutation/add-customer-id/index-unit-test.js
rename to src/resolvers/mutation/add-customer-id/index.unit.test.js
```

fix https://github.com/astronomer/issues/issues/1344


andriisoldatenko:
@adam2k can you please add why we need it?
also it would be great to have more readable branch names.
JFYI: issue link you can add in description

PR #306, opened 2020-05-28
Fix/snyk upgrade 781ef96e99ee5fc82e2d626110230ab6

adam2k:
Upgrade apollo-server package.


adam2k:
@samblackk let me know if this is normally how you resolve these issues.  I needed to run `apollo-server@2.14.1 apollo-server-express@2.14.1`.  It updated the packages and the package-lock.  I was able to run the server and access the playground without any errors.  It looks like the package-lock was almost completely changed.  I'm not sure if there's a way around that?

samblackk:
Sweet. The `package-lock` changing it's totally fine - that means that the packages have updated dependencies which is a big thing we're tackling right now.

PR #310, opened 2020-06-01
Rename mutation with more apt name, remove unused argument from schema

ryanahamilton:
Resolves astronomer/issues#1363.

- Renames `updateUser` mutation to more apt `updateSelf`.
- Removes mutation's unused `userId` argument from schema.

Usage updated in astronomer/astro-ui#580. There doesn't appear to be any usage of `updateUser` in Astro-CLI.

Will follow-up with adding to `prisma2` branch.


andriisoldatenko:
@ryanahamilton please make sure you replaced in all places:
https://github.com/astronomer/houston-api/blob/2203c523e7ab01b6bcaaef47f7c7d2c4e5fac74f/src/resolvers/mutation/verify-email/index.js#L40
```
 pt updateUser
./src/resolvers/mutation/verify-email/index.js
40:  await ctx.db.mutation.updateUser({ where, data });
./src/resolvers/mutation/verify-email/index.unit.test.js
42:    const updateUser = jest.fn();
47:      mutation: { updateUser }
53:    expect(updateUser).toHaveBeenCalledTimes(1);
./src/resolvers/mutation/update-self/index.js
15:  return ctx.db.mutation.updateUser({ where, data }, info);
./src/resolvers/mutation/update-self/index.unit.test.js
37:    const updateUser = jest.fn();
41:      mutation: { updateUser }
52:    expect(updateUser.mock.calls).toHaveLength(1);
./src/routes/v1/oauth/callback/post.js
91:    await prisma.updateUser({
```

ryanahamilton:
@andriisoldatenko I don't think we need to change the `updateUser` in the referenced `verifyEmail` resolver—it still utilizes the same db function (similar to how the newly renamed `updateSelf` does as well).

The [reported coverage drop](https://codecov.io/gh/astronomer/houston-api/compare/59d62c2803ca2f60af106826d81acf1c0a930389...7e215bb32bf3908122233eb1b141dd42f1b0e962/src/src/lib/deployments/logs/index.js) occurs in the seemingly unrelated `src/lib/deployments/logs/index.js`. Something in relation to the presence of Elasticsearch.

image

PR #312, opened 2020-06-02
Decouple and track deployment rollout status

schnie:
Draft PR to get some initial thoughts on the deployment rollout process.


package.json
```diff
@@ -49,14 +49,15 @@
     "haikunator": "^2.0.0",
     "handlebars": "^4.6.0",
     "handlebars-layouts": "^3.1.4",
-    "https-proxy-agent": ">=2.2.3",
     "http-status-codes": "^1.3.0",
+    "https-proxy-agent": ">=2.2.3",
```
andriisoldatenko, line 53: any reason of this change?

schnie: Hmm, all I did just `npm install nats -s`, so not sure why.

database/datamodel.graphql
```diff
@@ -135,9 +145,21 @@ type Deployment {
...
+type Rollout {
+  id: ID! @id
+  deployment: Deployment! @relation(name: "DeploymentRollouts")
+  config: Json!
+  version: String!
+  airflowVersion: String!
+  status: RolloutStatus
+  createdAt: DateTime! @createdAt
+  updatedAt: DateTime! @updatedAt
+}
```
ryanahamilton, line 161: I can definitely see wanting to report in the UI the User who created the latest Rollout. I know that has some overlap with the audit logs feature, but I'm wondering if that User ID should be captured in this table?

package.json
adam2k: That same thing happened for me.   They weren't in alphabetical order, so npm updated it. `-` is ahead of `s`.

src/lib/deployments/database/index.js
```diff
@@ -155,20 +155,63 @@ export function createConnection(config) {
  * @param {Object} conn An existing database connection.
  * @param {String} name Name for the new database.
```
andriisoldatenko, line 156: need update comments ^ on line 155

src/workers/deployment-updated/index.js
```diff
@@ -0,0 +1,106 @@
...
+  opts.setManualAckMode(true);
+  opts.setAckWait(300 * 1000);
+  opts.setDurableName(clientID);
+
+  // Subscribe and assign event handler
+  const sub = nc.subscribe("houston.deployment.updated", opts);
+  sub.on("message", function(msg) {
+    deploymentUpdated(msg).catch(err => log.error(err));
+  });
+});
```
schnie, line 36: Would be great to abstract this block of code out into something reusable by all the workers. Would be nice to keep those worker files super scoped to just the important logic and pull this boilerplate out.

adam2k: Yeah, good idea.  We can make a worker factory to generate these handlers.

src/workers/deployment-deleted/index.js
```diff
@@ -0,0 +1,48 @@
...
+import config from "config";
+
+const clusterID = "test-cluster";
+const clientID = "deployment-deleted";
+const subject = "houston.deployment.deleted";
+const messageHandler = function(msg) {
+  deploymentDeleted(msg).catch(err => log.error(err));
+};
+// Create NATS client.
+const nc = ncFactory(clusterID, clientID, subject, messageHandler);
```
schnie, line 15: @adam2k nice, looks great. I wonder if we could take it a small step further. Instead of accepting the `messageHandler`, which is a wrapped function, could the factory accept the actual function defined below, and have it also attach the `catch()` handler. One more piece of boilerplate we could clean out of these files.

adam2k: ah yeah, good idea.  I can make that update.

src/resolvers/mutation/create-deployment/index.js
```diff
@@ -174,40 +150,18 @@ export default async function createDeployment(parent, args, ctx, info) {
...
+  );
+  // Send event that a new deployment was created.
+  const nc = publisher(`create-deployment-${id}`);
+  await nc.on("connect", async () => {
+    nc.publish(DEPLOYMENT_CREATED, id);
+    nc.close();
+    log.info(
+      `Create Deployment published to ${DEPLOYMENT_CREATED} with ID: ${id}`
+    );
+    return Promise.resolve();
```
adam2k, line 164: @andriisoldatenko @schnie this block of code (and similar blocks in the other endpoints) is causing a lot of issues when running the test suite.  I'm still trying to unpack what is happening.  There are tons of console statements when I run `yarn coverage` that look like 

```
/Users/adamvandover/Documents/GitHub/houston-api/node_modules/google-protobuf/google-protobuf.js:81
jspb.BinaryConstants.TWO_TO_32=4294967296;jspb.BinaryConstants.TWO_TO_52=4503599627370496;jspb.BinaryConstants.TWO_TO_63=0x7fffffffffffffff;jspb.BinaryConstants.TWO_TO_64=1.8446744073709552E19;jspb.BinaryConstants.ZERO_HASH="\x00\x00\x00\x00\x00\x00\x00\x00";goog.dom={};goog.dom.NodeType={ELEMENT:1,ATTRIBUTE:2,TEXT:3,CDATA_SECTION:4,ENTITY_REFERENCE:5,ENTITY:6,PROCESSING_INSTRUCTION:7,COMMENT:8,DOCUMENT:9,DOCUMENT_TYPE:10,DOCUMENT_FRAGMENT:11,NOTATION:12};goog.debug={};goog.debug.Error=function(a){if(Error.captureStackTrace)Error.captureStackTrace(this,goog.debug.Error);else{var b=Error().stack;b&&(this.stack=b)}a&&(this.message=String(a));this.reportErrorToServer=!0};goog.inherits(goog.debug.Error,Error);goog.debug.Error.prototype.name="CustomError";goog.asserts={};goog.asserts.ENABLE_ASSERTS=goog.DEBUG;goog.asserts.AssertionError=function(a,b){goog.debug.Error.call(this,goog.asserts.subs_(a,b));this.messagePattern=a};goog.inherits(goog.asserts.AssertionError,goog.debug.Error);goog.asserts.AssertionError.prototype.name="AssertionError";goog.asserts.DEFAULT_ERROR_HANDLER=function(a){throw a;};goog.asserts.errorHandler_=goog.asserts.DEFAULT_ERROR_HANDLER;

Error {
  message: 'Failure: Cannot coerce to Uint8Array: object',
  reportErrorToServer: true,
  messagePattern: 'Failure: Cannot coerce to Uint8Array: object'
}
```

That error message is consistent with errors where I was not correctly closing the NATS client.  I moved these lines of code closer together so we could pinpoint when an error occurs with NATS.  Previously the publisher call was towards the top of this method and publish was towards the bottom.  I was hoping that it is an async issue and by resolving a promise inside of the `on connect` event it would prevent the error, but that isn't working.

adam2k: Reverted this block back.  I think this is unrelated to the other tests that are failing that I'm skipping.

config/default.yaml
```diff
@@ -16,6 +16,23 @@ prisma:
   secret: ~
   debug: false
 
+# NATS configuration.
+nats:
+  host: nats://localhost
+  port: 4222
```
ashb, line 22: This is slightly odd -- given the host has the `nats://` prefix I would have expected it to also include the port (i.e. be `nats://localhoost:4222`, or for host just to be `localhost`

src/lib/deployments/database/index.js
```diff
@@ -155,20 +155,63 @@ export function createConnection(config) {
...
+  // Check if schema exists
+  const result = await conn.raw(
+    `SELECT EXISTS(SELECT schema_name FROM information_schema.schemata WHERE schema_name = '${schema}')`
+  );
+  const exists = first(result.rows).exists;
+  if (!exists) {
+    // Create a new schema, granting the new user access.
+    await conn.raw(`CREATE SCHEMA ${schema} AUTHORIZATION ${user}`);
+  }
+}
```
ashb, line 203: This is racey -- two processes could check it, not find it and both try to create it.
```suggestion
  // Create a new schema, granting the new user access.
  await conn.raw(`CREATE SCHEMA IF NOT EXISTS ${schema} AUTHORIZATION ${user}`);
}
```

schnie: Could we do the user and database creation as `IF NOT EXISTS` too?

ashb: It doesn't look like CREATE USER or DATABASE have the `IF NOT EXISTS` variant.

Perhaps the best approach is to try and create, but catch and ignore "already exists" errors?

ashb: https://stackoverflow.com/questions/8092086/create-postgresql-role-user-if-it-doesnt-exist

```sql
DO
$do$
BEGIN
   IF NOT EXISTS (
      SELECT FROM pg_catalog.pg_roles  -- SELECT list can be empty for this
      WHERE  rolname = 'my_user') THEN

      CREATE ROLE my_user LOGIN PASSWORD 'my_password';
   END IF;
END
$do$;```

config/default.yaml
adam2k: Good point!  I can move that around.

src/lib/deployments/database/index.js
schnie: Yea I think in the POC code, I tried to use `IF NOT EXISTS` on the db creation, and realized that wasn't supported, so kind of copy/pasted the same idea to all three of these create statements (db, user, schema).

I think since this is a _creation_ event, we shouldn't be racing to process two duplicate events for the same deployment, unless we somehow double-publish what should be a single event, which should never happen under normal circumstances. 

In the unlikely event of a double-publish of a single event, and the subsequent race to process them, one of these will ultimately fail and be retried. In which case, it should successfully re-run and end up as a no-op, since we've at least made sure to make the event idempotent. 

I think this type of racy behavior would be more likely to happen on update events since they can happen repeatedly for a single deployment. But, those event handlers don't have anything tricky like this since they are only handling helm upgrades, which can already be re-run cleanly.

adam2k:
I'm going to close this PR.  We have a lot of this code merged from other smaller PRs and most of this is outdated now.

PR #313, opened 2020-06-03
Deployment status subscription returning nothing indicates pending deployment

sjmiller609:
This change is to restore the functionality that when the deployment is coming up, houston does not publish to the client anything for this subscription.


src/resolvers/subscription/deployment-status/index.js
```diff
@@ -51,12 +51,15 @@ export async function subscribe(parent, args, { pubsub }) {
...
       const res = await Promise.resolve(getMetric(releaseName));
-      publish({
-        deploymentStatus: {
-          result:
-            res && res.result && res.result.length ? res.result[0].value[1] : 0
-        }
-      });
+      const pending = res && res.result && res.result.length <= 0;
+      const green = res && res.result && res.result.length > 0;
+      if (!pending) {
```
andriisoldatenko, line 56: i'm not sure that it's best way to do it. I prefer to send both statuses.

sjmiller609: UI behavior is currently:
'no response' = blue
'result: 0' = red
'result: 1' = green

So to follow that suggestion, we need to modify UI

sjmiller609: https://astronomerteam.slack.com/archives/C02NZELT2/p1591213781285300?thread_ts=1591212640.277700&cid=C02NZELT2

sjmiller609: https://astronomerteam.slack.com/archives/C02NZELT2/p1591214205287100?thread_ts=1591212640.277700&cid=C02NZELT2

sjmiller609: I do agree with what you are saying though

ryanahamilton:
Just for some extra context, this subscription logic will be replaced in conjunction with the forthcoming [Deployment rollout status](astronomer/houston-api/pull/312) feature.

src/resolvers/subscription/deployment-status/index.js
```diff
@@ -25,20 +25,46 @@ export function buildURI(query) {
   )}&time=${now}`;
 }
 
-export async function getMetric(releaseName) {
-  const req = await request({
+export function isNotPending(res) {
+  return (
```
adam2k, line 29: I forgot that we could use lodash here

```Suggestion
  return defaultTo(get(res, "data.result[0].value.length"), 0) > 1;
```

src/resolvers/subscription/deployment-status/index.js
```diff
@@ -25,20 +25,46 @@ export function buildURI(query) {
...
+  if (isNotPending(res)) {
+    metric = res.data.result[0].value[1];
+  }
+  return metric;
 }
 
 // Start the subscription
 export async function subscribe(parent, args, { pubsub }) {
   let { releaseName } = args;
+  const metric = await Promise.resolve(getMetric(releaseName));
```
adam2k, line 67: Is the Promise.resolve required here?

src/resolvers/subscription/deployment-status/index.unit.test.js
```diff
@@ -0,0 +1,46 @@
+import { getMetric } from "./index";
+import request from "request-promise-native";
+jest.mock("request");
```
adam2k, line 3: 🎉

src/resolvers/subscription/deployment-status/index.js
sjmiller609: Perhaps redundant with async on getMetric?

src/resolvers/subscription/deployment-status/index.js
```diff
@@ -25,20 +25,46 @@ export function buildURI(query) {
...
-  const req = await request({
+export function isNotPending(res) {
+  return defaultTo(get(res, "data.result[0].value.length"), 0) > 1;
+    res &&
+    res.data &&
+    res.data.result &&
+    res.data.result.length > 0 &&
+    res.data.result[0].value &&
+    res.data.result[0].value.length > 1
+  );
```
adam2k, line 36: You can remove these lines now.

```Suggestion
```

sjmiller609: oops

src/resolvers/subscription/deployment-status/index.js
adam2k: yeah, I think so because it's no longer returning a Promise.

src/resolvers/subscription/deployment-status/index.js
```diff
@@ -25,20 +26,39 @@ export function buildURI(query) {
...
+  let metric = null;
+  const res = await prometheusRequest(`
       rate(airflow_scheduler_heartbeat{deployment=~"${releaseName}", type="counter"}[1m])
-    `)
-  }).catch(err => log.debug(err));
-  return { result: req.data ? req.data.result : [] };
+    `).catch(err => log.error(err));
+  if (!res.data || res.status !== "success") {
+    log.error(
+      `Did not get response from prometheus as expected, we expected to find json with path data.result as an array and status = 'success'. Response was: ${res}`
```
samblackk, line 48: `res` here returns `[object Object]` which doesn't help debugging. This should have been caught and a parsed error object returned before merging this PR.

samblackk:
This PR wasn't vetted against the UI and has caused breaking changes. Let's get better at checking against components using these subscriptions when making changes, please.

sjmiller609:
> This PR wasn't vetted against the UI and has caused breaking changes. Let's get better at checking against components using these subscriptions when making changes, please.

Hey @samblackk I just released it to prod. Can you please let me know what changes are effecting the UI? I expected this change to have no apparent effect to the UI.

src/resolvers/subscription/deployment-status/index.unit.test.js
```diff
@@ -0,0 +1,46 @@
...
+    request.mockReturnValue({
+      status: "success",
+      data: { result: [{ value: [1234, "0"] }] }
+    });
+    const releaseName = "sample-release-1234";
+    const metricResponse = await getMetric(releaseName);
+    expect(request).toBeCalledTimes(1);
+    expect(metricResponse).toEqual("0");
+  });
+  test("that an unsuccessful prometheus response will return null to indicate pending (blue UI bubble). This makes sense because no information should show as pending to the user.", async () => {
```
sjmiller609, line 36: @samblackk ah now I understand why the local dev issue

sjmiller609: this is the behavior changed, i expect

sjmiller609: sorry about that!

PR #325, opened 2020-06-08
Hotfixes from 0.15 cloud release

sjmiller609:
These are the hotfixes that I already included into release-0.15 branch for the cloud prod release


.circleci/config.yml
```diff
@@ -66,16 +66,16 @@ workflows:
...
           filters:
             branches:
               only: '/release-.*/'
       - approve-release:
           type: approval
           requires:
-            - scan
-            - scan-trivy
+            - test
+            - build
```
sjmiller609, line 78: sometimes we need to release a patch quickly, and we should not block on account of a new security vulnerability when considering the situation of an immediate-need hotfix. This will still show red X on PR with new vulns.

src/lib/deployments/config/index.js
```diff
@@ -394,7 +394,7 @@ export function defaultAirflowTag({ airflowVersion }) {
   // Special case: so that if a user pushes a 1.10.5-based image up, we continue to run the 1.10.7 initContainers.
   // Otherwise run the exactly matching version of the initContainer., remove after 1.10.5 gone
   // TODO: remove this logic in future
-  if (airflowVersion === "1.10.5") {
+  if (!airflowVersion || airflowVersion === "1.10.5") {
```
sjmiller609, line 397: airflowVersion is not set directly after an upgrade, and we still want to use 1.10.7

sjmiller609:
I leave it up to code owners to decide if we ought to merge this before or after fixing new CVE

PR #326, opened 2020-06-08
bump alpine to 3.12 in ap-base

andriisoldatenko:


schnie:
Looks like this isn't building right @andriisoldatenko

andriisoldatenko:
@schnie correct, fixed in next commit, looks like there is no python2 🎉

andriisoldatenko:
@schnie PTAL!

andriisoldatenko:
@schnie can i merge now?

PR #328, opened 2020-06-08
Remove auth requirement from AppConfig query

ryanahamilton:
The data returned in AppConfig is not sensitive and is being [requested pre-auth in the UI](https://github.com/astronomer/astro-ui/blob/master/src/containers/AuthContainer/index.jsx#L15-L16). This data isn't sensitive and seems reasonable would be exposed for app usage.


andriisoldatenko:
@ryanahamilton to fix scan-trivy you can rebase from master.

schnie:
Does this fail in the console now?

PR #331, opened 2020-06-09
Unit test cases for workspaces

kushalmalani:
fix astronomer/issues#1365


config/test.yaml
```diff
@@ -17,6 +17,18 @@ roles:
   SYSTEM_VIEWER:
     permissions:
       system.monitoring.get: false
+  
+  #
+  # Workspace Roles
+  #
+  WORKSPACE_VIEWER:
```
schnie, line 24: Just curious, why are we overriding this one in the test yaml? Everything in default.yaml should be available in test mode.

kushalmalani: I just changed this. I was under the impression that test.yaml config was used for test cases. just realised it overrides default.yaml config, so this was pretty much the copy of it. I removed this

src/resolvers/types/workspace/index.unit.test.js
```diff
@@ -39,4 +57,56 @@ describe("Workspace", () => {
     invites(parent, {}, { prisma });
     expect(prisma.inviteToken.findMany.mock.calls).toHaveLength(1);
   });
+
+  test("should return array of flags with workspace capabilities", () => {
+    const parent = { id: casual.uuid };
+    const user = createMockUser();
```
adam2k, line 63: 🎉🚀

src/resolvers/types/workspace/index.unit.test.js
```diff
@@ -39,4 +57,56 @@ describe("Workspace", () => {
...
+    const prisma = {
+      workspace: { findOne: jest.fn().mockReturnValue(workspace) }
+    };
+    const paywallEnabledValue = await paywallEnabled(parent, {}, { prisma });
+    expect(prisma.workspace.findOne.mock.calls).toHaveLength(1);
+    expect(paywallEnabledValue).toBeTruthy();
+  });
+
+  test("should return false if a user is not blocked from viewing their workspace", async () => {
+    const uuid = casual.uuid;
```
adam2k, line 94: This is a nit pick, but why not call this `const id`?  Then you could do `const parent = { id };`

kushalmalani: Good find. Let me change that

PR #332, opened 2020-06-09
Use Airflow 1.10.7 migration for 1.10.6 images

sjmiller609:
This is a bug in Astronomer 0.15.0


sjmiller609:
i'm going to pick into 0.15 release branch and ship to prod immediately if no objections, seems straightforward

PR #336, opened 2020-06-11
Increasing code coverage for query logs

kushalmalani:
Fix https://github.com/astronomer/issues/issues/1407


src/resolvers/query/logs/index.unit.test.js
```diff
@@ -0,0 +1,51 @@
...
+    }
+  };
+
+  test("should return an array of mocked logs", async () => {
+    const log = await logs({}, args, { prisma });
+    for (let i = 0; i < log.length; i++) {
+      expect(log[i].release).toBe(releaseName);
+      expect(log[i].component).toBe(component);
+    }
+  });
```
adam2k, line 29: This is pretty minor, but I think this'll work and make it a little easier to read `item` probably isn't the best name, but something like this:

```Suggestion
log.forEach((item) => {
      expect(item.release).toBe(releaseName);
      expect(item.component).toBe(component);
});
```

src/resolvers/query/logs/index.unit.test.js
```diff
@@ -0,0 +1,51 @@
...
+
+  test("should return an array of mocked logs when no timestamp is sent", async () => {
+    const args = {
+      deploymentUuid,
+      component
+    };
+
+    const log = await logs({}, args, { prisma });
+    for (let i = 0; i < log.length; i++) {
```
adam2k, line 38: Same as the comment above

src/resolvers/query/logs/index.unit.test.js
kushalmalani: Yeah, should be taken care off

PR #342, opened 2020-06-18
Hotfixes from 6/17 incident regarding elevated logging error rate

sjmiller609:


src/lib/deployments/validate/authorization/index.js
```diff
@@ -15,18 +16,29 @@ export default async function validateDeploymentCredentials(
   passwordField
 ) {
   // Return false if no password
-  if (!password) return false;
+  if (!password) {
+    log.error("no password provided");
```
schnie, line 20:
```suggestion
    log.error("No password provided");
```

src/lib/deployments/validate/authorization/index.js
```diff
@@ -15,18 +16,29 @@ export default async function validateDeploymentCredentials(
...
+    return false;
+  }
 
   // Return false is releaseName doesn't look right
-  if (!RELEASE_NAME_PATTERN.test(releaseName)) return false;
+  if (!RELEASE_NAME_PATTERN.test(releaseName)) {
+    log.error(
+      `release name ${releaseName} does not match the RELEASE_NAME_PATTERN`
```
schnie, line 27:
```suggestion
      `Release name ${releaseName} does not match the RELEASE_NAME_PATTERN`
```

src/lib/constants/index.js
```diff
@@ -53,6 +53,6 @@ export const DOCKER_REGISTRY_CONTENT_TYPE =
 export const MEDIATYPE_DOCKER_MANIFEST_V2 =
   "application/vnd.docker.distribution.manifest.v2+json";
 
-export const RELEASE_NAME_PATTERN = /^(?![0-9]+$)(?!.*-$)(?!-)[a-zA-Z0-9-]{1,63}$/g;
+export const RELEASE_NAME_PATTERN = /^(?![0-9]+$)(?!.*-$)(?!-)[a-zA-Z0-9-]{1,63}$/;
```
andriisoldatenko, line 56: @sjmiller609 can you please add test or explain why do we need to change this line?

andriisoldatenko: JFYI:
>The g modifier is used to perform a global match (find all matches rather than stopping after the first match).

andriisoldatenko: skip my message

ashb: I don't see a problem with making this change (since we aren't calling it more than once on the same string), but I also don't see this flag being there causing any problems either.

PR #343, opened 2020-06-18
Mutation fix + schema cleanup

ryanahamilton:
Resolves astronomer/issues#1418.

Schema changes addressed in UI by astronomer/astro-ui#632. I've looked through CLI and found no queries or mutations impacted by these changes.

The issue was resolved by removing the expected `env` argument from the `updateDeployment` mutation—it was not compatible with the new data shape that we've introduced. Additionally, the same argument was removed from the `createDeployment` mutation (that wasn't ever put to use).

Additional schema cleanup was done to remove unnecessary fields and implement full typing of everything related to environment variables (no `JSON` types!).

This also fixes an issue where the `environmentVariables` field on `type Deployment` wasn't returning data.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,11 +21,11 @@ jobs:
     steps:
       - trivy-scan:
           image_name: ap-houston-api
-  publish-master:
+  publish-main:
     executor: docker-executor
     steps:
       - push:
-          comma_separated_tags: "master,latest"
+          comma_separated_tags: "main,latest"
           image_name: ap-houston-api
   publish-release-branch:
     executor: docker-executor
@@ -57,13 +57,13 @@ workflows:
       - scan:
           requires:
             - build
-      - publish-master:
+      - publish-main:
           requires:
             - scan
             - scan-trivy
           filters:
             branches:
-              only: master
+              only: main
       - publish-release-branch:
           requires:
             - test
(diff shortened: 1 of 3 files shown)


src/schema.graphql
```diff
@@ -60,6 +60,12 @@ input UserSearch {
   fullName: String
 }
 
+input UpdatedEnvironmentVariable {
```
ryanahamilton, line 63: I don't love the `UpdatedEnvironmentVariable ` name, but it had to be different than `type EnvironmentVariable`.

andriisoldatenko: but why do we need to similar types with different names?

src/schema.graphql
```diff
@@ -214,27 +220,17 @@ type DeployInfo {
...
 type EnvironmentVariable {
-  releaseName: String!
-  createdAt: DateTime!
-  updatedAt: DateTime!
   key: String,
   value: String,
   isSecret: Boolean,
 }
 
-type DeploymentVariablesResponse {
```
andriisoldatenko, line 225: @schnie cc

src/schema.graphql
andriisoldatenko: @ryanahamilton pls review this message from Greg ->https://github.com/astronomer/houston-api/pull/272#discussion_r437424941

ryanahamilton: We need `input` for mutation input parameters—they cannot use `type`.

src/schema.graphql
ryanahamilton: This shared response type was no longer needed since `deploymentUuid ` and `releaseName` weren't in use.

ryanahamilton:
This branch/PR was unable to change base branch, so it has been replaced by #346 346

PR #344, opened 2020-06-18
Refactor update deployment variables and add nats

adam2k:
This was too big of a refactor, so I didn't to push the changes into the deployments rollouts branch directly.  Let me know what you all think and if we should hold off on these changes.  I can cherry pick the NATS only changes directly into deployment-rollouts.

### DIFF
--- a/src/resolvers/mutation/update-deployment-variables/index.js
+++ b/src/resolvers/mutation/update-deployment-variables/index.js
@@ -19,10 +19,10 @@ const nc = nats.connect("test-cluster", "update-deployment-variables");
  * @param {Object} _ The result of the parent resolver.
  * @param {Object} args The graphql arguments.
  * @param {Object} ctx The graphql context.
- * @return [EnvironmentVariablePayload] The variable payload.
+ * @return [EnvironmentVariable] The environmentVariables.
  */
 export default async function updateDeploymentVariables(_, args, ctx) {
-  const { deploymentUuid, releaseName } = args;
+  const { deploymentUuid } = args;
   const { commander } = ctx;
   const environmentVariables = await updateEnvironmentVariables(
     args,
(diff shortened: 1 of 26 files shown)


src/workers/updated-deployment-variables/index.js
```diff
@@ -0,0 +1,72 @@
...
+};
+// Create NATS client.
+const nc = ncFactory(clusterID, clientID, subject, messageHandler);
+
+/*
+ * Handle a deployment deletion.
+ */
+export async function updatedDeploymentVariables(msg) {
+  const msgJson = msg.getData();
+  const { id, updatedVariables } = JSON.parse(msgJson);
```
schnie, line 27: @adam2k does `updatedVariables` contain the actual blob of all variables here?

Since these can contain sensitive info, and could be a big blob, I wonder if we should calculate the checksum inside the API request, and pass that checksum as the message.

src/resolvers/mutation/update-deployment-variables/index.js
```diff
@@ -131,11 +116,65 @@ export async function mergeEnvVariables(currentVariables, newVariables) {
...
+ * @return {String} Message to publish to NATS
+ */
+function formatNcMsg(id, environmentVariables) {
+  const msg = {
+    id,
+    environmentVariables
+  };
+
+  return JSON.stringify(msg);
+}
```
schnie, line 180: I wonder if we should make all our messages JSON :thinking:

adam2k: I was getting errors when I tried that.  It looks like these are the choices for `nc.publish` (because it's a network request):
```publish(subject: string, data?: Uint8Array|string|Buffer, callback?:AckHandlerCallback): string;```

src/workers/updated-deployment-variables/index.js
adam2k: Yeah, I wasn't sure how many vars people are typically set.  I'll look into making that change.

adam2k: We'd probably need to encrypt this with an RSA key so we can decrypt on the other end.  If we hash this we can't do anything with it because we're not storing any of the env vars in a database that we can query later, right?

schnie: The env vars are stored in a Kubernetes `Secret`, via the synchronous call to Commander from the API layer, which persists the "desired state.

At this point, the new variables are written, but mounted by the pods that need to get the changes. To roll this out, we take the checksum and pass it into an annotation on these pods, via a helm value - `airflowPodAnnotations` (https://github.com/astronomer/houston-api/blob/main/src/resolvers/mutation/update-deployment-variables/index.js#L91).

Once this is applied, the pods that mount these env vars will reboot, mounting the new updated env vars. The backend worker takes care of actually rolling the change out, just like the rest of the calls. The difference here is that the backend store is a kube secret instead of Postgres, and that we have to pass some additional metadata about the original `args` that does not get stored in the database - `checksum`.

adam2k:
I'm planning on coming back to this PR after the initial NATS release.

PR #346, opened 2020-06-19
Mutation fix + schema cleanup

ryanahamilton:
(Replacement PR for #343 that was borked by change of base branch from master to main.)

Resolves astronomer/issues#1418.

Schema changes addressed in UI by astronomer/astro-ui#632. I've looked through CLI and found no queries or mutations impacted by these changes.

The issue was resolved by removing the expected `env` argument from the `updateDeployment` mutation—it was not compatible with the new data shape that we've introduced. Additionally, the same argument was removed from the `createDeployment` mutation (that wasn't ever put to use).

Additional schema cleanup was done to remove unnecessary fields and implement full typing of everything related to environment variables (no `JSON` types!).

This also fixes an issue where the `environmentVariables` field on `type Deployment` wasn't returning data.


src/lib/deployments/environment-variables/index.js
```diff
@@ -0,0 +1,46 @@
...
+  const { releaseName } = args;
+
+  const namespace = generateNamespace(releaseName);
+  const secretName = generateEnvironmentSecretName(releaseName);
+
+  // Get variable values from commander
+  const commanderValues = await ctx.commander.request("getSecret", {
+    namespace: namespace,
+    name: secretName
+  });
```
adam2k, line 19: This is a nit pick, but I don't think this is too confusing calling `secretName` `name` because it is only used in one spot after the `generateEnvironmentSecretName` call.  At the very least you could delete `: namespace` on line 17.

```Suggestion
  const namespace = generateNamespace(releaseName);
  const name = generateEnvironmentSecretName(releaseName);

  // Get variable values from commander
  const commanderValues = await ctx.commander.request("getSecret", {
    namespace,
    name
  });
```

src/lib/deployments/environment-variables/index.js
```diff
@@ -0,0 +1,46 @@
...
+      ],
+      "[]"
+    )
+  );
+  // Build payload array for results and strip out secrets
+  const vars = map(variables, variable => ({
+    key: variable.key,
+    value: includes(secretKeys, variable.key) ? "" : variable.value,
+    isSecret: includes(secretKeys, variable.key)
+  }));
```
adam2k, line 41: I think this could be moved into a separate method to keep things a little cleaner here, so you could do something like (`cleanedVars` instead of `vars`).
```
const cleanedVars = removeSecretValues(commanderValues);
```
And in that method you could move the `variables` and `secretKeys` declarations and return the `map(variables, variable => ({...`

andriisoldatenko: +1

ryanahamilton: @adam2k pushed some refactoring to address this!

src/lib/deployments/environment-variables/index.js
```diff
@@ -0,0 +1,54 @@
...
+  // Get variable values from commander
+  const commanderValues = await ctx.commander.request("getSecret", {
+    namespace,
+    name
+  });
+  const variables = objectToArrayOfKeyValue(
+    get(commanderValues, "secret.data")
+  );
+
+  const cleanVariables = removeSecretValues(commanderValues, variables);
```
adam2k, line 49: This looks good!

This is a nit pick:
I could see this working either as-is or moving `variables` into the `removeSecretValues` method so you only have a single param for `removeSecretValues`.  From the function signature it's not completely clear why `variables` is passed in as the second param.  At the same time, I get why you don't want to do have too many actions in the `removeSecretValues` method.

src/resolvers/mutation/update-deployment-variables/index.js
```diff
@@ -101,19 +102,21 @@ export default async function updateDeploymentVariables(parent, args, ctx) {
...
-    deploymentId: deploymentUuid,
-    config,
-    env,
-    payload
+  track(ctx.user.id, "Updated Deployment Variables", {
+    deploymentId: deploymentUuid
   });
 
+  // Remove secret values from response
+  const cleanVariables = map(updatedVariables, v => ({
```
adam2k, line 110: Could this `cleanVariables` also use `removeSecretValues` like how you did `sortVariables`?  I missed that this was similar on the first pass through.

src/lib/deployments/environment-variables/index.js
ryanahamilton: No, your suggestion is ideal, I had just missed that one. Just updated.

src/resolvers/mutation/update-deployment-variables/index.js
ryanahamilton: I certainly was considering that, but opted not to given the other instance derives the `isSecret` with `includes(secretKeys, v.key)`. It would require reconstructing a new array prior to passing into `cleanVariables`.

adam2k: ah okay, yeah.  They are slightly different.  I'll mark as resolved.

andriisoldatenko:
@ryanahamilton do you know why there is no codecov report?

ryanahamilton:
@andriisoldatenko I just figured it out and updated the branch in Codecov:

image

ryanahamilton:
@andriisoldatenko here's the coverage diff: https://codecov.io/gh/astronomer/houston-api/compare/main...deploy_var_schema

andriisoldatenko:
@ryanahamilton i can merge, can you confirm if it practically works in KinD?

ryanahamilton:
@andriisoldatenko This was all tested in kind last week. The small refactors from today were the only thing since then

PR #349, opened 2020-06-22
Upgrading openid client package to 3.15.3

kushalmalani:
fix astronomer/issues#1442


kushalmalani:
This is part1 of auth0 nonce implementation. Upgrading the package to 3.15.3. openid has introduced typescript validations and have deprecated some functions.

config/test.yaml
```diff
@@ -45,7 +45,7 @@ auth:
       # if those integrations are enabled, but do not have connection
       # information.
       enabled: true
-      clientId: ~
+      clientId: "fake-auth0-clientId"
```
andriisoldatenko, line 48: AFAIK you can change value directly in tests, it's not. So in future we can be in safe if somebody changes default values in `test.yaml`

src/lib/oauth/config/index.js
```diff
@@ -60,24 +59,29 @@ export async function getClient(name) {
   }
 
   if (!issuer) issuer = await _getIssuer(name);
+  const clientMetadata = new issuer.Client();
+
+  const client = new issuer.Client({
+    displayName: providerCfg[name].displayName || upperFirst(name),
+    ...clientMetadata
+  });
```
adam2k, line 67: This is pretty minor, but:
```Suggestion
  const displayName = providerCfg[name].displayName || upperFirst(name);
  const client = new issuer.Client({
    displayName,
    ...clientMetadata
  });

```

src/lib/oauth/config/index.js
```diff
@@ -60,24 +59,29 @@ export async function getClient(name) {
...
 async function _getIssuer(key, integration = "self", providerName = undefined) {
-  const issuer = await Issuer.discover(providerCfg[key].discoveryUrl);
-
-  issuer.metadata.name = key;
-  issuer.metadata.authUrlParams = merge(
-    {},
-    DEFAULT_CLIENT_ARGS,
-    get(providerCfg[key], "authUrlParams")
-  );
+  const metadata = await Issuer.discover(providerCfg[key].discoveryUrl);
```
adam2k, line 75: It seems like `providerCfg[key]` or `providerCfg[name]` is used a lot in each of these functions.  It might be easier to read if at the top of each function where you use this variable that you declare it with something like `const provider = providerCfg[key];`. It doesn't have to be provider.  Whatever you think the best name would be.

Then after that you can use `provider` instead of using the `key/name` reference.

src/lib/oauth/config/index.js
```diff
@@ -100,7 +104,7 @@ function subclassClient(issuer, clientId, integration, providerName) {
     authUrl(state) {
       const params = merge({}, this.issuer.metadata.authUrlParams, {
         redirect_uri: this.oauthRedirectUrl(),
-        nonce: shortid.generate(),
+        nonce: generators.nonce(),
```
adam2k, line 107: This isn't really something that we have to do here because it won't be uniform, but I was going to mention trying to use implicit values here for `nonce`, `provider` and `origin`.

But I'm thinking that we should come up with a single directory where we do all these transforms so we can kind of hide all of these snake case keys that we need to use for API calls.  Maybe in this actually belongs in the mutations directory?  Not something I'm asking you to do here, but let's figure it out soon.

src/resolvers/types/auth-config/index.unit.test.js
```diff
@@ -35,6 +36,17 @@ describe("Auth Config", () => {
 
   test("should get list of all supported auth providers", async () => {
     const url = "http://astronomer.io";
+    const clientId = "fake-google-clientId";
+    const displayName = "google";
+    const metadata = {
+      displayName
+    };
```
adam2k, line 43: 👍

src/routes/v1/oauth/callback/post.js
```diff
@@ -31,25 +31,25 @@ export default async function(req, res, next) {
...
     // Don't validate the nonce. Not great, but we don't store the nonce in a
     // session right now, so we can't validate this
     nonce: null
   });
 
-  const email = tokenSet.claims.email
-    ? tokenSet.claims.email.toLowerCase()
-    : tokenSet.claims.preferred_username
-    ? tokenSet.claims.preferred_username.toLowerCase()
+  const email = tokenSet.claims().email
```
adam2k, line 41: Any way to use lodash `get` to clean this up a little bit?  

Also, maybe above this line doing something like `const claims = tokenSet.claims();` would help to?  Either `claims` or `user`?  It seems like claims() is actually a user because there is an `email` and `username`.

src/routes/v1/oauth/callback/post.js
```diff
@@ -31,25 +31,25 @@ export default async function(req, res, next) {
...
     : null;
 
   // Grab user data
   // Some IDPs don't return useful info, so fall back to the claims if we don't have it
   const userData = merge(await provider.userinfo(tokenSet.access_token), {
     email,
-    sub: tokenSet.claims.sub,
-    name: tokenSet.claims.name || tokenSet.claims.unique_name
+    sub: tokenSet.claims().sub,
+    name: tokenSet.claims().name || tokenSet.claims().unique_name
```
adam2k, line 52: Same idea here with the `tokenSet.claims()`

src/lib/oauth/config/index.js
```diff
@@ -60,24 +59,29 @@ export async function getClient(name) {
...
-
-  issuer.metadata.name = key;
-  issuer.metadata.authUrlParams = merge(
-    {},
-    DEFAULT_CLIENT_ARGS,
-    get(providerCfg[key], "authUrlParams")
-  );
+  const metadata = await Issuer.discover(providerCfg[key].discoveryUrl);
+  const issuer = new Issuer({
+    name: key,
```
adam2k, line 77: Would it be better to rename the `key` param to `name`?

ashb:
This PR should be targeting main, not prisma2 please.

kushalmalani:
@ashb Just got confirmation from Vishwas, to merge this in main branch. Changing the destination

kushalmalani:
Resolving conflicts

kushalmalani:
Declining this PR. Another PR is up addressing these changes - https://github.com/astronomer/houston-api/pull/350

PR #350, opened 2020-06-24
Upgrading open id client package

kushalmalani:
fix astronomer/issues#1442


src/lib/oauth/config/index.js
```diff
@@ -100,19 +108,23 @@ function subclassClient(issuer, clientId, integration, providerName) {
...
 
     authUrl(state) {
+      const redirect_uri = this.oauthRedirectUrl();
+      const nonce = generators.nonce();
+      const provider = providerName || this.issuer.metadata.name;
+      const origin = oauthUrl();
       const params = merge({}, this.issuer.metadata.authUrlParams, {
-        redirect_uri: this.oauthRedirectUrl(),
-        nonce: shortid.generate(),
+        redirect_uri,
```
adam2k, line 116: This is minor, but I think in these cases where we have to do transforms we should stick with camel case variables.  

`redirect_uri: redirectUri`

It kind of breaks the rule of the implicit value setting, but if other engineers see snake case variables they might start using them.  Especially because some engineers are working on Python here where it's the norm to use snake case.

I think we should figure out a centralized location for these transforms though (like another file within each directory where we do this transformations).

src/resolvers/types/auth-config/index.unit.test.js
```diff
@@ -0,0 +1,59 @@
...
+    const signupUrl = "https://astronomer.io";
+    const config = externalSignupUrl();
+    expect(config).toBe(signupUrl);
+  });
+
+  //Enable this test once we move to prisma2
+  test.skip("should return true if there are no initial signups", () => {
+    const parent = { id: casual.uuid };
+    const prisma = {
+      user: { findMany: jest.fn().mockReturnValue([]) }
```
andriisoldatenko, line 27: you are mocking prisma2 `findMany ` it doesn't exists in prisma1

kushalmalani: Yeah, I skipped that test and added a comment to enable the test once we move to prisma2 on main.

I imported my work from the previous PR (which i declined) and saw this test case was not ready for main. Hence skipping it for now.

andriisoldatenko: why not adopt to prisma1?

kushalmalani: @andriisoldatenko I added test cases for InitialSignups with prisma1

andriisoldatenko: can we remove this one instead of skipping?

PR #351, opened 2020-06-24
Fix environment variable key validation

ryanahamilton:
Resolves astronomer/issues#1455.

- Moves the usage (and file location) of the variable key validation into the `updateDeploymentVariables` mutation so it is actually utilized.
- Updates the validation to allow numeric characters in the key (e.g. the `3`  in `AIRFLOW__CORE__ENCRYPT_S3_LOGS`). Adds a test to confirm this.


andriisoldatenko:
@ryanahamilton please add more tests especially negative ones. 
AFAIK you can’t create 3TEST_ENV in Linux/unix.

src/lib/deployments/environment-variables/validate/index.js
```diff
@@ -50,14 +50,14 @@ function validateHelmOverrides(envs) {
  * @param {[]Object} An array of environment variable key/value pairs
  */
 function validateEnvironmentVariables(envs) {
-  const matcher = /^(?=.*[A-Z])([A-Z_]+$)/g;
+  const matcher = /^(?=.*[A-Z0-9])([A-Z0-9_]+$)/g;
```
adam2k, line 53: Is it all right if the env var starts with a number?

ryanahamilton: No, can't be a number. I'm currently working on updating the validation and tests to fix this.

ryanahamilton:
@adam2k helped me work out an updated regex that allows numbers AND disallows leading with a number. I added additional test cases to confirm.

src/lib/deployments/environment-variables/validate/index.js
andriisoldatenko: >Environment variable names used by the utilities in the Shell and Utilities volume of IEEE Std 1003.1-2001 consist solely of uppercase letters, digits, and the '_' (underscore) from the characters defined in Portable Character Set and do not begin with a digit. 

https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap08.html

ryanahamilton: We have fixed it to follow those patterns. The only cases that are technically allowed, but we're disallowing are strings of multiple, only underscores (`__`, `___`).

ryanahamilton:
We need to fix the Codecov PR integration, but this PR nets a `+0.05%`.

https://codecov.io/gh/astronomer/houston-api/compare/main...fix/key_valdation

andriisoldatenko:
@schnie @ashb can you please recreate integration with codecov... looks like we loose ability to see comments.

PR #352, opened 2020-06-25
Add nats for v1 registry event

adam2k:
## Description

Add NATS for the v1/registry/event route to handle offline execution of `commander.updateDeployment` requests to prevent potential timeouts for clients.

## 🎟 Issue(s)

Resolves https://github.com/astronomer/issues/issues/1195

## 🧪 Functional Testing

Start Houston and send a request (update the deployment values to match real values for your local .env) like this: 

```
curl --location --request POST 'http://app.local.astronomer.io:8871/v1/registry/events?Content-Type=application/vnd.docker.distribution.events.v1+json' \
--header 'Content-Type: application/vnd.docker.distribution.events.v1+json' \
--header 'Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1dWlkIjoiY2tka29qbHF5MDAxcDA3NDg1aXNidmgxMyIsImlhdCI6MTU5NzY5MDcyOCwibmJmIjoxNTk3NjkwNzI4LCJleHAiOjE1OTc3NzcxMjh9.XHfkrAKTv5NiTJ6oqsF5c30cosu51VyXGiatulP0IXA' \
--data-raw '{
    "events": [
        {
            "id": "4402006a-b771-4a32-9e2b-949dc4c64135",
            "action": "push",
            "target": {
                "repository": "scientific-aurora-8260/airflow",
                "tag": "0.18.0"
            },
            "request": {
                "host": "Klein.net"
            }
        }
    ]
}'
```

## 📸 Screenshots

N/A

## 📋 Checklist


src/workers/nats-factory/index.js
```diff
@@ -0,0 +1,32 @@
+import nats from "node-nats-streaming";
```
schnie, line 1: @adam2k nitpick but what do you think about moving any of the shared/library code under the `src/lib` directory?

Ideally this is where code that is agnostic of entrypoint (resolvers/routes/workers/migrations/etc) can live here.

src/lib/constants/index.js
```diff
@@ -56,3 +56,5 @@ export const MEDIATYPE_DOCKER_MANIFEST_V2 =
 export const RELEASE_NAME_PATTERN = /^(?![0-9]+$)(?!.*-$)(?!-)[a-zA-Z0-9-]{1,63}$/g;
 export const RELEASE_NAME_AIRFLOW_PATTERN = /^\/((?![0-9]+$)(?!.*-$)(?!-)[a-zA-Z0-9-]{1,63})\/(airflow|flower)/;
 export const VALID_DOCKER_IMAGE_NAME = /^(?<releaseName>(?![0-9]+$)(?!.*-$)(?!-)[a-zA-Z0-9-]{1,63})\/airflow$/;
+
+export const REGISTRY_EVENT_UPDATED = "v1.registry.event.updated";
```
schnie, line 60:
```suggestion
export const REGISTRY_EVENT_UPDATED = "houston.deployment.image.updated";
```

@adam2k what do you think about this to keep consistency with the other names here https://github.com/astronomer/houston-api/pull/312/files#diff-e21e41780a16c3368512192a60551ee6R60-R63

src/workers/v1-registry-event/index.js
```diff
@@ -0,0 +1,64 @@
+import { prisma } from "generated/client";
```
schnie, line 1: If we rename the event, we could probably also rename this directory to match - `deployment-image-updated`

src/workers/nats-factory/index.js
adam2k: yeah, I think that would work better.  I can create a `src/lib/workers` directory and move the factory there.  That'll probably clean up the main workers directory some.

schnie:
Thinking through a few failure scenarios here.

- What happens to the API layer when NATS is offline?
- What happens to the workers if NATS is offline?
- What happens if we:
  1) take the workers offline
  2) kick multiple jobs (multiple changes) into NATS from serving layer
  3) bring the workers back up
  Do we end up at the desired state?

- Are there any situations where a job would never be able to complete?
- How would a retry happen if the database was not accessible when running the job?

andriisoldatenko:
@schnie more tech questions:
- how are we going to run worker?
i suggested to add:
>+    "worker": "babel-node src/workers/deployment-image-updated/index.js",
and change global to local in:
```
const nc = natsPubSub(clientID, subject, helmUpdateDeployment);
```

and run it similar we do for another scritps?
```
+// When a file is run directly from Node, require.main is set to its module.
+if (require.main === module) {
+  nc = client();
```

andriisoldatenko:
also what do you think about current implementation respect 1 worker per subject?
so in k8s we need to ability to scale/release not 1 worker but `number of workers` == `number of subjects`?

adam2k:
> also what do you think about current implementation respect 1 worker per subject?
> so in k8s we need to ability to scale/release not 1 worker but `number of workers` == `number of subjects`?

I need to research what the best approach is more.  I was thinking having each of these workers running in their own pods will give us the most ability to scale rapidly, but that might be too many resources being wasted for workers that aren't used too much.  

Is there a way to have the workers come up like spot instances when they are required almost like google functions?

adam2k:
> @schnie more tech questions:
> 
>     * how are we going to run worker?
>       i suggested to add:
> 
> 
> > * "worker": "babel-node src/workers/deployment-image-updated/index.js",
> >   and change global to local in:
> 
> ```
> const nc = natsPubSub(clientID, subject, helmUpdateDeployment);
> ```
> 
> and run it similar we do for another scritps?
> 
> ```
> +// When a file is run directly from Node, require.main is set to its module.
> +if (require.main === module) {
> +  nc = client();
> ```

Maybe we can have an index file that includes all these modules somewhere?  Then we can centeralize the `if (require.main === module) {` block?

schnie:
In terms of scaling, I was just thinking we could have a single kube `Deployment` that runs the new entrypoint (as discussed above). Each replica could subscribe to all topics and handle all the jobs we have for now. These jobs don't really do anything too intensive and have been running alongside each other synchronously already.

I feel like this is a nice simple way to start out, so that we don't need to manage too many worker `Deployment`s.

@adam2k great point, yea we could eventually autoscale these workers somehow. KEDA seems to have an out of the box autoscaler for NATS. We already have KEDA running, so this could be a good follow up extension, assuming everything else works.

adam2k:
> Thinking through a few failure scenarios here.
> 
>     * What happens to the API layer when NATS is offline?
> 
>     * What happens to the workers if NATS is offline?
> 
>     * What happens if we:
>       
>       1. take the workers offline
>       2. kick multiple jobs (multiple changes) into NATS from serving layer
>       3. bring the workers back up, do we end up at the desired state? whats the impact on the deployment? how fast are sequential updates applied?
> 
>     * Are there any situations where a job would never be able to complete?
> 
>     * How would a retry happen if the database was not accessible when running the job?

I'll need to see how we can test some of these cases.  It might be possible to test with mocking failures during the unit tests.

adam2k:
I'm running into an issue splitting the worker into it's own process.  When I run `docker-compose up -d`  and then `npx babel-node src/workers/deployment-image-update/index.js` running the worker cannot connect on port 4222.  When I put it on a different port the worker will connect to NATS, but then when I run `houston-api` and publish the event (I'm using Create New Deployment for testing) then the subscriber in the worker doesn't pick up on the event.  The subject name is the same on both and it works when I'm running it all within `houston-api`.  This is only a problem when I'm running with the babel-node command as a separate job.

adam2k:
>     * What happens to the API layer when NATS is offline?

- If I take down NATS while the API layer is up and the worker is running there is no immediate error in the API until a new request comes in.  The message would be pushed into the NATS server and as soon as the worker comes back online it would process all of the tasks in that particular channel.
- Side note: there is a monitoring endpoint at (`:8222/routez`) that we can use for health checks to make sure that NATS itself is up.  We'd need to probably write our own health checks for the workers.

>     * What happens to the workers if NATS is offline?

- If NATS is down when the worker starts the worker will crash and the `routez` endpoint will not be available.
- If NATS goes down while the worker is running the worker stays up for ~15 seconds and then shuts down because the connection to NATS is lost.

>     * What happens if we:
>       
>       1. take the workers offline

- Out of the box the NATS server keeps everything queued in memory.  When the workers come back online all the jobs are replayed and acked if completed correctly. 

>       2. kick multiple jobs (multiple changes) into NATS from serving layer
- This is a constraint that we'll need to put on code reviews.  The safest way to handle this is to store all the request values in a DB and then ONLY pass the ID to the NATS client that will complete the job.  This way we aren't using a lot of NATS server memory and we can reach back into our DB to replay the changes.  With this in mind we won't have collisions in data because we'll have ACID compliant transactions at the DB level.  The only eventually consistent piece is during deployment.  We might have several deployments queued up with only a few minor changes from each other.  
- Side note:  there could be problems in how we're currently handling the ENV vars for deployments because we aren't storing them anywhere on our systems.

>       3. bring the workers back up, do we end up at the desired state? whats the impact on the deployment? how fast are sequential updates applied?
- The sequential updates are immediate and extremely fast.  We would need to bench mark to get more accurate, but it doesn't seem like we would run into issues until we start to run up against using all the the NATS server memory with the queues.

>     * Are there any situations where a job would never be able to complete?
- If the entire NATS cluster is down and we don't have a script and/or DB flag to resend all the deployments that haven't run yet then we would have issues.
- Maybe we can add a script to run when NATS goes down and comes back up to ensure all the tasks are not in a pending state?
>     * How would a retry happen if the database was not accessible when running the job?
- As long as we have manual ack enabled that means we MUST `nc.ask()` or the job will keep retrying about every 5 seconds and will continue to redeploy.  I believe there is some kind of configuration to prevent too many retries.

schnie:
Yea if we can put some kind of cap on the re-deliveries, that'd be great. Otherwise we'll have to handle it some other way in our code.

@adam2k @andriisoldatenko we can deploy NATS as a `StatefulSet` and use the disk instead of memory for NATS right? If the workers are offline for whatever reason, I don't think we should loose any messages.

Regarding the secrets, we have another issue over [here](https://github.com/astronomer/issues/issues/1409) to clean it up. You are right, and this is a big problem right now for us. Many support tickets are stemming from this secret issue.

How do ya'll think we should handle the case where NATS is down and a request comes in from the API serving layer? If the database is down, there'd probably be an error persisting the request, which is fine. If we never send the notification because NATS is down, what should we do? In this case, we may have successfully saved their request, it just won't be propagated to actually run.

src/workers/deployment-image-update/index.js
```diff
@@ -0,0 +1,90 @@
...
+  const subject = DEPLOYMENT_IMAGE_UPDATED;
+  try {
+    // Create NATS PubSub Client
+    nc = natsPubSub(clientID, subject, queueGroup, helmUpdateDeployment);
+    log.info("NATS Deployment Update Worker Running...");
+    return nc;
+  } catch (err) {
+    log.error(err);
+  }
+}
```
schnie, line 27: @adam2k maybe to make this a little more functional, we could so something like `const nc = createClient()`, where `createClient()` is what you have as the `deploymentImageUpdate()` function. Instead of setting the top-level scoped client, we could have this function return the client for the rest of the module to use, which it seems to do already.

src/workers/deployment-image-update/index.js
```diff
@@ -0,0 +1,90 @@
...
+
+let nc = {};
+
+/**
+ * NATS Deployment Update Worker
+ */
+function deploymentImageUpdate() {
+  const clientID = "deployment-image-update";
+  const queueGroup = "houston-api";
+  const subject = DEPLOYMENT_IMAGE_UPDATED;
```
schnie, line 18: @adam2k I know I'm super nitpicking now, but maybe we could push these lines into where they are used. I'm just trying to keep this code super clean and concise since it'll be copy/pasted elsewhere :)

schnie: @adam2k would every one of our workers have `houston-api` as the queueGroup?

adam2k:
It looks like we might need to handle re-deliveries differently.  I don't see a way around it in the NATS config.  https://docs.nats.io/nats-server/configuration#connection-timeouts

That sounds good for`StatefulSet`.  We can start here: https://docs.nats.io/nats-on-kubernetes/minimal-setup#ha-setup-using-statefulsets

> How do ya'll think we should handle the case where NATS is down and a request comes in from the API serving layer? If the database is down, there'd probably be an error persisting the request, which is fine. If we never send the notification because NATS is down, what should we do? In this case, we may have successfully saved their request, it just won't be propagated to actually run.

With the HA cluster in place this is definitely a corner case.  Thinking about some of the ways that we could script something out like this seems like a lot of overhead when we're restarting workers.  I put a meeting on the calendar today to discuss some options and we can document them in a google doc or lucid chart for now.

src/workers/deployment-image-update/index.js
adam2k: Do you mean to move these vars into the natsPubSub library or inside of that try block?  I thought if they were in the library we wouldn't be able to configure them from each of the workers.

No, I was testing out that queue group when I was checking out replaying the queue.  We can either completely remove that `queueGroup` or we should make them different based on each worker.

src/workers/deployment-image-update/index.js
adam2k: I'll look into fixing that up some more.  We might need to add extra try catch blocks where ever `nc` is invoked if we go that route or potentially exit the worker with an error if we can't connect to NATS?

src/workers/deployment-image-update/index.js
```diff
@@ -0,0 +1,85 @@
...
+ * NATS Deployment Update Worker
+ */
+function createNatsClient() {
+  const clientID = "deployment-image-update-worker";
+  const subject = DEPLOYMENT_IMAGE_UPDATED;
+  const nc = natsPubSub(clientID, subject, helmUpdateDeployment);
+  log.info(`NATS ${clientID} Running...`);
+
+  return nc;
+}
```
schnie, line 22:
```suggestion
const nc = natsPubSub("deployment-image-update-worker", DEPLOYMENT_IMAGE_UPDATED, helmUpdateDeployment);
log.info(`NATS ${clientID} Running...`);
```

Now that I look at this again, could we shorten it down to this?

adam2k: Yeah, that'd work.  There is error handling using `nc.on("error", ...` so we don't actually need that in a try/catch like I originally though.  I kind of like the idea of having:

```
const CLIENT_ID = "deployment-image-update-worker";
const nc = natsPubSub(CLIENT_ID, DEPLOYMENT_IMAGE_UPDATED, helmUpdateDeployment);
log.info(`NATS ${clientID} Running...`);
```

Does that seem all right, @schnie?

andriisoldatenko: >const CLIENT_ID = "deployment-image-update-worker";
let's move to constants.js or so

package.json
```diff
@@ -6,6 +6,8 @@
     "build": "prisma generate && babel src -d dist --copy-files",
     "start": "prisma generate && yarn migrate && nodemon --exec babel-node -e js,graphql src/index.js",
     "serve": "yarn migrate && node dist/index.js",
+    "workers": "node dist/workers/index.js",
```
andriisoldatenko, line 9:
```suggestion
    "worker": "node dist/workers/index.js",
```
`yarn start worker`?

src/lib/constants/index.js
```diff
@@ -56,3 +56,5 @@ export const MEDIATYPE_DOCKER_MANIFEST_V2 =
 export const RELEASE_NAME_PATTERN = /^(?![0-9]+$)(?!.*-$)(?!-)[a-zA-Z0-9-]{1,63}$/g;
 export const RELEASE_NAME_AIRFLOW_PATTERN = /^\/((?![0-9]+$)(?!.*-$)(?!-)[a-zA-Z0-9-]{1,63})\/(airflow|flower)/;
 export const VALID_DOCKER_IMAGE_NAME = /^(?<releaseName>(?![0-9]+$)(?!.*-$)(?!-)[a-zA-Z0-9-]{1,63})\/airflow$/;
+
+export const DEPLOYMENT_IMAGE_UPDATED = "houston.deployment.image.updated";
```
andriisoldatenko, line 60:
```suggestion
// Nats related constants

export const DEPLOYMENT_IMAGE_UPDATED = "houston.deployment.image.updated";
```

src/lib/nats-streaming/index.js
```diff
@@ -0,0 +1,45 @@
+import log from "logger";
+import nats from "node-nats-streaming";
+import config from "config";
+
+export function natsPubSub(clientID, subject, messageHandler) {
```
andriisoldatenko, line 5:
```suggestion
export function PubSub(clientID, subject, messageHandler) {
```

src/lib/nats-streaming/index.js
```diff
@@ -0,0 +1,45 @@
...
+    const sub = nc.subscribe(subject, opts);
+    sub.on("message", messageHandler);
+
+    log.info(`Subscribing to: ${subject}`);
+  });
+
+  return nc;
+}
+
+export function natsPublisher(clientID) {
```
andriisoldatenko, line 28:
```suggestion
export function Publisher(clientID) {
```

andriisoldatenko: since you already inside nats folder

package.json
adam2k: Right now we only have one worker, but after this PR we're going to be adding others there, so I thought it made sense to leave as `workers`.

src/workers/deployment-image-update/index.js
adam2k: the `const CLIENT_ID` is different for every worker.  We can probably move to constants, but we'll need to change the name for each new worker.

package.json
andriisoldatenko: JFYI typically when we say worker it means 1 process, so you run 1 worker (1 process) with few subscriptions o different subjects. So IMO it should be `worker`.
wdyt @schnie ?

adam2k: Ok, that makes sense because we're not running each PubSub as a separate worker.  I can switch this to `worker`.

src/lib/nats-streaming/index.js
```diff
@@ -0,0 +1,51 @@
...
+  return nc;
+}
+/**
+ * @param  {String} clientID name of the client to identify requests (unique)
+ */
+export function publisher(clientID) {
+  const natsConfig = config.get("nats");
+  const clusterID = natsConfig.clusterID;
+  const url = `${natsConfig.host}:${natsConfig.port}`;
+  const opts = {
```
adam2k, line 38: I think I'll need to figure out TLS here.

src/lib/nats-streaming/index.js
```diff
@@ -0,0 +1,126 @@
...
+ * @param  {String} clientID name of the client to identify requests (unique)
+ * @param  {String} subject subject to publish and subscribe (shared)
+ * @param  {Function} messageHandler to accept the NATS message
+ */
+export function pubSub(clientID, subject, messageHandler) {
+  const nc = publisher(clientID);
+  let sub = {};
+
+  // Subscribe after successful connection
+  nc.connectHandler = () => {
```
adam2k, line 15: @schnie and @andriisoldatenko let me know what you think about this pattern.  I couldn't figure out another way that we could test all of these events without attaching handlers to the NC object.  The tests are making sure that these handlers will run correctly and about half are only logging without logic, but those lines are killing the test codecov results and preventing the build from passing.

src/workers/deployment-image-update/index.js
```diff
@@ -0,0 +1,84 @@
...
+  }
+}
+
+/**
+ * @param  {Object} deployment from prisma
+ */
+async function commanderUpdateDeployment(deployment) {
+  log.info(`NATS Streaming Update Deployment: ${deployment}`);
+
+  /* TODO: Uncomment after testing STAN is completed and the logs for deployments match
```
adam2k, line 45: I thought this was OK to leave commented out inside of this PR because as soon as we're done with testing we can add this code in.  Let me know if you think I should delete it and only leave the `// TODO: ...` line.

src/lib/constants/index.js
```diff
@@ -56,3 +56,9 @@ export const MEDIATYPE_DOCKER_MANIFEST_V2 =
 export const RELEASE_NAME_PATTERN = /^(?![0-9]+$)(?!.*-$)(?!-)[a-zA-Z0-9-]{1,63}$/g;
 export const RELEASE_NAME_AIRFLOW_PATTERN = /^\/((?![0-9]+$)(?!.*-$)(?!-)[a-zA-Z0-9-]{1,63})\/(airflow|flower)/;
 export const VALID_DOCKER_IMAGE_NAME = /^(?<releaseName>(?![0-9]+$)(?!.*-$)(?!-)[a-zA-Z0-9-]{1,63})\/airflow$/;
+
+// NATS related constants
+
+export const DEPLOYMENT_IMAGE_UPDATE = "houston.deployment.image.update";
+export const DEPLOYMENT_IMAGE_UPDATE_ID = "deployment-image-update-worker";
```
schnie, line 63:
```suggestion
export const DEPLOYMENT_IMAGE_UPDATE = "houston.deployment.image.update";
export const DEPLOYMENT_IMAGE_UPDATE_ID = "houston-deployment-image-update-worker";
```

Another nitpick, but these are almost the same, just slightly different ha

src/workers/deployment-image-update/index.js
schnie: @adam2k we can leave this code in I think. Houston will default `COMMANDER__ENABLED` to false unless explicitly overridden, like we do now for the API layer. Instead of actually making the call to Commander, the `commander.request()` function will just skip it and log that it was skipped.

src/lib/nats-streaming/index.js
schnie: Is there any reason to attach these event handler functions onto the `nc` objects? Could we define those functions as top level `const`s in this module and pass those as the event handlers? Assigning our own functions onto the object that the client library is returning to us feels a little unsafe, since we could be clobbering something internal or something added in the future.

src/workers/deployment-image-update/index.js
adam2k: Awesome!  I can add that back in.

src/lib/nats-streaming/index.js
adam2k: yeah, that's true.  I thought the handlers would be OK in this case, but I'll try to move to their own methods and see if I can mock them out in the tests.

andriisoldatenko:
@adam2k can you please add final comment about what this PR is fixing? 
Since this PR has long conversation and some unrelated/or already resolved issues in another PRs.

adam2k:
> @adam2k can you please add final comment about what this PR is fixing?
> Since this PR has long conversation and some unrelated/or already resolved issues in another PRs.

@andriisoldatenko Good point.  I think it's best to close this PR and I'll open a new one because the comments in this PR are no longer relevant.

adam2k:
The new PR is located here: https://github.com/astronomer/houston-api/pull/428

PR #353, opened 2020-06-29
Hotfix: Metric page lag/crash

samblackk:
Patch for https://github.com/astronomer/issues/issues/1456

After investigating the metric overload in Grafana I noticed the Prometheus metrics were updated there but not in the Houston prom query list. This change includes those updates as well as increasing the range selector vector to reduce the size of the data object coming back. This should help reduce the problem with the metrics page crashing when users have a large amount of tasks running. 

**Note:** This is not a full fix - we need to work on optimizing these metrics better. But this should provide some relief for the customer while we work on a bigger solution.

See this documentation from Prometheus for info:
https://prometheus.io/docs/prometheus/latest/querying/basics/#range-vector-selectors

Ready for review.


jpweber:
Does this need to be deployed now?

andriisoldatenko:
@samblackk i think it's for 0.17
cc @vishwas-astro

jpweber:
ok. Just making sure.

samblackk:
Sorry - mislabeled. 0.17 - we need to test this. I was only able to test it by proxying from staging. 
Load up the UI and Houston - have 18-20 running tasks for a bit, see if the UI crashes.

samblackk:
We can do all the testing on `dev` if we get this pushed up.

adam2k:
@samblackk would the `taskSuccessRate` and `taskFailRate` also have the same bug?   Are those also being used for metrics?  I noticed that `duration` is used in both of those places.

samblackk:
No - those come back as a single value.

src/resolvers/subscription/metrics/queries.js
```diff
@@ -67,25 +64,25 @@ export default function queries(deployment, since, step) {
...
         query: rangeQuery(
-          `airflow_executor_queued_tasks{deployment=~"${deployment}"}`
+          `rate(airflow_executor_queued_tasks{deployment=~"${deployment}"})[5m]`
         )
       },
       {
         name: "failedTasks",
         query: rangeQuery(
-          `increase(airflow_ti_failures{deployment=~"${deployment}"}${fstep})`
+          `rate(airflow_operator_failures{deployment=~"${deployment}"}[5m])`
```
sjmiller609, line 79: are we intentionally not using "airflow_ti_failures"?

sjmiller609: also, how does this query work if the user selects "30min" in the time drop down in the metrics tab?

samblackk: I'm using the queries that are being used in Grafana so the data is the same. We can revert back to `airflow_ti_failures` if the other shouldn't be used. The time query should work the same - it's in the query URL, not the metric itself.

sjmiller609: At risk of being annoying, I don't think I understand this. How about we build an image, push it somewhere, switch out the houston image running in dev cluster and take a look?

samblackk: Sure, lets try it.

schnie: Hmm, whats the difference between these two metrics - `airflow_ti_failures` and `airflow_operator_failures`. Are they essentially the same for our purpose here?

schnie: Looks like these counters are exactly the same, except the operator one has the operator name as part of the metrics - https://github.com/apache/airflow/blob/631ac484f14e7f3f9637e1229252769d61e388e1/airflow/models/taskinstance.py#L1269-L1270.

If possible, I think `ti_failures` might slightly cleaner because we can avoid the reliance on this mapping - https://github.com/astronomer/ap-vendor/blob/master/statsd-exporter/include/mappings.yml#L12-L16.

If we changed this mapping, this query would break, where if we used `ti_failures` the behavior remains the same. Minor thing, but may be nicer in the long run.

sjmiller609: What about the time period of the query? This change appears to remove the capability of the query to select a different time range, for example in this drop down: 
image

samblackk: I updated to reinstate the dynamic `step` and revert the metric change back to `ti` instead of `operator` for failure & success.

PR #355, opened 2020-07-01
Have OpenID Connect workflow go via Houston first, not direct to IdP

kushalmalani:
fix astronomer/issues#1464

Changes:

1. Auth logins will now have url like /oauth/start?provider=${providerName} instead of the big discoveryURL
2. Adding functions to get nonce form cookie and validating nonce
3. Setup a new custom route (*/start) to start auth0 login process and set generated nonce as a cookie in response
4. Error Handling
5. Adding new test cases and missing test cases to bump up coverage.


src/routes/v1/oauth/start/start.js
```diff
@@ -0,0 +1,11 @@
+import { getClient } from "../../../../lib/oauth/config";
+
+export default async function(req, res) {
+  const providerName = req.path.split("/")[1];
+  const client = await getClient(providerName);
+  const url = await client.authUrl({});
+  const urlParams = new URLSearchParams(url);
+  const nonce = urlParams.get("nonce");
+  res.cookie("nonce", nonce, { httpOnly: true });
```
andriisoldatenko, line 9: how it works with `https`?

kushalmalani: The httpOnly flag implies that client side scripts cannot access the cookie for an overrides. Incase of an XSS (Cross Site scripting) attack, the attacker cannot access the cookie. This should work seamlessly over https.

adam2k:
@kushalmalani are there some test steps that we can follow for functional testing?

kushalmalani:
> @kushalmalani are there some test steps that we can follow for functional testing?

@adam2k The steps should be the same. The user flow remains the same. Go on the signup/login page and try to login via any of the providers.

To test the attack, Try to copy the access token from the call back API from the browser network tab and use the same token to call again and auth0 will reject the request since the nonce was already used and Houston should throw a 403 error

src/lib/oauth/config/index.js
```diff
@@ -176,6 +186,34 @@ export function getClaim(claims, mapping, name) {
   return get(claims, mapped);
 }
 
+/*
+ * Return an object of cookies.
+ * @param {String} cookie string
+ * @return {Object} list of all cookies.
+ */
+export async function getCookieList(cookies) {
```
adam2k, line 194: Does this need to return a promise?

src/lib/oauth/config/index.js
```diff
@@ -176,6 +186,34 @@ export function getClaim(claims, mapping, name) {
...
+  const cookieList = {};
+
+  cookies &&
+    cookies.split(";").forEach(function(cookie) {
+      const parts = cookie.split("=");
+      cookieList[parts.shift().trim()] = decodeURI(parts.join("="));
```
adam2k, line 200:
```Suggestion
          const key = cookie.split("=").shift().trim();
          cookieList[key] = decodeURI(cookie);
```

src/lib/oauth/config/index.js
```diff
@@ -176,6 +186,34 @@ export function getClaim(claims, mapping, name) {
...
+
+/*
+ * Return an boolean response after checking if nonce is matching
+ * @param {String} id_token which comes from auth0. Need to decode this to extract nonce
+ * @param {String} nonce string
+ * @return {Object} list of all cookies.
+ */
+export function validateNonce(id_token, nonce) {
+  const decodedToken = jwt.decode(id_token);
+  return decodedToken.nonce === nonce ? true : false;
```
adam2k, line 214: Isn't this the same as:

```Suggestion
  return decodedToken.nonce === nonce; 
```

src/lib/oauth/config/index.js
```diff
@@ -176,6 +186,34 @@ export function getClaim(claims, mapping, name) {
...
+}
+
+/*
+ * Return an boolean response after checking if nonce is matching
+ * @param {String} id_token which comes from auth0. Need to decode this to extract nonce
+ * @param {String} nonce string
+ * @return {Object} list of all cookies.
+ */
+export function validateNonce(id_token, nonce) {
+  const decodedToken = jwt.decode(id_token);
```
adam2k, line 213:
```Suggestion
export function validateNonce(idToken, nonce) {
  const decodedToken = jwt.decode(idToken);
```

src/lib/oauth/config/index.unit.test.js
```diff
@@ -110,6 +121,14 @@ describe("oauth configuration", () => {
       expect(params.get("connection")).toBeNull();
     });
 
+    test("should return starturl for the provider without provider name", async () => {
+      const client = await getClient("google");
+      const url = client.startUrl();
+      expect(url).toBe(
+        "http://houston.astronomer.io:8871/v1/oauth/google/start"
+      );
```
adam2k, line 129: This is a nit:

```Suggestion
      const expected = "http://houston.astronomer.io:8871/v1/oauth/google/start";
      expect(url).toBe(expected);
```

src/lib/oauth/config/index.unit.test.js
```diff
@@ -151,4 +170,85 @@ describe("When google client_id is null", () => {
     const state = JSON.parse(params.get("state"));
     expect(state.integration).toBe("google-oauth2");
   });
+
+  test("should return starturl for the provider", async () => {
+    const client = await getClient("google");
+    const url = client.startUrl();
+    expect(url).toBe("http://houston.astronomer.io:8871/v1/oauth/google/start");
```
adam2k, line 177: Same comment as nit above.

src/lib/oauth/config/index.unit.test.js
```diff
@@ -110,6 +121,14 @@ describe("oauth configuration", () => {
       expect(params.get("connection")).toBeNull();
     });
 
+    test("should return starturl for the provider without provider name", async () => {
+      const client = await getClient("google");
+      const url = client.startUrl();
+      expect(url).toBe(
+        "http://houston.astronomer.io:8871/v1/oauth/google/start"
```
adam2k, line 128: Does this test always pass using this URL?  Will this be `http://houston.astronomer.io` for enterprise deployments?

It might be safer to do something like the jest equivalent of contains("v1/oauth/google/start")

src/lib/oauth/config/index.unit.test.js
```diff
@@ -151,4 +170,85 @@ describe("When google client_id is null", () => {
...
+    expect(claimEmail).toBe(email);
+  });
+});
+
+describe("Validate Nonce", () => {
+  const provider = "test-provider";
+  const integration = "example";
+  const origin = "http://houston.local.astronomer.io:8871/v1/oauth/callback";
+  const nonce = "fake-nonce-value";
+  const idTokenObj = {
```
adam2k, line 225: This is a nit: Is there a better name for this?  `authToken`, `auth` or something that avoids using `Obj` in the variable name.

src/routes/v1/oauth/start/start.js
adam2k, line 9:
```Suggestion
  const options = { httpOnly: true };
  res.cookie("nonce", nonce, options);
```

src/lib/oauth/config/index.unit.test.js
kushalmalani: The url would change depending on the environment

`${houston()}/${version()}/oauth/start?provider=${provider}`;

src/lib/oauth/config/index.js
kushalmalani: Dont see a need for it. What do you think?

ashb: Looks fine to me -- just operating on a (short) string, no I/O or network requests here.

src/routes/v1/oauth/callback/post.js
```diff
@@ -23,6 +28,12 @@ export default async function(req, res) {
   const firstUser = await isFirst();
   const publicSignups = config.get("publicSignups");
 
+  const cookieList = await getCookieList(req.headers.cookie);
+  const isNonceValid = await validateNonce(req.body.id_token, cookieList.nonce);
+
+  if (!isNonceValid) {
+    return res.status(403).send("Unable to login! Validation failed.");
```
ashb, line 35:
```suggestion
    log.error("Nonce was invalid. (received %s)", cookieList.nonce);
    return res.status(403).send("Unable to login! Validation failed.");
```

src/routes/v1/oauth/start/start.js
```diff
@@ -0,0 +1,12 @@
+import { getClient } from "../../../../lib/oauth/config";
+
+export default async function(req, res) {
+  const providerName = req.query.provider;
+  const client = await getClient(providerName);
+  const url = await client.authUrl({});
+  const urlParams = new URLSearchParams(url);
+  const nonce = urlParams.get("nonce");
+  const options = { httpOnly: true };
```
ashb, line 9: We should set secureOnly as well - probably only when `node.env !="development"`

src/routes/v1/oauth/callback/post.js
ashb: This is validating the nonce before we've checked anything else about the JWT payload, which is a bit of a no-no when it comes to crypto.

We should validate the payload is correctly signed first, and only then check the nonce.

ashb: Oh, in fact, since we are passing `{nonce}` to `provider.callback` (line 53) we don't need to check the nonce ourselves at all - openid-client does it for us.

This from their docs:

> `nonce`: `<string>` When provided the authorization response's ID Token nonce parameter will be checked to be the this expected one. Use of this check is required if you sent a nonce parameter into an authorization request.

src/routes/v1/oauth/start/start.js
```diff
@@ -0,0 +1,12 @@
+import { getClient } from "../../../../lib/oauth/config";
+
+export default async function(req, res) {
+  const providerName = req.query.provider;
+  const client = await getClient(providerName);
+  const url = await client.authUrl({});
+  const urlParams = new URLSearchParams(url);
+  const nonce = urlParams.get("nonce");
```
ashb, line 8: This feels a bit clunky (extracting the nonce out of the query parameter via parsing the URL)

Given `authUrl` is our function (and not from openid-client) we could change this:

- Either pass in the nonce: `const nonce = client.generateNonce(); const url = client.authUrl(nonce, {})`
- Return a URL and the nonce as two separate fields. `const {url, nonce } = client.authUrl({})`

(We should probably also give the `state` parameter to authUrl a default of `{}` so we don't have to pass in empty dict.)

src/lib/oauth/config/index.js
adam2k: Ok, you can remove `async` from this method.

src/lib/oauth/config/index.unit.test.js
```diff
@@ -139,16 +158,73 @@ describe("When google client_id is null", () => {
...
+    };
+    const claimEmail = getClaim(claims, {}, "email");
+    expect(claimEmail).toBe(email);
+  });
+});
+
+describe("getCookieList", () => {
+  test("should return a list of cookies", async () => {
+    const cookies = "ajs_anonymous_id=fake-id; nonce=some-fake-nonce";
+    const cookieList = await getCookieList(cookies);
```
ashb, line 225: This fn isn't async -- does `await` work on sync functions?

kushalmalani: You can await any function. If function does not return a promise, the return value is the value that is resolved by the await call. I can remove it if needed. Tests should work fine regardless

ashb: If it works no problem!

src/lib/oauth/config/index.unit.test.js
adam2k: What I'm wondering is if `"http://houston.astronomer.io:8871/v1/oauth/google/start"` will fail when `client.startUrl();` is giving a different URL based on enterprise CI (somebody running CI that we don't have control over).  For this test it might be safer to do something like what you have in your comment `${houston()}/${version()}/oauth/google/start`.  All you really care about in this test is that the startURL() is returning the correct address, right?

src/routes/v1/oauth/callback/post.js
```diff
@@ -35,12 +36,16 @@ export default async function(req, res) {
...
-  });
+  let tokenSet;
+  try {
+    tokenSet = await provider.callback(null, req.body, {
+      state: rawState,
+      nonce: cookieList.nonce
+    });
+  } catch (e) {
+    log.error(e);
+    return res.status(403).send("Unable to Login!");
```
adam2k, line 47: 👍

src/routes/v1/oauth/start/index.unit.test.js
```diff
@@ -0,0 +1,28 @@
+import router from "./index";
+import { providerCfg } from "../../../../lib/oauth/config";
```
schnie, line 2:
```suggestion
import { providerCfg } from "oauth/config";
```

@kushalmalani we use a module loader plugin that lets us import under the `src/lib` directory directly (https://github.com/astronomer/houston-api/blob/main/.babelrc#L18).

src/routes/v1/oauth/start/start.js
```diff
@@ -0,0 +1,18 @@
+import { getClient } from "../../../../lib/oauth/config";
```
schnie, line 1:
```suggestion
import { getClient } from "oauth/config";
```

src/routes/v1/oauth/start/start.js
```diff
@@ -0,0 +1,18 @@
...
+  const client = await getClient(providerName);
+  const nonce = generators.nonce();
+  const url = await client.authUrl(nonce);
+  const isDev = process.env.NODE_ENV === "development";
+  let options;
+  if (isDev) {
+    options = { httpOnly: true };
+  } else {
+    options = { httpOnly: true, secure: true };
+  }
```
schnie, line 15:
```suggestion
const options = { httpOnly: true };
if (!isDev) options.secure = true;
```

Nitpicking as usual but just something that could shorten it up here a little.

src/routes/v1/oauth/start/index.unit.test.js
kushalmalani: Yeah, that's just VS helping me importing :P. I can update that.

PR #356, opened 2020-07-06
Fix issue with overriding ENV Vars and extra_au

andriisoldatenko:
fix https://github.com/astronomer/issues/issues/1490

related astronomer/issues#1418


andriisoldatenko:
pls don't merge, need practical retest!

src/resolvers/mutation/update-deployment-variables/fragments.js
```diff
@@ -2,9 +2,13 @@ export const queryFragment = `fragment EnsureFields on Deployment {
   id
   config
   releaseName
-  workspace {
+    workspace {
```
ashb, line 5:
```suggestion
  workspace {
```

Extra space?

andriisoldatenko:
@sjmiller609 good point, but not sure that we have easy way to deal with such tests.
regarding first part of change we don't have real graphql server so we can't test porperly `queryFragment`.
regarding second part: it's just typo, since we removed `args.env` in original PR.

i know only we need to retest using practial steps:

1. Created an ENV Variable on orbit UI
2. Verified in scheduler container that it exists
3. Increased extra capacity from 10 to 60 AUs and verified it increase pod limit under metrics tab
4. Verified ENV var again on UI (It was available) and scheduler ( Not available), somehow it was removed from helm and containers
5. re-created the variable and verified it exist in scheduler
6. But Now extra capacity resources are set back to 18 pods (which is at 10 AUs) in metrics while the slider still shows 60 AUs
This is impacting users as every time they perform deployment ENV variables breaks and when they fix ENV variable extra capacity resources are set back to 10 AUs which tends to fail their tasks as pods are unable to initialize due to resource crunch (edited)

src/resolvers/mutation/update-deployment/index.js
```diff
@@ -102,7 +98,7 @@ export default async function updateDeployment(parent, args, ctx, info) {
         version: updatedDeployment.version
       },
       namespace: generateNamespace(updatedDeployment.releaseName),
-      rawConfig: JSON.stringify(generateHelmValues(updatedDeployment, values))
+      rawConfig: JSON.stringify(generateHelmValues(updatedDeployment))
```
sjmiller609, line 101: I think in this case, since we are removing the environment variable input, we want to ensure there is a test for the case of update deployment being called for a new environment variable that the result of 'generateHelmValues' includes the new environment variable in the response. I'll let code owners @adam2k decide for it.

adam2k: @sjmiller609 once this PR goes in, I believe the only place that we update the env vars is: `src/resolvers/mutation/update-deployment-variables/index.js`.  Do you think that we need to return the existing environment variables?  If so, that will be a slight change to the implementation and probably out of scope for this PR. 

@andriisoldatenko can you confirm that this `update-deployment` mutation will no longer update the deployment environment variables?

adam2k, line 101: Can one or more of the tests be updated to account for the new rawConfig value?

andriisoldatenko: yep on it.

andriisoldatenko:
@adam2k i can confirm we divided it last release into 2 separate mutations / UIs.

src/resolvers/mutation/update-deployment/index.js
```diff
@@ -91,9 +90,6 @@ export default async function updateDeployment(parent, args, ctx, info) {
 
   // If we're syncing to kubernetes, fire updates to commander.
   if (args.sync) {
-    // Map to a format that the helm chart expects.
-    const values = mapCustomEnvironmentVariables(updatedDeployment);
-
     // Update the deployment, passing in our custom env vars.
```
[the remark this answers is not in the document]
andriisoldatenko: fixed.

src/resolvers/mutation/update-deployment/index.js
andriisoldatenko: done.

andriisoldatenko:
@pritt20 can you confirm if it works in dev cluster?

practically LGTM from my practical tests in dev cluster.

src/resolvers/mutation/update-deployment-variables/fragments.js
andriisoldatenko: fixed.

andriisoldatenko:
@schnie @sjmiller609 @adam2k PTAL.
cc @ashb @ryanahamilton

src/resolvers/mutation/update-deployment/index.unit.test.js
```diff
@@ -377,4 +377,170 @@ describe("updateDeployment", () => {
     );
     expect(res.data.updateDeployment.id).toBe(id);
   });
+
```
adam2k, line 380: 🎉

andriisoldatenko:
@schnie are you OK to merge this and release to prod?
cc @vishwas-astro

schnie:
@andriisoldatenko so if someone is using the API directly, and they are using this mutation to set the env vars, it'll break when we roll this out right?

I feel like some folks are doing that, but not sure if they are on Cloud or not.

andriisoldatenko:
@schnie correct, but we changed way to work with env vars, we need to migrate these clients to new mutation.
so to support both we need revert few already merged PRs.

andriisoldatenko:
we already removed it here https://github.com/astronomer/houston-api/pull/346/files#diff-3eef5d7dead56e2c4426a8864c8d3701L100-L123

image

PR #359, opened 2020-07-08
Fix invalid test when runtime is slow

adam2k:
Fixes flaky test case that fails when the runtime is slow because multiple date objects are create synchronously and then compared against each other. 

```
FAIL  src/resolvers/types/auth-user/index.unit.test.js
  ● AuthUser › token resolves a valid token
    expect(received).toBe(expected) // Object.is equality
    Expected: 1594136505
    Received: 1594136504
```


adam2k:
Oops...I didn't mean to close this PR.  I'm reopening now.

PR #361, opened 2020-07-09
Encrypting nonce

kushalmalani:
Encrypting nonce to add another layer of security to avoid replay attacks

fix astronomer/issues#1506


kushalmalani:
@ashb I was thinking of using the same cert we use to sign auth jwt. Do you think it would be a good idea to do this since this is a short term implementation or should we generate new certificates for signing nonce?

Will add more test cases to increase coverage if you the implementation is ok.

src/lib/jwt/index.js
```diff
@@ -28,17 +28,22 @@ export function createAuthJWT(userId) {
...
-    expiresIn: ms(millis),
-    notBefore: 0,
-    mutatePayload: !!opts.returnPayload,
+  const options = {
     algorithm: alg
-  });
+  };
+
+  // We can't set expire date if payload is a string
+  if (typeof payload !== "string") {
```
andriisoldatenko, line 40: is it possible to add tests?

kushalmalani: Yeah, I am just waiting on ash to give me a green signal with the approach. Will be adding tests to make sure coverage is 100% for all the changes.

ashb:
Or maybe jws

ashb:
Yes, it's jws - http://self-issued.info/docs/draft-ietf-jose-json-web-signature.html#rfc.section.3.3

(I just think we should be careful about the names)

src/routes/v1/oauth/callback/post.js
```diff
@@ -40,7 +40,7 @@ export default async function(req, res) {
   try {
     tokenSet = await provider.callback(null, req.body, {
       state: rawState,
-      nonce: cookieList.nonce
+      nonce: createJWT(cookieList.nonce)
```
ashb, line 43: Since we have signed the nonce, after we've let oidc client do it's checks we should verify it is encrypted correctly

ashb: Oh, and when doing this verification we should ensure the right encryption alg is used (null is a "valid" jws encryption, but not one we want to allow)

schnie:
Yep, agree @ashb, maybe we should separate the code into two distinct functions or something?

src/routes/v1/oauth/callback/post.js
kushalmalani: Yeah, the alg part comes from jwtSigningParam in jwt. Its set based on the environment.

kushalmalani:
Going to separate it into a different function. I think naming conventions are important to avoid confusion

src/lib/jwt/index.unit.test.js
```diff
@@ -13,6 +13,14 @@ describe("createAuthJWT", () => {
   });
 });
 
+describe("createJWS", () => {
+  test("successfully creates a jws token", () => {
+    const word = casual.word;
+    const token = createJWS(word);
+    expect(token.length).toBeGreaterThan(1);
```
ashb, line 20: How stable is the output? Could we we verify exact input and output?

kushalmalani:
@ashb I am ok to use jws package. 

The problem with storing the signed encrypted version in cookie is the replay attack is still possible. The open id specs - (https://openid.net/specs/openid-connect-core-1_0.html#NonceNotes) also recommend to store the plain text nonce in session storage and encrypted nonce as part of idtoken. I was able to verify this implementation and this avoids replay attack.

src/lib/jwt/index.unit.test.js
kushalmalani: I think i can decode the token and match the output and input.

ashb:
> @ashb I am ok to use jws package. 
> 
> The problem with storing the signed encrypted version in cookie is the replay attack is still possible. The open id specs - (https://openid.net/specs/openid-connect-core-1_0.html#NonceNotes) also recommend to store the plain text nonce in session storage and encrypted nonce as part of idtoken. I was able to verify this implementation and this avoids replay attack.

Sounds good!

PR #364, opened 2020-07-13
Fix conflicts/deployment rollouts

adam2k:
Add all the first iteration of endpoints to work with NATS.


adam2k:
There is still some work to do here, but I'm going to squash and merge so everything is up to date on our test branch.

PR #370, opened 2020-07-15
Add NATS configuration and lib files

adam2k:
This is to get the basic NATS setup in place for the subsequent PRs for each Houston API endpoint that we are targeting.  This does not actually run any worker and none of the workers are included in this PR.

To test:

1. Make sure `docker-compose up -d` will spin up a local instance of the NATS Server
2. The configuration for NATS is correct to start (includes most defaults so we are explicit on the values)
3. Using Astro UI creating new deployments works correctly with database changes.

Fixes https://github.com/astronomer/issues/issues/1542


config/default.yaml
```diff
@@ -16,6 +16,23 @@ prisma:
   secret: ~
   debug: false
 
+# NATS configuration.
+nats:
+  host: nats://localhost
+  port: 4222
+  clusterID: development-cluster
```
schnie, line 23: @adam2k should this be something we override in prod, or should we set this to a value that makes sense for both envs and not worry about overriding it?

config/default.yaml
```diff
@@ -327,6 +344,7 @@ roles:
       ? deployment.images.pull
       ? deployment.metrics.get
       ? deployment.serviceAccounts.get
+      ? deployment.rollouts.get
```
schnie, line 347: @adam2k I think we got rid of the usage here. Should we pull this out?

config/default.yaml
adam2k: @schnie I was thinking that we would override this for cloud and enterprise in prod.  Maybe that's not necessary because we'll know what env that we're debugging?  

Would there every be a case where we would aggregate logs together from Cloud and Enterprise into a data lake?  That's the only time that maybe this would be useful.

schnie: Could be at some point, nothing now. The reason I'm suggesting is that we have so many layers of configuration that it really gets hard to fully know what needs to go where. If this could be one less config that we need to have a separate PR on another repo for, then I think its worth it for now. Its configurable so we can always add something on top later when we need to. Could we do something generic for now?

schnie: `houston`?

config/default.yaml
```diff
@@ -371,9 +387,9 @@ deployments:
   # Airflow image settings
   # This is a list of supported airflow versions and corresponding docker tag
   images:
-    - version: 0.0.0
+    - version: 1.10.10
       channel: stable
-      tag: 0.0.0-alpine3.10-onbuild
+      tag: 1.10.10-alpine3.10-onbuild
```
schnie, line 392: These were intentionally fake values so that we didn't feel the need to update this list. It gets overridden in prod by something anyway. What was the reason to change here?

src/lib/nats-streaming/index.js
```diff
@@ -0,0 +1,231 @@
+import log from "logger";
+import nats from "node-nats-streaming";
+import config from "config";
+/**
```
schnie, line 4: @adam2k can we add descriptions to these function docs so its easier for folks to understand what they do?

config/default.yaml
adam2k: Will revert.  I thought that those were actually used as defaults.

src/lib/nats-streaming/index.js
adam2k: Added the descriptions.

PR #373, opened 2020-07-15
Update: Prevent duplicate service account

samblackk:
Fix for https://github.com/astronomer/issues/issues/1470

This makes the `label` field on `ServiceAccount` unique and enforces the validation through the mutations.


database/datamodel.graphql
```diff
@@ -94,7 +94,7 @@ type InviteToken {
 type ServiceAccount {
   id: ID! @id
   apiKey: String @unique
-  label: String
+  label: String @unique
```
andriisoldatenko, line 97: @samblackk i was wondering if it crash when we have duplicates in table.?
cc @schnie wdyt?

schnie: Yea I would worry about that too.

samblackk: How should we roll this out then? Not use `@unique` and rely on the manual validation?

src/resolvers/mutation/update-service-account/index.js
```diff
@@ -16,6 +17,9 @@ export default async function updateServiceAccount(parent, args, ctx, info) {
   const where = { id: args.serviceAccountUuid };
   const role = args.payload.role;
 
+  // Validate if service account label exists or not
+  validateExistence(data.label);
```
vishwas-astro, line 21: @samblackk I am sure I am missing something here, but dont you want to call validateExistence (which throws an error if service account already exists) for only the create service account use cases and not the update ones?

samblackk: I'm working on this now - scoping service accounts to workspaces and deployments. So previous labels won't be accounted for - only newly created or updated ones will be. And we would want to call it for updating labels as well because the user has the ability to update the label - if we want to maintain the "unique" state it needs to be validated at both creation and update.

samblackk:
This has been updated to support scoped validation of service account labels. However, every test is failing with ```[[GraphQLError: Only absolute URLs are supported]]`` - which isn't related to the script I wrote. @kushalmalani is this a side effect of your auth changes? Do I need to fix something in these commits?

kushalmalani:
> This has been updated to support scoped validation of service account labels. However, every test is failing with `[[GraphQLError: Only absolute URLs are supported]]`- which isn't related to the script I wrote. @kushalmalani is this a side effect of your auth changes? Do I need to fix something in these commits?

@samblackk I dont this has anything to do with the auth changes. Looking at the code, i feel the following is the root cause of the test failures:

1. You would need to mock "validateExistence" in the tests. If you dont do that, it would make a call to that method and the error should come from this line - https://github.com/astronomer/houston-api/pull/373/files#diff-2f3f153ac87f885d136889d7952b23baR20

Prisma needs an endpoint to connect to and in the test cases it would be undefined since that value comes from an environment variable "PRISMA__ENDPOINT" in database/prisma.yml and hence the error. Mocking the function should fix your test cases.

samblackk:
@andriisoldatenko @schnie ready for 2nd review.

andriisoldatenko:
LGTM if works!

src/lib/service-accounts/existence/index.js
```diff
@@ -0,0 +1,68 @@
...
+import { find } from "lodash";
+
+/*
+ * Check if this label exists for this service account.
+ * Houston 1 relied on a multi-column unique index, which prisma does not currently support.
+ * This is the workaround for now. Issues for this are opened here:
+ * https://github.com/prisma/prisma/issues/171
+ * https://github.com/prisma/prisma/issues/1300
+ */
+export default async function validateExistence({
```
schnie, line 12: @samblackk what do you think about splitting this into 2 explicit functions - one for workspace and one for deployments? Historically, Houston 1 had a lot of logic that handled both things, since they can be similar, but I feel like decoupling things and having distinct, explicit pieces of code makes it easier to reason about and leaves room for things to evolve independently.

Not a huge deal in this particular case, but got me thinking. What do you think?

samblackk: I don't think it's necessary at this time. In the future we do split the funcality to do things differently, it's easy enough to decouple further.

PR #374, opened 2020-07-16
Always upgrade Airflow chart to the configured version

sjmiller609:
There is a feature called houston upgrader that is run in workers outside of the normal houston service using the 'upgrade-deployments' npm script. This is necessary to enable when we intended to immediately converge all Airflow charts to the configured version. However, for some changes to Airflow chart, it is preferable to update airflow user's charts the next time the user makes a change.

This patch changes the behavior of the updateDeployments mutation to perform the Airflow Helm chart version upgrade when the Houston config is updated.

I will also backport this to 0.16 and PR separately.

closes: https://github.com/astronomer/issues/issues/1555


src/resolvers/mutation/update-deployment/index.js
```diff
@@ -74,11 +74,13 @@ export default async function updateDeployment(parent, args, ctx, info) {
 
   // Validate our args.
   await validate(deployment.workspace.id, mungedArgs, deployment);
-
+  // Always update charts to the version configured
+  const airflowChartVersion = config.get("deployments.chart.version");
```
adam2k, line 78:
```Suggestion
  const version = config.get("deployments.chart.version");
```

src/resolvers/mutation/update-deployment/index.js
```diff
@@ -74,11 +74,13 @@ export default async function updateDeployment(parent, args, ctx, info) {
...
   // Create the update statement.
   const where = { id: args.deploymentUuid };
   const data = merge({}, updatablePayload, {
     config: mungedArgs.config,
+    version: airflowChartVersion,
```
adam2k, line 83:
```Suggestion
    version,
```

src/resolvers/mutation/update-deployment/index.js
andriisoldatenko: do you think it's better for readability reduce less characters?

schnie:
Should we also apply this change when they do a code push? This lives in the /v1/registry/events handler

schnie:
Oh wait, I just remembered, we added _another_ point where we apply `helm upgrade`. We recently added the `updateDeploymentVariables` mutation for the Astro UI env vars. This is here https://github.com/astronomer/houston-api/blob/main/src/resolvers/mutation/update-deployment-variables/index.js.

andriisoldatenko:
@adam2k can you pls help here to improve test coverage?

sjmiller609:
I'll make sure the unit test coverage is expanded. Happy to receive input!

Thank you greg, I will address this location also.

vishwas-astro:
@sjmiller609 dont we want to backport this to 0.16 release branch?

PR #379, opened 2020-07-21
Implement NATS for Update Deployment

adam2k:
Migrate update deployment to NATS.

Testing:

```
$ docker-compose up -d # Make sure you have NATS Server running
$ yarn
$ yarn build
$ yarn worker
```

Make sure that you see the update deployment worker starting up correctly and subscribing to message coming through.

Relates to issue:

astronomer/issues#1532


andriisoldatenko:
@adam2k will decrease coverage by 0.16%.
is it possible to fix?

adam2k:
> @adam2k will decrease coverage by 0.16%.
> is it possible to fix?

I'll try to rebase with `main` once the other PRs go in and if that does work I can try to add another test in here.

src/schema.graphql
```diff
@@ -477,7 +477,7 @@ type Mutation {
     deploymentUuid: Uuid!
     payload: JSON
     config: JSON
-    sync: Boolean
+    sync: Boolean # Deprecated
```
adam2k, line 480: @schnie is this the correct way to mark this? I tried `@deprecated`, but because this isn't a type I received an error `Directive "deprecated" may not be used on ARGUMENT_DEFINITION. GraphQL: Validation.`

schnie: Should be `@deprecated`. Think we have some examples in here

adam2k: hmm...I tried that, but I was getting an error in VS Code.  Let me try it again.

adam2k: I talked to @ryanahamilton about this.  We'll need to update the API and Astro-UI at the same time for this.  There's no way to mark the argument as `@deprecated` that I could find besides the comment.

ryanahamilton: astronomer/astro-ui#708

schnie: Does the CLI references this as well?

adam2k: Yes, there are 3 tickets for this.  We can release in 0.18:

https://github.com/astronomer/issues/issues/1577
https://github.com/astronomer/issues/issues/1578
https://github.com/astronomer/issues/issues/1579

PR #380, opened 2020-07-21
Implement NATS worker for deployment image update

adam2k:
Migrate update image deployment to NATS.

Testing:

```
$ docker-compose up -d # Make sure you have NATS Server running
$ yarn
$ yarn build
$ yarn worker
```

Make sure that you see the update image deployment worker starting up correctly and subscribing to message coming through.

Relates to issue:

astronomer/issues#1532


src/workers/deployment-image-update/index.js
```diff
@@ -0,0 +1,88 @@
...
+ * @param  {Object} message NATS message
+ */
+export async function helmUpdateDeployment(message) {
+  try {
+    const id = message.getData();
+    const deployment = await getDeploymentById(id);
+    const { releaseName } = deployment;
+
+    await commanderUpdateDeployment(deployment);
+    publishUpdateDeployed(id);
```
schnie, line 36: This function is a one-liner. Could we save all the boilerplate and pop that line up here?

adam2k: Yeah, I was originally thinking that could be some kind of lib function, but it's not necessary.  I can update this.

PR #381, opened 2020-07-21
Breaking down config json

kushalmalani:
Fixes astronomer/issues#1419

depend on:
https://github.com/astronomer/astro-ui/pull/704
https://github.com/astronomer/astro-cli/pull/349


src/lib/deployments/config/index.js
```diff
@@ -559,6 +559,23 @@ export function generateDefaultDeploymentConfig() {
...
+/*
+ * Generate the `config` object for a deployment
+ * @return {Object} The deployment config.
+ */
+export function generateDeploymentConfig(args) {
+  const deploymentConfig = {};
+  const config = ["executor", "workers", "webserver", "scheduler"];
+
+  Object.keys(args).map(item => {
+    if (config.includes(item)) {
```
adam2k, line 571: Are there any other additional checks to consider here like `executor: {}` or is that OK to add for `deploymentConfig`?

src/lib/deployments/config/index.unit.test.js
```diff
@@ -348,6 +352,35 @@ describe("generateNextTag", () => {
...
 });
 
+describe("generateDefaultDeploymentConfig", () => {
+  test("should return deployment config object back", () => {
+    const res = generateDefaultDeploymentConfig();
+    expect(res.executor).toBe("CeleryExecutor");
+  });
+});
+
+describe("generateDeploymentConfig", () => {
```
adam2k, line 362: Can there be more tests for this to cover different variations of `generateDeploymentConfig`?  And in this test making sure that `"workers", "webserver", "scheduler"` aren't a part of the response?

src/schema.graphql
```diff
@@ -225,10 +230,37 @@ type EnvironmentVariable {
...
+
 # Deployment properties are only to support the legacy API.
 # Properties have been moved to top-level fields on Deployment internally.
 type Deployment {
   id: ID!
+  executor: ExecutorType
+  workers: WorkersType
+  webserver: WebserverType
+  scheduler: SchedulerType
   config: JSON
```
ryanahamilton, line 264: Should we go ahead and flag as deprecated?
```suggestion
  config: JSON @deprecated
```

src/lib/deployments/config/index.js
kushalmalani: No, we dont need additonal checks since the incoming schema would validate that, and i have just conforming the deployment config from root level to inside an object so as to not change the data layer. This would be handled in the next rollout.

src/schema.graphql
kushalmalani: @schnie @andriisoldatenko Thoughts on deprecating config in schema?

schnie: Sounds good I think

src/schema.graphql
```diff
@@ -468,6 +500,10 @@ type Mutation {
     description: String
     version: String
     airflowVersion: String
+    executor: ExecutorType
+    workers: JSON
+    webserver: JSON
+    scheduler: JSON
```
schnie, line 506: Can we also nest the types for this somehow? How do these nested types look right now? What's being sent up?

src/lib/deployments/config/index.js
```diff
@@ -559,6 +560,30 @@ export function generateDefaultDeploymentConfig() {
...
+export function generateDeploymentConfig(args) {
+  // Generate defaults.
+  const defaults = defaultResources("default", false);
+
+  const deploymentConfig = {};
+  const config = ["executor", "workers", "webserver", "scheduler"];
+
+  Object.keys(args).map(item => {
+    if (config.includes(item)) {
+      if (_.isEmpty(args[item])) {
```
adam2k, line 576: can you import this as `isEmpty` intead of `_`?

src/lib/deployments/config/index.js
```diff
@@ -23,7 +23,8 @@ import {
   merge,
   set,
   split,
-  uniq
+  uniq,
+  _
```
adam2k, line 27: Can you switch to `isEmpty`.

src/schema.graphql
kushalmalani: We cant nest types here. We need to create inputs and add that here to conform to that structure. I have added it. Testing different combinations of input via GraphQL playground

src/lib/deployments/config/index.js
```diff
@@ -551,12 +552,83 @@ export function generateDefaultTag() {
...
-  return { executor: AIRFLOW_EXECUTOR_DEFAULT };
+export function generateCreateDeploymentConfig(args) {
+  // Generate defaults.
+  const defaults = defaultResources("default", false);
+
+  const deploymentConfig = {};
+  const config = ["executor", "workers", "webserver", "scheduler"];
+
+  config.map(item => {
+    if (item === "executor") {
```
adam2k, line 566: There is a lot of nested logic here.  Anyway to separate everything within the main `if/else` statements into a few different functions? 

Example:

Instead of:
```
    if (item === "executor") {
      if (item in args) {
        deploymentConfig[item] = args[item];
      } else {
        deploymentConfig[item] = AIRFLOW_EXECUTOR_DEFAULT;
      }
```

```
    if (item === "executor") {
      deploymentConfig[item] = getExecutorArgsItem(item, args);
    }
    ....
function getExecutorArgsItem(item, args) { // Maybe you can think of a better name than this.
  if (item in args) {
    return args[item];
  } 
  return  AIRFLOW_EXECUTOR_DEFAULT;
}
```

src/lib/deployments/config/index.js
```diff
@@ -551,12 +552,83 @@ export function generateDefaultTag() {
...
+      if (item in args) {
+        deploymentConfig[item] = args[item];
+      } else {
+        deploymentConfig[item] = AIRFLOW_EXECUTOR_DEFAULT;
+      }
+    } else {
+      if (item in args && !isEmpty(args[item])) {
```
adam2k, line 573: Similar to comment above, but you'll probably need to do this in a few more functions.  I think it'll make it easier to understand what this code is doing by naming each block of code in a function.

sjmiller609:
this change requires a new minor version, so needs to be 0.18 or later

src/lib/deployments/config/index.js
```diff
@@ -551,12 +551,42 @@ export function generateDefaultTag() {
...
+  } else {
+    set(deploymentConfig, "executor", AIRFLOW_EXECUTOR_DEFAULT);
+  }
+
+  // Filter out uneeded keys
+  delete deploymentConfig["workspaceUuid"];
+  delete deploymentConfig["label"];
+
+  return deploymentConfig;
+}
```
schnie, line 573: What do you think about shortening this to something like `createDeploymentConfig`. I know what it means but `generate` and `create` next to each other makes it a little more confusing I think.

Also, do we only care about `args.executor` in this function? Does it make sense to make that a top-level arg instead of the actual resolver `args`?

Whenever it makes sense, I like to decouple this library code from the actual resolvers and schema. This should make swapping or upgrading this easier and less impactful - like prisma2.

Would that allow us to drop the `delete` statements here? Whenever possible, I think we should gravitate towards functional programming within Houston. I know it doesn't matter as much here, but this helps keep Houston stateless and easier to reason about.

kushalmalani: @schnie Sure, changing the name shouldnt be a problem. We need the executor logic, since defaults do not have that.

I wanted to keep all of them together, since this function objective is to return the entire deployment config. Handling the executor and delete statements part in the resolver would look messy. 

I can move them around, but w.r.t to code readability and flow, keeping them together gives a clear picture of what is happening. Thoughts?

andriisoldatenko:
@kushalmalani i can confirm it works locally for me.

src/lib/deployments/config/index.js
schnie: I think the part that feels off to me is that we are intentionally merging values into the final structure, only to `delete` them out. I'm just wondering if we could maybe switch the function signature to `export function generateCreateDeploymentConfig(executor)` and merging that single value into the final object, rather than merging _all_ the resolvers args, then removing the ones that don't fit.

Does that make sense or am I missing something?

kushalmalani: Yeah, the delete part makes sense to me. its args.executor part is what i was more concerned about. The change is not that major, it was more for understanding the flow. Do you want to move the below to the resolver as well?

  if (args.executor) {
    set(deploymentConfig, "executor", args["executor"]);
  } else {
    set(deploymentConfig, "executor", AIRFLOW_EXECUTOR_DEFAULT);
  }

schnie: I think having it in the function is fine. Maybe simplify to a one-liner?
```
set(deploymentConfig, "executor", executor ? executor : AIRFLOW_EXECUTOR_DEFAULT);
```

src/schema.graphql
```diff
@@ -225,11 +259,44 @@ type EnvironmentVariable {
...
+  resources: ResourcesType
+}
+
+type SchedulerType {
+  resources: ResourcesType
+}
+
+type WorkersType {
+  replicas: Int,
+  terminationGracePeriodSeconds: Int,
```
schnie, line 287: @kushalmalani what do you think about removing the `Type` suffix on these?

src/resolvers/mutation/create-deployment/index.js
```diff
@@ -70,6 +70,25 @@ export default async function createDeployment(parent, args, ctx, info) {
     throw new WorkspaceSuspendedError();
   }
 
+  args.config = await createDeploymentConfig(args);
```
schnie, line 73: @kushalmalani what do you think about assigning this to a new `const`? Mostly following a pattern we have in Houston to keep things immutable and easier to reason about.

src/resolvers/mutation/create-deployment/index.js
```diff
@@ -70,6 +70,25 @@ export default async function createDeployment(parent, args, ctx, info) {
...
+    "statsd",
+    "pgbouncer",
+    "flower",
+    "redis",
+    "workers",
+    "executor",
+    "releaseName",
+    "type",
+    "description",
+    "properties"
```
schnie, line 88: @kushalmalani I don't think these would need to go into the `config` column. These exist as as top-level fields on the `Deployment`. The only things that should be going into this `config` column are values that would directly be applied into the Airflow Chart as Helm values. You can see the exact spec for this here - https://github.com/astronomer/airflow-chart/blob/master/values.yaml.

You should see that everything in the `config` matches up with the structure in this file.

src/resolvers/mutation/create-deployment/index.js
schnie: Also, instead of passing `args` directly here to this function, what do you think about doing the key filtering first, then passing the filtered object into this function, with a new name. I know I'm nitpicking here but `args` kind of implicitly hints at the fact that these are direct resovler args. This would no longer directly make sense, and can help to decouple our library code from our API resolver code.

src/resolvers/mutation/create-deployment/index.js
```diff
@@ -70,6 +70,25 @@ export default async function createDeployment(parent, args, ctx, info) {
...
+  ];
+  args.config = pick(args.config, keys);
```
schnie, line 90: Maybe a new `const` here as well? I feel like `args` should be left immutable throughout the running of a resolver.

src/schema.graphql
kushalmalani: I needed to add that suffix to differentiate them from inputs. If we want a strict validation, we need to define the inputs, since types cannot be used conform them for structure validation

schnie: Got it - what about maybe `ResourceLimits`, `ResourceRequests`, `ContainerResources`, `WebserverConfig`, `SchedulerConfig`, `WorkersConfig` or something?

src/resolvers/mutation/create-deployment/index.js
kushalmalani: I need this, since deployment validation would fail otherwise. I can assign a new variable here and then initialize that to args.config, but that's just redundant

src/resolvers/mutation/create-deployment/index.js
```diff
@@ -70,6 +70,22 @@ export default async function createDeployment(parent, args, ctx, info) {
...
+    "statsd",
+    "pgbouncer",
+    "flower",
+    "redis",
+    "workers",
+    "executor"
+  ];
+
+  const deploymentArgs = pick(args, keys);
+  const deploymentConfigData = await createDeploymentConfig(deploymentArgs);
```
schnie, line 86:
```suggestion
  const deploymentConfigData = createDeploymentConfig(deploymentArgs);
```

We don't need await here do we?

src/resolvers/mutation/create-deployment/index.js
```diff
@@ -70,6 +70,22 @@ export default async function createDeployment(parent, args, ctx, info) {
...
+  const keys = [
+    "scheduler",
+    "webserver",
+    "statsd",
+    "pgbouncer",
+    "flower",
+    "redis",
+    "workers",
+    "executor"
+  ];
```
schnie, line 83: @kushalmalani could we also remove the ones that a user would never be able to tweak from the UI? statsd, pgbouncer, flower, redis shouldn't be exposed via the API.

src/resolvers/mutation/create-deployment/index.js
kushalmalani: Yeah, we dont. Forgot to clean that up.

src/resolvers/mutation/create-deployment/index.js
```diff
@@ -70,6 +70,26 @@ export default async function createDeployment(parent, args, ctx, info) {
...
+  const deploymentConfigData = createDeploymentConfig(deploymentArgs);
+
+  // Filter out needed keys
+  const keys = [
+    "scheduler",
+    "webserver",
+    "statsd",
+    "pgbouncer",
+    "flower",
+    "redis",
```
schnie, line 86: @kushalmalani can we omit these? They can't be passed in via the API so they shouldn't ever need to be plucked from the config here.

src/lib/deployments/config/index.js
```diff
@@ -182,7 +182,7 @@ export function constraints(deployment) {
   const elasticsearchEnabled = config.get("elasticsearch.enabled");
 
   // Get the executor on this deployment.
-  const executor = get(deployment, "config.executor", AIRFLOW_EXECUTOR_DEFAULT);
```
schnie, line 185: @kushalmalani is this right? This is reading the value from database record, which would still be nested under `config` right?

kushalmalani: Good catch. Yeah, I can update that

src/resolvers/mutation/create-deployment/index.js
```diff
@@ -70,6 +70,12 @@ export default async function createDeployment(parent, args, ctx, info) {
     throw new WorkspaceSuspendedError();
   }
 
+  // Filter out needed keys
+  const argKeys = ["scheduler", "webserver", "workers", "executor"];
```
schnie, line 74: @kushalmalani should we centralize this array somewhere? Constants? Its used in two places in this PR now.

src/resolvers/mutation/update-deployment/index.js
```diff
@@ -43,6 +44,7 @@ export default async function updateDeployment(parent, args, ctx, info) {
     "version"
   ]);
 
+  args.config = updateDeploymentConfig(args);
```
schnie, line 47:
```suggestion
  args.config = updateDeploymentConfig(args);

```

PR #386, opened 2020-07-22
add nats cluster

andriisoldatenko:
The purpose of this PR is just to demonstrate how to run nats cluster with nats-streaming cluster locally


docker-compose-nats-cluster.yml
```diff
@@ -0,0 +1,50 @@
...
+    image: postgres
+    restart: always
+    environment:
+      POSTGRES_USER: postgres
+      POSTGRES_PASSWORD: postgres
+    ports:
+    - 5432:5432
+    volumes:
+      - postgres:/var/lib/postgresql/data
+  nats:
```
adam2k, line 29: Could we do one of the two options:

1. Add a comment above `nats` or at the top of this file that explains when to use this over the basic `docker-compose`?
2. If it doesn't impact our systems much, I'd recommend that we convert the current `docker-compose.yml` to this version so we will have one less local dev env option.

docker-compose-nats-cluster.yml
```diff
@@ -0,0 +1,50 @@
+version: "3"
```
adam2k, line 1: This is what I mean for suggestion #1 in the comment below.

```Suggestion
# This file is intended to use for debugging NATS Streaming (STAN) issues.  Use docker-compose.yml for local development.
version: "3"
```

andriisoldatenko: @adam2k pls check my version.

schnie:
I guess this drops code coverage? Looks like that's why its failing. Can we ignore this file or something?

andriisoldatenko:
@schnie PTAL. fixed codecoverage.

PR #389, opened 2020-07-27
Remove sync from update deployment schema

adam2k:
Remove sync from update deployment schema and update tests.

Fixes https://github.com/astronomer/issues/issues/1578

Additional Notes:

I removed a couple of tests that we no longer need because the initial test hits all of the same cases.


schnie:
LGTM

PR #390, opened 2020-07-28
Add fine-grained role permissions for deployment variables

ryanahamilton:
Resolves astronomer/issues#1595.

Adds new, specific permissions to for deployment variables queries and mutations.

Removes the `workspaceUuid` query arg no longer needed now that we're utilizing deployment level permissions.


src/schema.graphql
```diff
@@ -614,7 +613,7 @@ type Mutation {
     deploymentUuid: Uuid!
     releaseName: String!
     environmentVariables: [InputEnvironmentVariable!]!
-  ): [EnvironmentVariable] @auth(permissions: ["deployment.config.update", "system.deployments.update"], op: OR)
+  ): [EnvironmentVariable] @auth(permissions: ["deployment.variables.update", "system.deployments.update"], op: OR)
```
schnie, line 616: @ryanahamilton what do you think about the `system`-level perms here? Do we keep it generic at the system level, or make some new `system.deployments.updateVariables` permission? But that also feels inconsistent to some degree. maybe `system.deployments.variables.update`?

This is a common enough thing that I feel like we need to standardize on something. I'm not sure if 4 dots would break anything or not. I kind of feel like it would work.

ryanahamilton: I had wanted to ask you about this… We have `system.serviceAccounts.update`, but `system.variables.update` certainly didn't seem to fit that pattern given variables being a property of deployments. I think I like the `system.deployments.variables.update` pattern that you suggest. 

Another question… as a property of deployments, would a `system.deployments.update` still supersede/satisfy the minimum `deployment.variables.update` requirement? i.e. should we have both system permissions like below? (or am I overthinking??)

```graphql
@auth(permissions: ["deployment.variables.update", "system.deployments.update", "system.deployments.variables.update"], op: OR)
```

schnie: I think `@auth(permissions: ["deployment.variables.update", "system.deployments.variables.update"], op: OR)` seems pretty clean. If we do this we should probably sync up all the other similar ones to this.

schnie:
@ryanahamilton hows this work in practice? @samblackk @kushalmalani @adam2k @andriisoldatenko any thoughts on this usage of 4 sections for some system-level permissions that essentially mirror a typical deployment or workspace-level permission? 

Kind of hazy what the system should be, so it'd be nice to come to some sort of standard for these.

ryanahamilton:
The conventions don't really impact consumption by the client. We only utilize the [related capabilities keys](https://github.com/astronomer/houston-api/blob/variables_permissions/src/resolvers/types/deployment/index.js#L138).

schnie:
> The conventions don't really impact consumption by the client. We only utilize the [related capabilities keys](https://github.com/astronomer/houston-api/blob/variables_permissions/src/resolvers/types/deployment/index.js#L138).

I'm more worried bout the 4 section, system level permission. When a system admin makes this request, that permission will be checked live, within the context of the request. I want to make sure that that flow is exercised and we know it works. Did you try as a system admin who does not have a direct `RoleBinding` on the deployment?

samblackk:
IMO the 4 section sys-level roles seems like the cleanest way to handle it - both contextually and systematically.

ryanahamilton:
@schnie I have tested/confirmed in Playground and Astro UI as system admin who does not have a direct RoleBinding on the deployment

PR #391, opened 2020-07-28
Add NATS scaffolding and NATS delete deployment

adam2k:
Implement Delete Deployment for NATS.  This PR also includes the scaffolding for NATS Publishers.

Fixes https://github.com/astronomer/issues/issues/1534


src/publishers/index.js
```diff
@@ -0,0 +1,50 @@
+import { publisher } from "nats-streaming";
+import { findIndex } from "lodash";
+import { DELETE_DEPLOYMENT } from "constants";
+
+class Publishers {
```
adam2k, line 5: I know we typically wanted to stay away from Classes in houston-api, but I couldn't come up with a better way to create a singleton to manage all of the publishers.

schnie: Might be a good exception here

samblackk: I agree - the use of a new Class here is a the cleanest option. Good work!

PR #392, opened 2020-07-28
Implement NATS for create deployment

adam2k:
Implement NATS Publisher for create deployment.  Once NATS is confirmed to receive published messages we can remove the duplicate commander call in the create deployment route.

Fixes https://github.com/astronomer/issues/issues/1533

**Screenshots:**
Screen Shot 2020-08-04 at 9 14 49 AM

Worker:
Screen Shot 2020-08-04 at 9 15 04 AM

Houston API
Screen Shot 2020-08-04 at 9 16 01 AM


andriisoldatenko:
@adam2k did you have a chance to test with real NATS server and commander?

andriisoldatenko:
i think you have to rebase from main branch.

adam2k:
> i think you have to rebase from main branch.

Why do I need to rebase?  There's only one commit and it's branching from main.  I created this branch yesterday.

adam2k:
> @adam2k did you have a chance to test it with real NATS server and commander?

So far I did local testing with Houston-API, the workers and Astro-UI.  For the real NATS Server do you mean to test it using kind or do you mean testing in Dev or Staging?

adam2k:
Please hold on merging this until I can resolve the general NATS connection issues that we're seeing in Dev for Houston-API -> NATS Streaming.

src/resolvers/mutation/create-deployment/index.js
```diff
@@ -174,6 +179,9 @@ export default async function createDeployment(parent, args, ctx, info) {
   //   `{ id }`
   // );
 
+  nc.publish(DEPLOYMENT_CREATED, id);
+
+  // TODO: Remove everything after this point once the the worker is confirmed to receive messages
```
adam2k, line 184: @schnie @andriisoldatenko the block of code below this TODO is creating a database connection and passing that info along via commander.  This is a little harder to decouple this from the worker than `delete deployment`.  I believe we have two options for this iteration:

1. Modify the worker to only log the create deployment ID (everything in this mutation stays as-is).
2. Remove the synchronous code from this endpoint and do everything below this TODO in the worker

To keep it consistent with delete deployment.  I'm voting for option #1.  It might be overly cautious, but that'll prevent us from scrambling to patch last minute bugs in prod this week.

adam2k: If we decide to go with option 2 then I think we should also do the same thing in delete deployment (remove the commander call and let that happen inside of the worker).

adam2k: Decided to remove updateDeployment and leave everything else in place.  We can re-add those portions when we've confirmed everything working correctly.

src/workers/deployment-created/index.js
```diff
@@ -99,14 +93,7 @@ export async function deploymentCreated(msg) {
...
     });
 
-    // Update the status of the rollout, and some deployment details.
-    await prisma.updateDeployment({
-      where: { id },
-      data: {
-        registryPassword: hashedRegistryPassword,
-        elasticsearchPassword: hashedElasticsearchPassword
-      }
-    });
```
adam2k, line 109: @schnie @andriisoldatenko I'm removing `updateDeployment` for now because commander won't actually be deploying any changes.  I've made a follow up ticket to add this code back in once we've confirmed these changes:  https://github.com/astronomer/issues/issues/1619

PR #394, opened 2020-07-29
Non-root user docker image.

andriisoldatenko:
fix https://github.com/astronomer/issues/issues/1290


schnie:
@andriisoldatenko since we're already creating an `astro` user in the base image, can we have all our derived images just switch to that user?

Maybe we even put the `USER astro` command in the base image? Would that cover all the platform images?

andriisoldatenko:
@schnie hm... to install some deps like apk add we need to switch to USER root do some installations and than switch back to User astro.

andriisoldatenko:
```
docker run -it astronomerinc/ap-base:3.12-1 sh
/ # id
uid=0(root) gid=0(root) groups=0(root),1(bin),2(daemon),3(sys),4(adm),6(disk),10(wheel),11(floppy),20(dialout),26(tape),27(video)
/ # whoami
root
/ #
```

andriisoldatenko:
@schnie PTAL

andriisoldatenko:
@adam2k looks like we need to ignore DOckerfile from codecoverage report?

andriisoldatenko:
@schnie not sure... i see codecov error for `src/lib/deployments/logs/index.js` file.

andriisoldatenko:
@schnie PTAL!

andriisoldatenko:
pls don't merge found small bug in k8s cluster.

Dockerfile
```diff
@@ -13,14 +13,23 @@
...
 
 ARG BUILD_NUMBER=-1
 LABEL io.astronomer.docker.build.number=$BUILD_NUMBER
 LABEL io.astronomer.docker.module="astronomer"
 LABEL io.astronomer.docker.component="houston-api"
 
+# install sudo as root
+RUN apk add --update sudo
+
+# create houston user and group
```
schnie, line 27:
```suggestion
# Create houston user and group
```

Dockerfile
```diff
@@ -13,14 +13,23 @@
...
-FROM astronomerinc/ap-base:3.12
+FROM astronomerinc/ap-base:3.12-1
 LABEL maintainer="Astronomer <humans@astronomer.io>"
 
 ARG BUILD_NUMBER=-1
 LABEL io.astronomer.docker.build.number=$BUILD_NUMBER
 LABEL io.astronomer.docker.module="astronomer"
 LABEL io.astronomer.docker.component="houston-api"
 
+# install sudo as root
```
schnie, line 24:
```suggestion
# Install sudo as root
```

schnie:
LGTM. Why did we switch the random function out? Just curious

andriisoldatenko:
>LGTM. Why did we switch the random function out? Just curious
@schnie random function makes unit test flaky

src/lib/deployments/logs/index.js
```diff
@@ -28,7 +28,7 @@ export const generateLogMessage = (function* generateLogs() {
 export function generateMockLogRecords(release, component) {
   return {
     hits: {
-      hits: range(random(2)).map(() => ({
+      hits: range(5).map(() => ({
```
andriisoldatenko, line 31: @schnie correct usage can be:
```
hits: range(1, 2).map(() => ({
```

andriisoldatenko: otherwise reandom(2) can return -> 0 so we can see code coverage decreasing.

andriisoldatenko: i just prefer to avoid random functions in tests.

schnie: ah ok

PR #399, opened 2020-07-31
Upgrade nodemon

adam2k:
Upgrading nodemon to remove vulnerable dependency `dot-prop@v4.2.0`.

Note: The older version of `dot-prop@v4.2.0` is still a dependency of prisma


andriisoldatenko:
@adam2k this change doesn't solve security issue. So we can ignore it as in my https://github.com/astronomer/houston-api/pull/394/files PR. Until we migrate to prisma2.

PR #403, opened 2020-08-06
Catch publish errors when connection is severed

adam2k:
This is a quick fix to catch errors that will bubble up to the end user while we are testing NATS.  When the connection is no longer in a "reconnecting" state and "closes" the `nc.publish` calls throw errors.  The only time `nc.publish` throws is when the connection is closed.


schnie:
@adam2k are we planning to keep this around long-term or trim this down after testing? Maybe some of these logs should go to `DEBUG` level?

adam2k:
> @adam2k are we planning to keep this around long-term or trim this down after testing? Maybe some of these logs should go to `DEBUG` level?

Yes, this is short term.  I'll swap to debug.  

Should this even go into main or should these PR changes only go to `release-0.18`?

schnie:
Main is fine, just wondering.

PR #404, opened 2020-08-06
Fix: Prevent duplicate service account

samblackk:
Adds additional check for roleBinding object in serviceAccount result to avoid "Cannot read property 'workspace' of null".
Also fixes another issue found with updating service accounts.

This needs to be backported to `0.16` as well as into `0.18`

Plz merge asap


schnie:
Is there any quick tests we can add to fix the codecov?

adam2k:
> Is there any quick tests we can add to fix the codecov?

Added a couple tests.

PR #415, opened 2020-08-14
Replacing mapCustomEnvironmentVariables with secretName

kushalmalani:
## Description

The PR address to send secretName instead of the map of custom environment variables. This optimizes the way we mount secrets to our pods.

## 🎟 Issue(s)

Resolves astronomer/issues#1409
Related [astronomer/airflow-charts#113](https://github.com/astronomer/airflow-chart/pull/113)

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist

- [x] Updated any related [documentation](https://github.com/astronomer/astro-docs/)


kushalmalani:
Adding test cases to bump up code coverage.

src/resolvers/mutation/update-deployment-variables/index.js
```diff
@@ -81,8 +80,10 @@ export default async function updateDeploymentVariables(parent, args, ctx) {
     queryFragment
   );
 
-  // Map the user input env vars to a format that the helm chart expects.
-  const values = mapCustomEnvironmentVariables(deployment, updatedVariables);
+  // Send the secretName for pods to use
+  const values = {
+    mountAllFromSecretName: `${deployment.releaseName.replace(/_/g, "-")}-env`
```
adam2k, line 85: Does this change only apply to updating deployment variables?  The `mapCustomEnvironmentVariables` function is also used in the update deployment worker.

This might be easier to read as something like:
```
const mountAllFromSecretName = `${deployment.releaseName.replace(/_/g, "-")}-env`;
const values = {
    mountAllFromSecretName
};
```

or if you want to take it further maybe turn this into a utility:

```
`${deployment.releaseName.replace(/_/g, "-")}-env`;
```

schnie:
@kushalmalani since we're passing the same values in two places, what do you think about adding a function (`environmentVariables(deployment)`) to this function directly https://github.com/astronomer/houston-api/blob/main/src/lib/deployments/config/index.js#L44-L66?

src/lib/deployments/config/index.js
```diff
@@ -72,6 +74,15 @@ export function generateHelmValues(deployment, values = {}) {
   return helmValues;
 }
 
+export function environmentVariables(deployment) {
+  // Send the secretName for pods to use
+  const values = {
+    mountAllFromSecretName: `${kebabCase(deployment.releaseName)}-env`
+  };
+
+  return values;
```
adam2k, line 83: This is a nit:

```Suggestion
  return {
    mountAllFromSecretName: `${kebabCase(deployment.releaseName)}-env`
  };
```

src/resolvers/mutation/update-deployment-variables/index.js
```diff
@@ -94,6 +90,7 @@ export default async function updateDeploymentVariables(parent, args, ctx) {
     .digest("hex");
 
   // This annotation is a sha512 hash of the user-provided Airflow environment variables
+  const values = {};
   values.airflowPodAnnotations = { "checksum/airflow-secrets": hash };
```
adam2k, line 94: This is a nit:

```
  const airflowPodAnnotations = { "checksum/airflow-secrets": hash };
  const values = { airflowPodAnnotations };
```

src/lib/deployments/config/index.js
```diff
@@ -72,6 +74,13 @@ export function generateHelmValues(deployment, values = {}) {
   return helmValues;
 }
 
+export function environmentVariables(deployment) {
+  // Send the secretName for pods to use
+  return {
+    mountAllFromSecretName: `${kebabCase(deployment.releaseName)}-env`
```
schnie, line 80: @kushalmalani we can probably skip the `kebabCase` here. We know it'll be correctly formatted at this point.

src/resolvers/mutation/update-deployment-variables/index.js
```diff
@@ -94,7 +90,8 @@ export default async function updateDeploymentVariables(parent, args, ctx) {
     .digest("hex");
 
   // This annotation is a sha512 hash of the user-provided Airflow environment variables
-  values.airflowPodAnnotations = { "checksum/airflow-secrets": hash };
+  const airflowPodAnnotations = { "checksum/airflow-secrets": hash };
+  const values = { airflowPodAnnotations };
```
schnie, line 94: @kushalmalani this is that last part I wanted to chat about in our last call (or before that). I think this looks fine, but just wanted to note that this part right here is the important part that makes sure that we actually reboot the pods that need to re-mount the secret if it changed due to new env var data.

This gets passed into all airflow pods, https://github.com/astronomer/airflow-chart/blob/master/templates/scheduler/scheduler-deployment.yaml#L61-L63, and makes sure to change the pod spec, forcing the reboot. At which point it'll remount all the new env vars.

Did you confirm that this part is working?

Also @adam2k I think this is an important thing that we'll need to deal with for the NATS implementation of this mutation. Since we'll be firing the secret update synchronously still, we'll need to pass this hash as part of the message (or store in db?), so that the worker that rolls the helm change out, can pass this annotation in.

PR #420, opened 2020-08-17
Adjust defaultExtraAu validation

samblackk:
## Description

Adjusts validation and fallback for `extra_au` between executors when `properties` isn't provided in the `createDeployment` mutation. 

## 🎟 Issue(s)

Attached to https://github.com/astronomer/issues/issues/1612

## 🧪 Functional Testing

1. Create a deployment via the CLI with `KubernetesExecutor`
2. Check the deployment config to find `properties.extra_au` is defined and set to `10`
3. Check UI to find `10` value for extra au slider

## 📋 Checklist

- [x] **N/A** Updated any related [documentation](https://github.com/astronomer/astro-docs/)


src/resolvers/mutation/create-deployment/index.js
```diff
@@ -114,8 +117,11 @@ export default async function createDeployment(_, args, ctx, info) {
 
   // Add default props if exists
   const properties = {
-    [DEPLOYMENT_PROPERTY_EXTRA_AU]: executorConfig.defaultExtraAu || 0,
-    ...args.properties
+    ...args.properties,
+    [DEPLOYMENT_PROPERTY_EXTRA_AU]:
+      (args.properties && args.properties.extra_au) ||
```
adam2k, line 122: can you use lodash get here?

```Suggestion
get(args, "properties.extra_au") ||
```

src/resolvers/mutation/create-deployment/index.js
```diff
@@ -114,8 +117,11 @@ export default async function createDeployment(_, args, ctx, info) {
...
+      executorConfig.defaultExtraAu ||
+      (isKube ? 10 : 0)
```
adam2k, line 124: This is a nit: 

Does it make sense to do this evaluation on line 56 under where you define `isKube`?

andriisoldatenko: and i prefer to extract `10` to constats.js or just module level constant. I don't like magic numbers.

adam2k: ah yeah, good idea!

samblackk:
@adam2k @andriisoldatenko Your suggestions are pushed.

src/lib/constants/index.js
```diff
@@ -90,3 +90,7 @@ export const DEPLOYMENT_CONFIG_KEYS = [
   "workers",
   "executor"
 ];
+
+// Resource related constants
+
+export const KUBERNETES_MIN_EXTRA_AU = 10;
```
schnie, line 96: Could we reuse the `defaultExtraAu` setting instead of creating a new const that mirrors it?

src/resolvers/mutation/create-deployment/index.js
```diff
@@ -112,10 +114,16 @@ export default async function createDeployment(_, args, ctx, info) {
...
+
   // Add default props if exists
   const properties = {
-    [DEPLOYMENT_PROPERTY_EXTRA_AU]: executorConfig.defaultExtraAu || 0,
-    ...args.properties
+    ...args.properties,
+    [DEPLOYMENT_PROPERTY_EXTRA_AU]:
+      get(args, "properties.extra_au") ||
+      executorConfig.defaultExtraAu ||
+      (isKube ? KUBERNETES_MIN_EXTRA_AU : 0)
```
schnie, line 126: @samblackk what would happen if someone makes an API request and specifies `KubernetesExecutor` with `properties.extra_au` set to 5, for example? Seems like it would allow that config to go through right?

samblackk: From the UI - you can't set it to anything less that 10. And in the CLI you can't configure it at all. We can add a catch here, but from our components a user is locked to 10 or up.

samblackk:
@schnie Applied your recommendations.

src/resolvers/mutation/create-deployment/index.js
```diff
@@ -99,10 +99,24 @@ export default async function createDeployment(_, args, ctx, info) {
...
+    args,
+    "properties.extra_au",
+    executorConfig.defaultExtraAu
+  );
+
+  // Validate extra au amount is at least the default
+  const extraAu =
+    providedExtraAU < executorConfig.defaultExtraAu
+      ? executorConfig.defaultExtraAu
+      : providedExtraAU;
```
schnie, line 114: @samblackk nice! I guess this raises another question. Should we auto-bump them here, like we have, or should we throw a validation error like we do for other invalid API requests? I know our UI and CLI won't have this issue, but just wondering in the case of an outside request.

samblackk: @mrrobby @andriisoldatenko thoughts? Should we throw validation error or simply auto-set the correct extraAu if none is provided?

andriisoldatenko: i prefer to fallback to default value, instead of raise error.

schnie: Maybe fallback is fine for now. Once the schema has been cleaned up a bit more, we could look at documenting and validating the input more precisely.

PR #422, opened 2020-08-18
Bring all Prisma2 changes into the main branch

andriisoldatenko:


andriisoldatenko:
@schnie @adam2k @kushalmalani @samblackk PTAL!

src/index.js
```diff
@@ -1,37 +1,40 @@
...
 const helmConfig = config.get("helm");
-const prismaConfig = config.get("prisma");
 
 // Enable global proxy support for got, axios, and request libs. set
 // GLOBAL_AGENT_HTTPS_PROXY, GLOBAL_AGENT_HTTP_PROXY, or GLOBAL_AGENT_NO_PROXY
 // environment variables to control. It does not look at HTTP_PROXY or
 // http_proxy etc variables.
 import "global-agent/bootstrap";
 
+const prisma = new PrismaClient();
```
adam2k, line 33: Does this prisma client need to `disconnect()` on app shutdown or does that happen automatically?

src/lib/deployments/config/index.unit.test.js
```diff
@@ -797,37 +797,6 @@ describe("deploymentOverrides", () => {
     expect(res.webserver.resources.requests.memory).toEqual("384Mi");
     expect(res.workers.replicas).toEqual(1);
   });
-
```
adam2k, line 800: Is this test being moved somewhere else or is there a similar test in its place?

src/lib/deployments/database/index.js
```diff
@@ -246,6 +246,12 @@ export async function ensureUserAndSchema(
  * @param {Object} The parsed connection object.
  */
 export function parseConnection(conn) {
-  if (isString(conn)) return parse(conn);
+  const { ssl } = config.get("database");
+  if (isString(conn)) {
+    return {
+      ssl: ssl,
```
adam2k, line 252:
```Suggestion
      ssl,
```

knexfile.js
```diff
@@ -1,14 +1,18 @@
...
-const { connection, migrations } = config.get("database");
+const { connection, migrations, ssl } = config.get("database");
 
 // Export the knex configuration.
 // This needs to be old school export for knex.
 module.exports = {
   client: "postgres",
-  connection,
+  connection: {
+    ssl: ssl,
```
adam2k, line 13:
```Suggestion
      ssl,
```

src/index.js
andriisoldatenko: we never shutdown houston api.

knexfile.js
```diff
@@ -1,14 +1,18 @@
...
+    ...parse(connection)
+  },
```
adam2k, line 15:
```Suggestion
  connection: parseConnection(connection)
```

src/lib/deployments/validate/authorization/index.js
```diff
@@ -14,19 +15,36 @@ export default async function validateDeploymentCredentials(
   password,
   passwordField
 ) {
+  const prisma = new PrismaClient();
```
adam2k, line 18: Does this cause an error if lines 23 or lines 31 return false?  That would mean this connection doesn't close on line 41.  I'd suggest moving this line to 34 to prevent that issue.

src/lib/deployments/validate/authorization/index.js
```diff
@@ -14,19 +15,36 @@ export default async function validateDeploymentCredentials(
...
   // Get the password for this deployment
-  const truePassword = await prisma
-    .deployment({ releaseName })
-    [passwordField]();
+  const deployment = await prisma.deployment.findOne({
+    where: { releaseName: releaseName },
+    select: { [passwordField]: true }
+  });
+
+  const truePassword = deployment ? deployment[passwordField] : null;
```
adam2k, line 40: Do we need this ternary? Is it possible that deployment would be null here?

It will probably work like:

`const truePassword = deployment[passwordField];` `// it could be a real value or undefined

if deployment can be null then you'd need to check it

`const truePassword = deployment && deployment[passwordField];`

knexfile.js
andriisoldatenko: it won't work, we need `ssl: true`.

src/index.js
```diff
@@ -60,12 +65,7 @@ const server = new ApolloServer({
   context: ctx => {
     const context = {
       ...ctx,
-      db: new Prisma({
-        typeDefs: "src/lib/generated/schema/prisma.graphql",
-        endpoint: prismaConfig.endpoint,
-        secret: prismaConfig.secret,
-        debug: prismaConfig.debug
-      }),
+      prisma: prisma,
```
schnie, line 68:
```suggestion
      prisma,
```

src/lib/authentication/index.js
```diff
@@ -40,9 +41,11 @@ export async function wsOnConnect(connParams) {
     ""
   ).replace("Bearer ", "");
 
+  const prisma = new PrismaClient();
   // Find the user
-  const user = await getAuthUser(token);
+  const user = await getAuthUser(prisma, token);
 
+  await prisma.disconnect();
```
schnie, line 48: Just a question - should we generate `prisma` client here and pass into `getAuthUser()` or create the client inside the function? Just asking because it doesn't seem to be used anywhere else in here.

src/lib/rbac/index.js
```diff
@@ -43,20 +39,20 @@ export function hasPermission(user, permission, entityType, entityId) {
...
 
+  const thisBind = binding[0];
+
   // If we didn't find a roleBinding, return false.
-  if (!binding) return false;
+  if (!thisBind) return false;
 
   // Otherwise return if this role has an appropriate permission.
-  const role = get(ROLES, binding.role, { permissions: [] });
+  const role = get(ROLES, thisBind.role, { permissions: [] });
```
schnie, line 53: Why did this change? I'd imagine this could would be unaffected.

prisma/migrations/20200812220552-init/README.md
```diff
@@ -0,0 +1,325 @@
+# Migration `20200812220552-init`
```
schnie, line 1: Just a question - What's the difference between this migration and the knex migrations? Do we still need to use both going forward?

Also, what is causing the need for this migration? Just want to make sure I understand.

src/lib/authentication/index.js
andriisoldatenko: yes we can.

src/lib/rbac/index.js
andriisoldatenko: cc @samblackk can you remind me why we change this?

prisma/migrations/20200812220552-init/README.md
andriisoldatenko: @schnie 
in prisma1 we typically can do: `prisma deploy --force` (it can create new database and tables with new columns).
in prisma2 we don't have similar command, but we need to generate migrations for tables, then we can run houston app, otherwise if we run houston assuming we need to create new DB/tables it causes error: "no database/ no tables".

Also useful to know from docs (https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-migrate):
>Prisma Migrate is a tool that lets you change your database schema, e.g. by creating new tables or adding columns to existing tables. These changes are called schema migrations. Prisma Migrate is available as part of the Prisma CLI via the prisma migrate command.

> Prisma Migrate is currently in an experimental state. This means that it is not recommended to use Prisma Migrate in production. Instead, you can perform schema migrations using plain SQL or another migration tool of your choice and then bring the changes into your Prisma schema using introspection.

regarding knex... i see we doing some manipulations with tables... so at least during migration process to prisma2 we still need knex + we using knex on top on prisma migrations to tweak some constraints. We can't move these new knex migrations to prisma2 migrations because if we run on top on existing DB it crashed with errors like column doesn't exists or we are trying to drop empty columns.

Regarding cascading deletes it useful to read [here](https://www.prisma.io/docs/guides/database-workflows/cascading-deletes/postgresql), but shortly prisma2 delegates cascading deletes to database rather than to javascript like prisma1, it means we need `src/migrations/20200625161933_add-on-delete-cascade-prisma2.js`

Summary:
**Option A (empty database fresh installation):**
- `prisma migrate up --experimental --create-db` it creates new tables and database or check if it exists and skip based on `src/migrations/20190612121012_fix_relationships.js`
- `knex migrate` we use to add on delete cascade for prisma2 `20200625161933_add-on-delete-cascade-prisma2.js`, we can't move it to prisma migrations, because it doesn't support some custom SQL only schema changes.

**Option B (non empty database, tables/daata from prisma1):**
run custom SQL:
```
ALTER TABLE "houston$default"."RoleBinding" RENAME COLUMN "user" TO "userId";
ALTER TABLE "houston$default"."RoleBinding" RENAME COLUMN "serviceAccount" TO "serviceAccountId";
ALTER TABLE "houston$default"."RoleBinding" RENAME COLUMN "workspace" TO "workspaceId";
ALTER TABLE "houston$default"."RoleBinding" RENAME COLUMN "deployment" TO "deploymentId";
ALTER TABLE "houston$default"."Email" RENAME COLUMN "user" TO "userId";
ALTER TABLE "houston$default"."OAuthCredential" RENAME COLUMN "user" TO "userId";
ALTER TABLE "houston$default"."InviteToken" RENAME COLUMN "workspace" TO "workspaceId";
ALTER TABLE "houston$default"."Deployment" RENAME COLUMN "workspace" TO "workspaceId";
ALTER TABLE "houston$default"."Deployment" alter column config type jsonb using config::JSON;
ALTER TABLE "houston$default"."DockerImage" alter column labels type jsonb using labels::JSON;
ALTER TABLE "houston$default"."DockerImage" alter column env type jsonb using env::JSON;
ALTER TABLE "houston$default"."DockerImage" RENAME COLUMN "deployment" TO "deploymentId";
ALTER TABLE "houston$default"."User" RENAME COLUMN "localCredential" TO "localCredentialId";
CREATE TYPE "houston$default"."Role" AS ENUM ('SYSTEM_ADMIN', 'SYSTEM_EDITOR', 'SYSTEM_VIEWER', 'WORKSPACE_ADMIN', 'WORKSPACE_EDITOR', 'WORKSPACE_VIEWER', 'DEPLOYMENT_ADMIN', 'DEPLOYMENT_EDITOR', 'DEPLOYMENT_VIEWER', 'USER');
CREATE TYPE "houston$default"."InviteSource" AS ENUM ('SYSTEM', 'WORKSPACE');
alter table "houston$default"."InviteToken" ALTER COLUMN source TYPE "houston$default"."InviteSource" using source::"InviteSource";
alter table "houston$default"."InviteToken" ALTER COLUMN role TYPE "houston$default"."Role" using role::"Role";
alter table "houston$default"."RoleBinding" ALTER COLUMN role TYPE "houston$default"."Role" using role::"Role";
```
and than same steps as Option A)

knexfile.js
adam2k: It always has to be `ssl: true`?  Should we make that change in parseConnection and use it in both places where we connect?

adam2k: wait...I'm actually kind of confused by that comment because in this file we have:

```
const { connection, migrations, ssl } = config.get("database");
```

And inside of the `parseConnection` function we have:

```
const { ssl } = config.get("database");
```

It's referencing the exact same `ssl` env var, right?

prisma/migrations/migrate.lock
```diff
@@ -0,0 +1,3 @@
+# Prisma Migrate lockfile v1
+
+20200812220552-init
```
adam2k, line 3: Is this auto generated or do we need a newline at EOF here?

prisma/schema.prisma
```diff
@@ -0,0 +1,180 @@
+generator client {
```
adam2k, line 1: Can you confirm this is auto-generated, so we do need to review?

src/lib/deployments/validate/properties/index.js
```diff
@@ -4,7 +4,7 @@ import * as constants from "constants";
 
 // Validate component resources
 export default function validateProperties(properties) {
-  // Filter down allowlisted deployment properties.
+  // Filter down whitelisted deployment properties.
```
adam2k, line 7: Can you revert this?  `allowlisted` is a better term here.  See this ruby issue as an example of why: https://github.com/rails/rails/issues/33677

knexfile.js
andriisoldatenko: yes

prisma/migrations/migrate.lock
andriisoldatenko: it's auto generated.

prisma/schema.prisma
andriisoldatenko: it's auto generated. pls skip this folder.

src/lib/deployments/validate/properties/index.js
andriisoldatenko: yep, it was bad merge resolve!

src/lib/logger/index.js
```diff
@@ -1,27 +1,21 @@
...
-const { combine, colorize, errors, timestamp, printf, splat } = format;
+const { combine, colorize, timestamp, printf, splat } = format;
 
 const baseFormatter = combine(
   format(info => {
     info.level = `${info.level.toUpperCase()}`;
     return info;
   })(),
   splat(),
-  errors({ stack: true }),
```
adam2k, line 11: Why aren't we printing the stack anymore?  Is this an unrelated change?

andriisoldatenko: same here bad merge.

src/lib/rbac/index.js
```diff
@@ -109,20 +105,50 @@ export function checkSystemPermission(user, permission) {
...
+            role: true,
+            workspace: true,
+            deployment: true,
+            serviceAccount: true
+          }
+        }
+      }
+    });
+
+    return user;
```
adam2k, line 126: This is a nit, but consider:

```Suggestion
    return await prisma.user.findOne({
      where: { id },
      include: {
        roleBindings: {
          select: {
            id: true,
            user: true,
            role: true,
            workspace: true,
            deployment: true,
            serviceAccount: true
          }
        }
      }
    });
```

src/resolvers/mutation/add-card/index.unit.test.js
```diff
@@ -1,15 +1,7 @@
-import resolvers from "resolvers";
+import { schema } from "../../../schema";
```
adam2k, line 1:
```Suggestion
import { schema } from "schema";
```

src/resolvers/mutation/add-customer-id/index.unit.test.js
```diff
@@ -1,14 +1,6 @@
-import resolvers from "resolvers";
+import { schema } from "../../../schema";
```
adam2k, line 1:
```Suggestion
import { schema } from "schema";
```

src/resolvers/mutation/confirm-email/index.js
```diff
@@ -10,19 +10,10 @@ import { USER_STATUS_ACTIVE, USER_STATUS_PENDING } from "constants";
...
-      verified
-      primary
-      user {
-        id
-        status
-      }
-     }`
-  );
+  const email = await ctx.prisma.email.findOne({
+    where: { token: args.token },
```
adam2k, line 14: Is it possible for `args` to be `null`?  This could throw if `args` can be null.

src/resolvers/mutation/confirm-email/index.unit.test.js
```diff
@@ -1,8 +1,6 @@
-import resolvers from "resolvers";
+import { schema } from "../../../schema";
```
adam2k, line 1:
```Suggestion
import { schema } from "schema";
```

src/resolvers/mutation/create-deployment/index.js
```diff
@@ -25,12 +23,12 @@ import {
...
 /*
  * Create a deployment.
- * @param {Object} _ The result of the parent resolver.
+ * @param {Object} parent The result of the parent resolver.
  * @param {Object} args The graphql arguments.
  * @param {Object} ctx The graphql context.
  * @return {Deployment} The newly created Deployment.
  */
-export default async function createDeployment(_, args, ctx, info) {
+export default async function createDeployment(parent, args, ctx) {
```
adam2k, line 31:
```Suggestion
export default async function createDeployment(_, args, ctx) {
```

src/resolvers/mutation/create-deployment/index.js
```diff
@@ -25,12 +23,12 @@ import {
 
 /*
  * Create a deployment.
- * @param {Object} _ The result of the parent resolver.
+ * @param {Object} parent The result of the parent resolver.
```
adam2k, line 26:
```Suggestion
 * @param {Object} _ The result of the parent resolver.
```

src/resolvers/mutation/confirm-email/index.unit.test.js
```diff
@@ -25,21 +23,15 @@ mutation confirmEmail($token: String!) {
...
-});
-
 describe("confirmEmail", () => {
   const token = casual.uuid;
 
   describe("when token is invalid", () => {
     test("should return an error", async () => {
       const q = jest.fn().mockReturnValue(null);
-      const db = { query: { email: q } };
+      const prisma = { email: { findOne: q } };
```
adam2k, line 32: To match the other tests:

```Suggestion
      const findOne = jest.fn().mockReturnValue(null);
      const prisma = { email: { findOne } };
```

src/resolvers/mutation/create-system-rolebinding/index.unit.test.js
```diff
@@ -1,16 +1,8 @@
-import resolvers from "resolvers";
+import { schema } from "../../../schema";
```
adam2k, line 1:
```Suggestion
import { schema } from "schema";
```

src/resolvers/mutation/create-system-service-account/index.unit.test.js
```diff
@@ -1,16 +1,8 @@
-import resolvers from "resolvers";
+import { schema } from "../../../schema";
```
adam2k, line 1:
```Suggestion
import { schema } from "schema";
```

src/resolvers/mutation/create-token/index.js
```diff
@@ -18,22 +18,29 @@ import { USER_STATUS_ACTIVE, USER_STATUS_PENDING } from "constants";
...
+      },
+      select: {
+        id: true,
+        status: true,
+        localCredential: { select: { password: true } }
       }
-    },
-    `{ id, status, localCredential { password } }`
+    }
+    // `{ id, status, localCredential { password } }`
```
adam2k, line 43: Can you remove this line?

src/resolvers/mutation/delete-deployment-service-account/index.js
```diff
@@ -11,28 +10,24 @@ import { addFragmentToInfo } from "graphql-binding";
...
 
   // Look for the service account.
-  const serviceAccount = await ctx.db.query.serviceAccount(
+  const serviceAccount = await ctx.prisma.serviceAccount.findOne(
     {
       where: { id: serviceAccountUuid }
-    },
-    `{ roleBinding { workspace { id }, deployment { id } } }`
+    }
+    // `{ roleBinding { workspace { id }, deployment { id } } }`
```
adam2k, line 23: Can you remove this line?

src/resolvers/mutation/delete-deployment-service-account/index.unit.test.js
```diff
@@ -43,17 +36,20 @@ describe("deleteDeploymentServiceAccount", () => {
     };
 
     // Mock up some db functions.
-    const serviceAccount = jest.fn().mockReturnValue({
-      id: casual.id,
+    const findOne = jest.fn().mockReturnValue({
+      id: id,
```
adam2k, line 40: This is a nit:

```Suggestion
      id,
```

src/resolvers/mutation/delete-deployment-service-account/index.unit.test.js
```diff
@@ -43,17 +36,20 @@ describe("deleteDeploymentServiceAccount", () => {
...
-
     // Construct db object for context.
-    const db = {
-      query: { serviceAccount },
-      mutation: { deleteServiceAccount }
+    const prisma = {
+      serviceAccount: {
+        findOne,
+        delete: jest.fn().mockReturnValue({
+          id: id,
```
adam2k, line 49: This is a nit:
```Suggestion
          id,
```

src/resolvers/mutation/delete-deployment/index.js
```diff
@@ -10,23 +8,20 @@ import {
...
 /*
  * Delete a deployment.
- * @param {Object} _ The result of the parent resolver.
+ * @param {Object} parent The result of the parent resolver.
  * @param {Object} args The graphql arguments.
  * @param {Object} ctx The graphql context.
  * @return {Deployment} The deleted Deployment.
  */
-export default async function deleteDeployment(_, args, ctx, info) {
+export default async function deleteDeployment(parent, args, ctx) {
```
adam2k, line 16: Can you revert back to `_` because we're not using the `parent` variable.

src/resolvers/types/auth-user/index.js
```diff
@@ -1,23 +1,40 @@
...
 // Grab the user object for this id.
-export function user(parent, args, ctx, info) {
-  return ctx.db.query.user(
-    { where: { id: parent.userId } },
-    addFragmentToInfo(info, fragment)
-  );
+export async function user(parent, args, context) {
+  return await context.prisma.user.findOne({
+    where: { id: parent.userId },
+    select: {
```
adam2k, line 9: Same comment here about the selection object. Could we separate into a different file?

src/routes/v1/elasticsearch/handler.js
```diff
@@ -8,6 +9,9 @@ import validateDeploymentCredentials from "deployments/validate/authorization";
 export default async function(req, res) {
   const authorization = req.get("authorization");
   if (!authorization || authorization.substr(0, 5) !== "Basic") {
```
adam2k, line 11: Should `"Basic"` be in the constants file?

src/routes/v1/oauth/callback/index.unit.test.js
```diff
@@ -1,518 +1,47 @@
...
-    prisma.updateUser = jest
-      .fn()
-      .mockName("updateUser")
-      .mockReturnValue({
-        $fragment: () => true
-      });
-  });
-
-  afterEach(() => {
+describe.skip("POST /oauth", () => {
```
adam2k, line 8: Is this intentionally skipped and are these tests being move to another test file?

src/schema/mutation.js
```diff
@@ -0,0 +1,445 @@
...
+    t.field("updateDeployment", {
+      type: "Deployment",
+      args: {
+        deploymentUuid: arg({ type: "Uuid", nullable: true }),
+        payload: jsonArg({ nullable: true }),
+        executor: arg({ type: "ExecutorType", nullable: true }),
+        workers: arg({ type: "Workers", nullable: true }),
+        webserver: arg({ type: "Webserver", nullable: true }),
+        scheduler: arg({ type: "Scheduler", nullable: true }),
+        // TODO: add deprecation to `config` and `sync`
```
adam2k, line 106: Is there another GitHub issue to add this deprecation notice for config and sync?

src/schema/permissions.js
```diff
@@ -0,0 +1,245 @@
...
+  ctx.user = ctx.req.session.user;
+  const exists = ctx.user !== null;
+  if (!exists) throw new PermissionError();
+  return exists;
+});
+
+const hasRBAC = role =>
+  rule({ cache: "contextual" })(async (parent, args, ctx) => {
+    // Get the 1st constraint of the permission
+    let entityType = role.split(".")[0];
```
adam2k, line 15:
```Suggestion
    const entityType = role.split(".")[0];
```

src/schema/permissions.js
```diff
@@ -0,0 +1,245 @@
...
+        break;
+      case "serviceAccount":
+        entityId = args.serviceAccountUuid;
+        break;
+      default:
+        break;
+    }
+
+    const allowed = hasPermission(ctx.user, role, entityType, entityId);
+    return allowed;
```
adam2k, line 34:
```Suggestion
    return hasPermission(ctx.user, role, entityType, entityId);
```

src/schema/types.js
```diff
@@ -0,0 +1,889 @@
...
+    "The `DeploymentConfig` compound type contains data describing available options for Airflow deployments in the system.",
+  definition(t) {
+    t.field("defaults", { type: "JSON" });
+    t.field("limits", { type: "JSON" });
+    t.field("astroUnit", { type: "AstroUnit" });
+    t.int("maxExtraAu", { nullable: true });
+    t.field("executors", { type: "JSON" });
+    t.boolean("singleNamespace");
+    t.boolean("loggingEnabled");
+    t.string("latestVersion"); // @depricated
```
adam2k, line 733:
```Suggestion
    t.string("latestVersion"); // @deprecated
```

src/scripts/upgrade-deployments/index.js
```diff
@@ -31,19 +41,28 @@ async function upgradeDeployments() {
...
+        airflowVersion: true,
+        releaseName: true,
+        workspace: { select: { id: true } }
+      }
+    };
     log.debug(`Query: ${JSON.stringify(q1)}`);
-    deployments = await prisma
-      .deployments(q1)
-      .$fragment(`{ id releaseName version workspace { id } }`);
+    deployments = await prisma.deployment.findMany(q1);
```
adam2k, line 60: Is there a better name for this than `q1`?  This query is looking for deployments that are not deleted, right?  Would something like `const notDeleted = ...` or something similar work better?

src/workers/deployment-deleted/index.js
```diff
@@ -44,15 +45,22 @@ export async function deploymentDeleted(msg) {
     log.info(`Deployment ${releaseName} successfully deleted`);
   } catch (err) {
     log.error(err);
+  } finally {
+    await prisma.disconnect();
   }
 }
 
 /**
  * @param  {String} id deployment id
```
adam2k, line 54: This comment will need to be updated to add the `prisma` object.

src/workers/deployment-image-update/index.js
```diff
@@ -73,8 +75,18 @@ async function commanderUpdateDeployment(deployment) {
  * @param  {String} id for the deployment to request via prisma
```
adam2k, line 75: Can you add the `prisma` object to the comment block here?

src/lib/deployments/logs/index.js
```diff
@@ -28,7 +28,7 @@ export const generateLogMessage = (function* generateLogs() {
 export function generateMockLogRecords(release, component) {
   return {
     hits: {
-      hits: range(5).map(() => ({
+      hits: range(random(2)).map(() => ({
```
kushalmalani, line 31: Why are we changing this? random(2) could pick 0 as a number and that would not generate any mock logs

src/resolvers/mutation/create-token/index.unit.test.js
```diff
@@ -1,17 +1,9 @@
-import resolvers from "resolvers";
+import { schema } from "../../../schema";
```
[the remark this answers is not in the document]
andriisoldatenko: @kushalmalani 
```
npx jest src/resolvers/query/self/index.unit.test.js
 FAIL  src/resolvers/query/self/index.unit.test.js
  ● Test suite failed to run

    Cannot find module 'schema' from 'index.unit.test.js'

    > 1 | import { schema } from "schema";
        | ^
      2 | import casual from "casual";
      3 | import { graphql } from "graphql";
      4 |

      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:259:17)
      at Object.<anonymous> (src/resolvers/query/self/index.unit.test.js:1:1)
```
unfortunately but no.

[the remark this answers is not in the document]
adam2k: @andriisoldatenko I think you can resolve these imports by adding the schema alias here: 

https://github.com/astronomer/houston-api/blob/migrate-to-prisma2/.babelrc#L24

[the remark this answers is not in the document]
andriisoldatenko: i'll try.

src/schema/subscription.js
```diff
@@ -0,0 +1,52 @@
+import deploymentStatus from "../resolvers/subscription/deployment-status";
+import deploymentLogs from "../resolvers/subscription/log";
+import deploymentMetrics from "../resolvers/subscription/metrics";
+
```
kushalmalani, line 4: Nit: Can we consolidate all imports together?

src/index.js
adam2k: When new pods are initializing and the existing pods are terminating is there only a `SIGTERM` to Houston API?  Will that orphan this prisma client connection or is there some automatic cleanup that happens?

knexfile.js
adam2k: Why can't we reuse the `parseConnection(connection)` function in both places then so we're not duplicating code?

src/scripts/upgrade-deployments/index.js
andriisoldatenko: @adam2k can we please avoid refactoring or improvements of readability just to not mix prisma1/prisma2 changes?

https://github.com/astronomer/houston-api/blob/main/src/scripts/upgrade-deployments/index.js#L35-L38

src/schema/subscription.js
andriisoldatenko: @kushalmalani can you suggest change? not sure that understand you.

src/resolvers/mutation/delete-workspace/index.unit.test.js
```diff
@@ -1,14 +1,6 @@
-import resolvers from "resolvers";
+import { schema } from "../../../schema";
```
adam2k, line 1: Did the `jest config` change work to make this line alias to: 

`import { schema } "schema";`

src/resolvers/mutation/update-deployment/index.js
```diff
@@ -23,15 +21,15 @@ import {
...
   const pub = publishers.get(UPDATE_DEPLOYMENT);
   pub.publish(DEPLOYMENT_UPDATED_STARTED, args.workspaceUuid);
   const id = args.deploymentUuid;
   // Get the deployment first.
-  const deployment = await ctx.db.query.deployment(
-    { where: { id } },
-    queryFragment
-  );
+  const deployment = await ctx.prisma.deployment.findOne({
+    where: { id: args.deploymentUuid },
```
adam2k, line 30:
```suggestion
    where: { id },
```

src/resolvers/subscription/deployment-status/index.js
```diff
@@ -56,15 +56,14 @@ export async function getMetric(releaseName) {
 }
 
 // Start the subscription
-export async function subscribe(parent, args, { pubsub }) {
-  // Return sample data (prom not available)
+export async function subscribe(parent, args, ctx, { pubsub }) {
```
adam2k, line 59: Do we need `ctx` here or can this be removed?

src/routes/v1/registry/events/post.js
```diff
@@ -36,9 +36,17 @@ export default async function(req, res) {
...
       );
 
-      const deployment = await prisma
-        .deployment({ releaseName })
-        .$fragment(`{ config deletedAt airflowVersion }`);
+      const prisma = new PrismaClient();
+
+      // Get the existing config for this deployment.
+      const deployment = await prisma.deployment.findOne({
+        where: { releaseName: releaseName },
```
adam2k, line 43:
```suggestion
        where: { releaseName },
```

src/schema/subscription.js
adam2k: Do you mean like:

```
import  { deploymentStatus, deploymentLogs, deploymentMetrics } from "resolvers";
```

src/schema/types.js
```diff
@@ -0,0 +1,886 @@
+import authConfig from "../resolvers/types/auth-config";
```
adam2k, line 1: Is there an automated way to validate these types?  Do we need to write tests for these or does Prisma validate these with the DB automagically?

andriisoldatenko: @adam2k not sure that understand you, it works 1 to 1 as in prisma1. Only difference now it's written explicitly.

adam2k: I'm asking how do we confirm that it's all correct?  During the functional testing only?

prisma1_to_prisma2.sql
```diff
@@ -0,0 +1,21 @@
+set search_path to "houston$default";
```
adam2k, line 1: Are there tests for this custom sql migration?

andriisoldatenko: it's tested via installing prisma1 than create data e.g. users/deployments/workspaces than run `prisma1_to_prisma2.sql` and after we can run `helm upgrade`. It's custom SQL script for infra team to make release and don't loose any data.

prisma1_to_prisma2.sql
```diff
@@ -0,0 +1,21 @@
...
+ALTER TABLE "houston$default"."RoleBinding" RENAME COLUMN "user" TO "userId";
```
adam2k, line 2: Is there a requirement for camel cased database columns for prisma?  It's typically snake case in SQL databases and then a mutation in the codebase to switch it to camel case for our application code.  We can do it fairly easily with the lodash library.

prisma1_to_prisma2.sql
adam2k: Is there a helper script to confirm the changes that they can use to make sure we don't lose anything in prod?  It seems kind of risky otherwise.

prisma1_to_prisma2.sql
adam2k: nvm...I see `localCredential` is already camel cased.  We can discuss this in one of our engineering meetings later.

src/migrations/20200625161933_add-on-delete-cascade-prisma2.js
```diff
@@ -0,0 +1,66 @@
+import log from "logger";
+import dotenv from "dotenv";
+
+dotenv.config();
```
adam2k, line 4: Do we need `dotenv` or can we use `import config from "config";` here instead?

prisma1_to_prisma2.sql
adam2k: Another idea, couldn't this be a knex migration instead?  Then we can rollback if something goes wrong with the query?  We'll also be able to reproduce the exact migration and it'll make the enterprise upgrades easier.  It's one less manual step for us.

adam2k: I can help you out if you'd like for migrating this sql into a knex migration.

andriisoldatenko: this file unrelated to this repo, instead of slack i just committed it somewhere. We can't use knex because we run knex after we run prisma migrate.

andriisoldatenko: but we need to run this script before prisma2 release.

src/migrations/20200625161933_add-on-delete-cascade-prisma2.js
andriisoldatenko: fixed in 9317c59b41483d30b00c5a89fda13df109f62fd4

prisma1_to_prisma2.sql
adam2k: That makes sense.  Can we move this script here so the infra team can access it easier during the LTS upgrades? https://github.com/astronomer/astronomer/tree/master/bin/migration-scripts

src/resolvers/mutation/create-token/index.unit.test.js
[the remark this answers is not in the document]
adam2k: This fix was merged in.

src/lib/deployments/logs/index.js
adam2k: It's probably OK because the logs could be empty, right?

adam2k:
This is looking really good!  I'm getting one additional bug on the AWS install when I try to create a workspace.

Screen Shot 2020-09-08 at 7 03 16 PM

src/lib/rbac/index.js
schnie: @samblackk any idea?

knexfile.js
andriisoldatenko: @adam2k fixed in 7213e3b43fad93cd3c6f913c4afabf94474134ef

src/index.js
andriisoldatenko: @schnie fixed in 55277a25a4c24ce56b1c8634b4393a96945c24ea

src/lib/deployments/validate/properties/index.js
andriisoldatenko: @adam2k  fixed in 3449ed55c971974ff7beb75f07e4eed2df8f5c2f

src/resolvers/mutation/delete-workspace-service-account/index.js
```diff
@@ -8,31 +7,23 @@ import { addFragmentToInfo } from "graphql-binding";
...
 
   // Look for the service account.
-  const serviceAccount = await ctx.db.query.serviceAccount(
+  const serviceAccount = await ctx.prisma.serviceAccount.findOne(
     {
       where: { id: serviceAccountUuid }
-    },
-    `{ roleBinding { workspace { id }, deployment { id } } }`
+    }
+    // `{ roleBinding { workspace { id }, deployment { id } } }`
```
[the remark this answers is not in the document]
andriisoldatenko: @adam2k fixed in 6fce82a5a114be4658ca21c02910c38a44422052

src/lib/jwt/index.unit.test.js
```diff
@@ -14,18 +13,6 @@ describe("createAuthJWT", () => {
   });
 });
 
-describe("createJWS", () => {
```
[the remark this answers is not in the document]
andriisoldatenko: @kushalmalani reverted back your test.

src/resolvers/subscription/deployment-status/index.js
andriisoldatenko: yep

src/lib/jwt/index.unit.test.js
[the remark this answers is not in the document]
andriisoldatenko: https://github.com/astronomer/houston-api/pull/422/commits/fc14707bd37f53daea0224abd7e73a425626320e

kushalmalani (approved):
We can ignore the nits and cosmetic comments. The rest looks good to me as long as its functional on the cluster.

adam2k (approved):
Nice work!  Goodbye, Prisma 1!  🚀 🔺 📈 🎉

schnie (approved):

PR #435, opened 2020-08-25
Make workspace required for Deployments table

sjmiller609:
## Description

Should we make more fields in required? What other fields could benefit from the required property, if any?

## 🎟 Issue(s)

We found while shipping 0.19 to stage cloud 17 deployments with no workspace that were orphaned and did not show up in the UI but did have a row in the Deployment table and existed in kubernetes. We deleted them manually, and did not find any such deployments on production.

## 🧪 Functional Testing

todo: draft PR

## 📸 Screenshots

## 📋 Checklist


andriisoldatenko:
@sjmiller609 IMO it won't work, you have to write data migration like you did manually. 
Also you have to change graphql schema.

schnie:
Looks good, `version` and `airflowVersion` could be made required, but I'd hold off until we get to the story about airflow versioning, which should be up soon.

schnie:
I assumed it would work since we will always populate this foreign key when creating a deployment. I would think this field is fully under our control and wouldn't require any changes to the GraphQL schema.

sjmiller609:
Can one of you please give me the entrypoint for data migration (like where does it go, where are the existing migrations)?

For what it ought to do - I'm thinking the expected behavior should be fail data migration, causing houston to crash with informative log, if we find that there is a deployment without a workspace? Another option is delete them automatically, but that seems a bit risky, but on the other hand I think these would be missing from the UI... what do you think?

andriisoldatenko:
@schnie @sjmiller609 
Quick test locally of this patch:
```diff
diff --git database/datamodel.graphql database/datamodel.graphql
index 7831eff..04cf7b9 100644
--- database/datamodel.graphql
+++ database/datamodel.graphql
@@ -130,7 +130,7 @@ type Deployment {
   airflowVersion: String
   alertEmails: [String!] @scalarList(strategy: RELATION)
   roleBindings: [RoleBinding!]! @relation(name: "DeploymentRoleBindings", onDelete: CASCADE)
-  workspace: Workspace
+  workspace: Workspace!
   createdAt: DateTime! @createdAt
   updatedAt: DateTime! @updatedAt
   deletedAt: DateTime
```


```
postgres=# select "releaseName",workspace from "Deployment";
        releaseName         | workspace
----------------------------+-----------
 planetic-telescope-3522    |
 native-nova-3615           |
 mathematical-universe-3271 |
 heliocentric-synodic-4047  |
 galactian-sunspot-0616     |
 ancient-syzygy-8403        |
 advanced-spectroscope-6369 |
(7 rows)
```

```
$ prisma deploy --force
Deploying service `houston` to stage `default` to server `local` 840ms

Errors:

  Deployment
    ✖ You are updating the field `workspace` to be required. But there are already nodes for the model `Deployment` that would violate that constraint.
```

andriisoldatenko:
@schnie @sjmiller609 
we should wait until prisma2 or change rules... we run knex after ` prisma deploy --force`, so we can't fix it without manual steps e.g.:
```
delete from "Deployment" where workspace is null;
```

sjmiller609:
I wonder how these got in there in the first place... but yes I agree with andrii we should consider the case where they are present, even though it only showed up in stage cloud and not prod cloud.

sjmiller609:
in stage cloud we also found the kube resources existed behind the scenes. I think it's tricky to know what to do with them, like should they be deleted or not?

schnie:
Do we have Deployment records in prod without a Workspace now? I can't think of a time when that would ever be the case. Could those local ones be from development when changing things around?

The workspace id should be set on deployment creation and never touched again.

adam2k:
> Do we have Deployment records in prod without a Workspace now? I can't think of a time when that would ever be the case. Could those local ones be from development when changing things around?
> 
> The workspace id should be set on deployment creation and never touched again.

@schnie there were a bunch of deployments in Dev without Workspace IDs.  It looks like that was the root cause of the high active DB connections.  It causes un-acked messages in the workers.  I think we should apply this change and consider making the `workspaceId` non-null in the `deployment` table.

PR #437, opened 2020-08-25
Feature/user email validation

adam2k:
## Description

Resolve invalid email addresses being used when adding a user to a workspace.  

## 🎟 Issue(s)

Resolves astronomer/issues#1630

## 🧪 Functional Testing

1. Start Houston API
2. Using astro cli (this is version 0.19.0, but 0.18.0 or other previous versions should also work) run the following commands:

```
$ astro auth login localhost
$ astro workspace user add Testing
Your Astro CLI Version (0.19.0) is ahead of the server version (0.15.1).
Consider downgrading your Astro CLI to match. See https://www.astronomer.io/docs/cli-quickstart for more information.
Error: Invalid email address. Please check the address and try again
```

Note the Houston Logs: 
```
2020-08-25T18:03:38 ERROR Invalid email address. Please check the address and try again
```

## 📸 Screenshots

Screen Shot 2020-09-01 at 11 49 05 AM

## 📋 Checklist


src/lib/users/index.js
```diff
@@ -187,6 +188,24 @@ export async function validateInviteToken(inviteToken, email) {
...
 }
 
+/**
+ * Validate an Email address
+ * @param  {String} email
+ * @return {String} validated and formatted email address
+ */
+export function validateEmailAddress(email) {
+  // eslint-disable-next-line no-useless-escape
+  const pattern = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
```
andriisoldatenko, line 198: can you please add source of regular expression? does it support unicode?

andriisoldatenko: can you please move it to constants.js?

andriisoldatenko: something like https://github.com/astronomer/houston-api/blob/6e75c5c44e022a724aac1e1f182fb7a8936ed08d/src/lib/constants/index.js#L58

andriisoldatenko (changes requested):

src/lib/users/index.js
adam2k: Yeah, good idea.  I can move to constants.

I found this regex here: https://emailregex.com/

adam2k: I'm not sure about unicode.  Do you have a list of characters that I can use for test cases to validate?  We might need to modify this regex.

adam2k: I can use this as a starting point. https://en.wikipedia.org/wiki/List_of_Unicode_characters

adam2k: I found a pretty good example here.  https://jsfiddle.net/aossikine/qCLVH/3/

src/lib/constants/index.js
```diff
@@ -58,6 +58,8 @@ export const MEDIATYPE_DOCKER_MANIFEST_V2 =
...
...-\uA48F\uA490-\uA4CF\uA700-\uA71F\uA800-\uA82F\uA840-\uA87F\uAC00-\uD7AF\uF900-\uFAFF\-\.\d]+)((\.([a-zA-Z\u0080-\u00FF\u0100-\u017F\u0180-\u024F\u0250-\u02AF\u0300-\u036F\u0370-\u03FF\u0400-\u04FF\u0500-\u052F\u0530-\u058F\u0590-\u05FF\u0600-\u06FF\u0700-\u074F\u0750-\u077F\u0780-\u07BF\u07C0-\u07FF\u0900-\u097F\u0980-\u09FF\u0A00-\u0A7F\u0A80-\u0AFF\u0B00-\u0B7F\u0B80-\u0BFF\u0C00-\u0C7F\u0C80-\u0CFF\u0D00-\u0D7F\u0D80-\u0DFF\u0E00-\u0E7F\u0E80-\u0EFF\u0F00-\u0FFF\u1000-\u109F\u10A0-\u10FF\u1100-\u11FF\u1200-\u137F\u1380-\u139F\u13A0-\u13FF\u1400-\u167F\u1680-\u169F\u16A0-\u16FF\u1700-\u171F\u1720-\u173F\u1740-\u175F\u1760-\u177F\u1780-\u17FF\u1800-\u18AF\u1900-\u194F\u1950-\u197F\u1980-\u19DF\u19E0-\u19FF\u1A00-\u1A1F\u1B00-\u1B7F\u1D00-\u1D7F\u1D80-\u1DBF\u1DC0-\u1DFF\u1E00-\u1EFF\u1F00-\u1FFF\u20D0-\u20FF\u2100-\u214F\u2C00-\u2C5F\u2C60-\u2C7F\u2C80-\u2CFF\u2D00-\u2D2F\u2D30-\u2D7F\u2D80-\u2DDF\u2F00-\u2FDF\u2FF0-\u2FFF\u3040-\u309F\u30A0-\u30FF\u3100-\u312F\u3130-\u318F\u3190-\u319F\u31C0-\u31EF\u31F0-\u31FF\u3200-\u32FF\u3300-\u33FF\u3400-\u4DBF\u4DC0-\u4DFF\u4E00-\u9FFF\uA000-\uA48F\uA490-\uA4CF\uA700-\uA71F\uA800-\uA82F\uA840-\uA87F\uAC00-\uD7AF\uF900-\uFAFF]){2,63})+)$/i;
```
adam2k, line 62: We could either go for something involved like this or keep the validation very basic at first?

adam2k:
> Looks good to me, assuming the regex is legit

I'm debating if we should go with this approach for unicode support or do something more basic, but may allow for some invalid use cases.

vishwas-astro:
> > Looks good to me, assuming the regex is legit
> 
> I'm debating if we should go with this approach for unicode support or do something more basic, but may allow for some invalid use cases.

My preference is to go with something simple that we can test.

src/lib/constants/index.js
andriisoldatenko: @adam2k wha't the source of RegExpression?

src/lib/constants/index.js
```diff
@@ -58,6 +58,7 @@ export const MEDIATYPE_DOCKER_MANIFEST_V2 =
 export const RELEASE_NAME_PATTERN = /^(?![0-9]+$)(?!.*-$)(?!-)[a-zA-Z0-9-]{1,63}$/;
 export const RELEASE_NAME_AIRFLOW_PATTERN = /^\/((?![0-9]+$)(?!.*-$)(?!-)[a-zA-Z0-9-]{1,63})\/(airflow|flower)/;
 export const VALID_DOCKER_IMAGE_NAME = /^(?<releaseName>(?![0-9]+$)(?!.*-$)(?!-)[a-zA-Z0-9-]{1,63})\/airflow$/;
+export const EMAIL_PATTERN = /^\S+@\S{4,}$/i;
```
adam2k, line 61: @samblackk @ryanahamilton FYI for email validation.  I don't think this should impact Astro UI, but wanted to point this out just in case.

ryanahamilton: Thanks for the heads-up. I don't believe we're validating any patterns on the client beyond the browser-enforced patterns for email input types, so all should be good.

adam2k: Thanks!  I posted a screenshot in the PR comments, so I think that error toast should work.  Just giving a heads up if you want to make any changes.

src/lib/constants/index.js
adam2k: I changed it to something more simple, but this outdated regex came from this jsFiddle: 

> I found a pretty good example here. https://jsfiddle.net/aossikine/qCLVH/3/

andriisoldatenko:
@adam2k we need to resolve conflicts

adam2k:
> @adam2k we need to resolve conflicts

@andriisoldatenko resolved all conflicts.

adam2k:
I'm switching this back to draft, so this does **not** get merged until after the Primsa 2 deployment goes out.

src/lib/errors/errors.js
```diff
@@ -68,6 +68,12 @@ export class EmailNotConfirmedError extends ApolloError {
   }
 }
 
+export class InvalidEmailError extends Error {
+  constructor() {
+    super("Invalid email address. Please check the address and try again");
```
andriisoldatenko, line 73:
```suggestion
    super("Invalid email address. Please use valid email address");
```
wdyt ?
cc @paolaperaza

andriisoldatenko (approved):

PR #454, opened 2020-09-14
code cleanup

andriisoldatenko:
## Description

- [x] Fix hanging during post-install hook upgrade-deployments.
- [x] code cleanup


schnie (approved):

src/scripts/upgrade-deployments/index.js
```diff
@@ -68,7 +68,10 @@ async function upgradeDeployments() {
   log.info(`Found ${deployments.length} deployments to upgrade.`);
 
   // Return early if we have no deployments.
-  if (deployments.length === 0) return;
+  if (deployments.length === 0) {
+    await prisma.disconnect();
```
adam2k, line 72: You don't have to change this here, but I think we should start using `$disconnect` because `disconnect` is deprecated. https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-client/connection-management#overview

adam2k (approved):
Looks good to me!

PR #459, opened 2020-09-18
Fix: Return emails when fetching User from DB

ryanahamilton:
## Description

In Prisma2, the `emails` field is not part of the default selection set. This `include` explicitly tells it to return `emails` when querying Users.

## 🎟 Issue(s)

Resolves un-filed issue identified by @kushalmalani:

> Unable to see email address when selecting a workspace user on dev.  Works fine on staging though. This is causing for updateUser call to fail with the following in console
>
> `Uncaught TypeError: user.emails is null`
> image

## 🧪 Functional Testing

Open to suggestions on how we might extend our tests to better protect against this?

## 📸 Screenshots

Response now includes values for `emails` field and is displayed in the UI:

Image 2020-09-18 at 10 20 12 AM

## 📋 Checklist

- [x] Updated any related [documentation](https://github.com/astronomer/docs/)


src/resolvers/query/users/index.js
```diff
@@ -12,5 +12,8 @@ export default async function users(parent, args, ctx) {
   const query = userQuery(args);
 
   // Run final query
-  return ctx.prisma.user.findMany({ where: { ...query } });
+  return ctx.prisma.user.findMany({
```
adam2k, line 15: Is this paginated or will this return every user?

src/resolvers/query/users/index.js
```diff
@@ -12,5 +12,8 @@ export default async function users(parent, args, ctx) {
...
+    where: { ...query },
+    include: { emails: true }
```
adam2k, line 17: Is the `email` the only field we need here or should we include other values like `fullName`, `profile`, etc?

adam2k (approved):
This looks good to me.  I have a couple questions about the query that I left, but they might not need to be addressed right away.

kushalmalani (approved):

src/resolvers/query/users/index.js
andriisoldatenko: AFAIK it will return all fields except related fields.

andriisoldatenko (approved):

PR #462, opened 2020-09-23
fix potential memory leaks due to high connection usage

andriisoldatenko:
## Description

## 🎟 Issue(s)

Resolves astronomer/issues#1841

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist


adam2k (approved):
🎉

vishwas-astro:
@andriisoldatenko do we still need a disconnect somewhere?

adam2k (changes requested):
I think we'll need to remove all of the `await prisma.disconnect();` statements.

andriisoldatenko:
> I think we'll need to remove all of the `await prisma.disconnect();` statements.

i did, can you see any missing parts?
```
pt "prisma.disconnect()" | grep -v workers | grep -v scripts | wc -l
       0
```

adam2k:
> > I think we'll need to remove all of the `await prisma.disconnect();` statements.
> 
> @adam2k
> 
> i did, can you see any missing parts?
> 
> ```
> pt "prisma.disconnect()" | grep -v workers | grep -v scripts | wc -l
>        0
> ```

They're in the test files.  Since we're making prisma a singleton we need to manage it differently in the jest tests.  If you check `setup-test.js` we can probably disconnect prisma there after every test to prevent those open handler issues in some of the tests.  It's the same thing that was happening with NATS.

src/lib/prismaclient/index.js
```diff
@@ -0,0 +1,4 @@
+import { PrismaClient } from "@prisma/client";
+
+const prisma = new PrismaClient();
+export default prisma;
```
sjmiller609, line 4: Is this executed each time it's imported, i.e. this object is unique for each file, like similar to existing at the top of the file?

This is lacking the disconnection logic wrapped in a try/finally, which is necessary to ensure it's cleaned up according to the [prisma documentation](https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-client/connection-management):

> When deploying your application as a long-running process or when running a Node.js script that uses Prisma Client, it's recommended to always explicitly call prisma.$disconnect() in your code even when an exception is thrown as shown below

```
main()
  .catch(e => {
    throw e
  })
  .finally(async () => {
    await prisma.$disconnect()
  })
```

sjmiller609 (changes requested):
I hope we can meet the following criteria for this change, I think most of which are accomplished in this PR already:

- [ ] all cases of "new PrismaClient" are wrapped in try / finally to close the connection
- [x] all cases of "new PrismaClient" occur at application start up
- [x] all clients are closed in default behavior when the applications stops gracefully (plug pull or other hard kill will unavoidably rely on DB to timeout the connection and that is acceptable)

src/lib/prismaclient/index.js
adam2k: @sjmiller609 this connection is a singleton, so it shouldn't close until we shutdown the app.

adam2k: Wait...this is interesting.  It also looks like we're not actually connecting ever either.  I'll do some functional testing and see what we need to do here.

adam2k: @sjmiller609 I'm going to remove all the disconnect calls see https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-client/connection-management#overview

`Unless you want to employ a specific optimization, calling prisma.$connect() is not necessary thanks to the lazy connect behavior:`

Adding all those connect calls is probably what initially caused some of these problems.

adam2k: I'll look into adding `$disconnect` to the other areas as well.

adam2k:
@andriisoldatenko after looking into this more I'm not sure we should be using `new PrismaClient()` as a singleton.  It seems like the issue might either be that we're not setting a `connection_limit` OR we're not disconnecting in the `finally` block in all cases where we are using Prisma.  We might be able to revert these changes or create a new branch with the try, catch, finally logic in place and see if that resolves the connection memory spiking issue.  https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-client/connection-management#overview

andriisoldatenko:
> https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-client/connection-management#overview

@adam2k @sjmiller609 @schnie 
we are setting `connection_limit ` but not explicitly:

>If theconnection_limit argument is omitted, the default number of connections is calculated according to this formula: num_physical_cpus * 2 + 1 where num_physical_cpus represents the number of physical CPUs on your machine.

no need to disconnect if we are using/sharing 1 instance of prisma client, so we have 1 connection pool per pod, which works 1 to 1 to prisma1. I prefer to keep this change just to have more clean code.

so try/finally doesn't make any sense if we are starting connection pool during start pod and terminating on ctrl+c or sigterm.

src/lib/prismaclient/index.js
andriisoldatenko: correct @sjmiller609  we don't need 
>all cases of "new PrismaClient" are wrapped in try / finally to close the connection

adam2k:
The only other question is if the PrismaClient does error while the node.js app is running will it reconnect with this implementation?

src/lib/prismaclient/index.js
```diff
@@ -0,0 +1,4 @@
+import { PrismaClient } from "@prisma/client";
+
+const prisma = new PrismaClient();
```
adam2k, line 3: I'm going to look into this part to make sure that it's not still making multiple connections.  We might need to wrap in a class definition like we did in `publishers`

andriisoldatenko:
> The only other question is if the PrismaClient does error while the node.js app is running will it reconnect with this implementation?
@adam2k 

i don't want to make it complicated.
Prisma1 never reconnects.

src/lib/prismaclient/index.js
andriisoldatenko: @adam2k this like doesn't create multiply connections it creates connection pool.

adam2k: @andriisoldatenko I'm going to do the same removal of `$disconnect` on the workers here too.  https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-client/deployment#recommended-connection-limit

andriisoldatenko: sure.

src/lib/prismaclient/index.js
sjmiller609: why not call disconnect?

adam2k (approved):
This looks good to me! 🎉   I think we should test in dev and staging.

src/lib/prismaclient/index.js
sjmiller609: I'm not talking about literally try /finally right here. I'm talking about ensuring the 'finally' style logic is used, and the finally part should execute during application shutdown.

sjmiller609:
I think it's missing the disconnect logic

src/lib/prismaclient/index.js
andriisoldatenko: @sjmiller609 can you please be more concrete?

andriisoldatenko:
> I think it's missing the disconnect logic

it has disconnect logic where it make sense.
Example:
- https://github.com/astronomer/houston-api/commit/f153e123b6f9b65e4e15a6abe60910f553bb995d
- https://github.com/astronomer/houston-api/commit/888b6faa644d888e72cbd7837d9cd49b54c65a27
- https://github.com/astronomer/houston-api/commit/56ced28ef3d36223a63bc1ad0c4abcdf049db0c5

src/lib/prismaclient/index.js
sjmiller609: I'm talking about interrupt or crash tolerant cleanup logic to always run.

>  it's recommended to always explicitly call prisma.$disconnect() in your code ... as shown below

```
main()
  .catch(e => {
    throw e
  })
  .finally(async () => {
    await prisma.$disconnect()
  })
```

Of course, we can't just add the finally wrapping exactly how this is done now. Perhaps the finally line lines somewhere in this file? https://github.com/astronomer/houston-api/blob/main/src/index.js

adam2k: If we can access the `exit` hook we might be able to accomplish something like 

```process.on("exit", async () => {
  log.info("About to exit from the app!.");
  await prisma.$disconnect();
  log.info("Disconnected prisma.");
});
```

But so far I can't make this method `async`.

src/lib/prisma-client/index.js
```diff
@@ -0,0 +1,14 @@
...
+class Prisma {
+  constructor() {
+    if (!Prisma.instance) {
+      Prisma.instance = new PrismaClient();
+    }
+    return Prisma.instance;
+  }
+}
+
+const prisma = new Prisma();
```
sjmiller609, line 12: I'm talking about this line

andriisoldatenko: replied in https://github.com/astronomer/houston-api/pull/462#issuecomment-698391219

andriisoldatenko: replied in https://github.com/astronomer/houston-api/pull/462#issuecomment-698391219

sjmiller609: the documentation says

> it's recommended to **always** explicitly call prisma.$disconnect() in your code

src/lib/prismaclient/index.js
adam2k: Hmm... it looks like this may work?

```
2020-09-24T10:48:08 INFO About to exit from the app!.
2020-09-24T10:48:08 INFO Disconnected prisma.
```

src/lib/prisma-client/index.js
sjmiller609: something like in the file that starts the server,

```
import prisma
...
try:
  start server
finally:
  prisma.disconnect
```

sjmiller609 (approved):

src/lib/prisma-client/index.js
andriisoldatenko: @sjmiller609 confirmed to implement it in https://github.com/astronomer/issues/issues/1845

PR #482, opened 2020-10-01
1812 user invites roles

kushalmalani:
## Description

Dropping the InviteToken table and utilize the User and RoleBindings table to manage invitations instead

## 🎟 Issue(s)

Resolves astronomer/issues#1812

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist


src/lib/users/index.js
```diff
@@ -165,14 +145,14 @@ export async function validateInviteToken(prisma, token, email) {
   if (!token) return [];
 
   // Grab the invite token.
-  const inviteToken = await prisma.inviteToken.findOne({
-    where: { token }
+  const inviteToken = await prisma.user.findMany({
+    where: { username: email, status: "pending", token }
```
andriisoldatenko, line 149: can we use constant instead of string: "pending"?

src/migrations/20200921231650_copy_invites_to_users.js
```diff
@@ -0,0 +1,43 @@
+import log from "logger";
+import cuid from "cuid";
+
+const schema = "houston$default";
```
andriisoldatenko, line 4: please use 
```suggestion
const schema = config.get("database.schema");
```

andriisoldatenko (changes requested):

package.json
```diff
@@ -33,6 +33,7 @@
     "commander": "github:astronomer/commander",
     "config": "^2.0.1",
     "cookie-parser": "^1.4.3",
+    "cuid": "^2.1.8",
```
andriisoldatenko, line 36: @adam2k i think we can reuse this lib in your PR ^^.

adam2k: @andriisoldatenko I'm wondering if we should use this lib or roll our own?  Right now we're only using two functions (one in this PR `cuid()` and then in my PR I'm using `isCuid()`).  What do you think?  I'm thinking less potential snyk PRs to deal with if this lib ends up with any issues.

adam2k: nvm...this library is pretty lightweight.  I can add this into my PR too.

andriisoldatenko:
Please correct if i’m wrong just trying to understand the whole problem:

because we have https://github.com/astronomer/houston-api/blob/a9b3d9766e7d66edfc3ab74f580a0815e3279d25/prisma/schema.prisma#L103 `InviteToken` only with workspace we can’t invite users to deployments ONLY?

so instead of adding deploymentId to InviteToken we decided to add token column to User model? (edited) 

i mean similar to RoleBinding :

```model RoleBinding {
  id             String          @default(cuid()) @id
  role           Role?
  user           User?            @relation(fields: [userId], name: "RoleBindingToUser", references: [id])
  userId         String?
  serviceAccount ServiceAccount? @relation(fields: [serviceAccountId], name: "RoleBindingToServiceAccount", references: [id])
  serviceAccountId String?
  workspace      Workspace?      @relation(fields: [workspaceId], name: "RoleBindingToWorkspace", references: [id])
  workspaceId    String?
  deployment     Deployment?     @relation(fields: [deploymentId], name: "DeploymentRoleBindings", references: [id])
  deploymentId   String?
  createdAt      DateTime        @default(now())
  updatedAt      DateTime        @updatedAt
}
```
we can have workspace/or deployment reference.
cc @schnie @adam2k @ryanahamilton

src/lib/users/index.js
```diff
@@ -165,14 +145,14 @@ export async function validateInviteToken(prisma, token, email) {
...
   });
 
   // Throw error if token not found.
-  if (!inviteToken) throw new InviteTokenNotFoundError();
+  if (inviteToken.length === 0) throw new InviteTokenNotFoundError();
```
adam2k, line 153: is it possible for inviteToken to be null?  I'd recommend

```Suggestion
  if (!inviteToken || inviteToken.length === 0) throw new InviteTokenNotFoundError();
```

src/migrations/20200921231650_copy_invites_to_users.js
```diff
@@ -0,0 +1,43 @@
...
+
+  await knex.transaction(async function(trx) {
+    log.debug(`Getting records from InviteToken`);
+    inviteTokens = await trx
+      .withSchema(schema)
+      .select("*")
+      .from("InviteToken");
+  });
+
+  await knex.transaction(async function(trx) {
```
adam2k, line 17: Do we need a knex transaction or can we do this with a Prisma migration instead?  It'll be easier to manage if we can use Prisma only unless it won't work in this situation.

src/resolvers/mutation/invite-user/index.js
```diff
@@ -2,47 +2,45 @@ import { DuplicateEmailError, UserInviteExistsError } from "errors";
...
-    { where: { address: email } }
-    // `{ user { id } }`
-  );
+  const emailRow = await ctx.prisma.email.findMany({
+    where: { address: email, user: { status: "active" } },
+    include: { user: true }
+  });
 
-  if (emailRow) {
+  if (emailRow.length > 0) {
```
adam2k, line 18: Is it possible for emailRow to be null?  If so:

```Suggestion
  if (size(emailRow) > 0) {
```

src/resolvers/mutation/workspace-add-user/index.js
```diff
@@ -20,15 +20,15 @@ export default async function workspaceAddUser(_, args, ctx) {
...
+  const emailRow = await ctx.prisma.email.findMany({
+    where: { address: email, user: { status: "active" } },
     include: { user: true }
   });
 
   if (!role.startsWith(`${ENTITY_WORKSPACE}_`))
     throw new UserInputError("invalid workspace role");
 
-  const user = emailRow ? emailRow.user : null;
+  const user = emailRow.length > 0 ? emailRow[0].user : null;
```
adam2k, line 31: This could be:

```Suggestion
  const user = get(emailRow, "[0].user", null);
```

src/resolvers/mutation/workspace-add-user/index.js
```diff
@@ -20,15 +20,15 @@ export default async function workspaceAddUser(_, args, ctx) {
   const email = validateEmailAddress(emailArg);
```
adam2k, line 20: I'm noticing that there are some `email.toLowerCase()` and other references that aren't.  I'd suggest formatting the email to lowercase somewhere up here to prevent downstream bugs with string comparisons.

src/resolvers/mutation/workspace-update-user-role/index.js
```diff
@@ -16,15 +16,11 @@ export default async function workspaceUpdateUserRole(parent, args, ctx) {
...
-  // This function uses roleBindings/inviteTokens query because even though we
-  // expect the value to be unique Prismas doesn't support multi-column uniques
-  // yet :(
-
   // Check for user by incoming email arg.
   const roleBindings = await ctx.prisma.roleBinding.findMany({
     where: {
       workspace: { id: workspaceUuid },
-      user: { emails: { some: { address: email } } }
+      user: { username: email }
```
adam2k, line 23: This almost reads like a mistake, even though it's not.  I'd consider changing line 14 to something like

`const { email: username, workspaceUuid, role } = args;`

and then this line

```Suggestion
      user: { username }
```

src/resolvers/query/invites/index.js
```diff
@@ -12,5 +12,24 @@ export default async function invites(parent, args, ctx) {
   const query = inviteQuery(args);
 
   // Run final query
-  return await ctx.prisma.inviteToken.findMany({ where: { ...query } });
+  const invites = await ctx.prisma.roleBinding.findMany({
+    where: {
+      user: { status: "pending", ...query }
```
adam2k, line 17: @andriisoldatenko mentioned this in an earlier comment, but can "pending" be a const?

src/resolvers/types/workspace/index.js
```diff
@@ -19,9 +19,27 @@ export function deployments(parent) {
...
+  // Conforming to type Invite to main API response format
+  return invites.map(invite => {
+    return {
+      id: invite.user.id,
+      role: invite.role,
+      email: invite.user.username,
+      token: invite.user.token,
+      createdAt: invite.user.createdAt,
+      updatedAt: invite.user.updatedAt
+    };
```
adam2k, line 42: To avoid repeating can this be converted into a lib function and used here and `src/resolvers/query/workspace-invites/index.js`?

It seems like eventually one of this blocks could be updated, but not the other.

adam2k (commented):
This is starting to look pretty good.  I have a couple of nits and some clean up comments.  I think if some of the consolidation goes in to make the code more DRY it'll make it easier to spot any potential bugs.

kushalmalani:
> Please correct if i’m wrong just trying to understand the whole problem:
> 
> because we have
> 
> https://github.com/astronomer/houston-api/blob/a9b3d9766e7d66edfc3ab74f580a0815e3279d25/prisma/schema.prisma#L103
> `InviteToken` only with workspace we can’t invite users to deployments ONLY?
> 
> so instead of adding deploymentId to InviteToken we decided to add token column to User model? (edited)
> 
> i mean similar to RoleBinding :
> 
> ```
>   id             String          @default(cuid()) @id
>   role           Role?
>   user           User?            @relation(fields: [userId], name: "RoleBindingToUser", references: [id])
>   userId         String?
>   serviceAccount ServiceAccount? @relation(fields: [serviceAccountId], name: "RoleBindingToServiceAccount", references: [id])
>   serviceAccountId String?
>   workspace      Workspace?      @relation(fields: [workspaceId], name: "RoleBindingToWorkspace", references: [id])
>   workspaceId    String?
>   deployment     Deployment?     @relation(fields: [deploymentId], name: "DeploymentRoleBindings", references: [id])
>   deploymentId   String?
>   createdAt      DateTime        @default(now())
>   updatedAt      DateTime        @updatedAt
> }
> ```
> 
> we can have workspace/or deployment reference.
> cc @schnie @adam2k @ryanahamilton

@andriisoldatenko Yes, thats accurate. The solution to create/delete many records in InviteTokens table, everytime we add a user to workspace seems inefficient and there is a redundancy in data we were storied which would have been optimized using User and RoleBindings table

src/lib/users/index.js
kushalmalani: It would be not, since findMany returns a list....so it would be an empty array at best. But it would hurt to use isEmpty, just to be on the safe side. Will update.

src/migrations/20200921231650_copy_invites_to_users.js
kushalmalani: Dont we use prisma only to manage our schema? We use knex for migration of data in 0.20, so i wanted to maintain same behavior.

@andriisoldatenko Thoughts?

src/resolvers/mutation/invite-user/index.js
kushalmalani: Again, no...but we can add it to be on safer side

src/resolvers/mutation/workspace-add-user/index.js
kushalmalani: validateEmailAddress already does that. I forgot to remove the lowercase references after merging with main

src/resolvers/types/workspace/index.js
kushalmalani: Yeah, that makes sense. Updating that.

prisma/migrations/migrate.lock
```diff
@@ -1,4 +1,5 @@
 # Prisma Migrate lockfile v1
 
 20200908231430-initial
-20200923194140-add-deployment-revision
\ No newline at end of file
+20200922172359-add-new-columns-user-table
```
andriisoldatenko, line 4: wrong order cc @adam2k

src/migrations/20200921231650_copy_invites_to_users.js
schnie: My understanding was that the prisma migrations only do schema related changes. The reason we added `knex` migrations back in the day, was to give us a way to move actual data in migrations. Can prisma migrations do that now?

But within the migrations themselves, we might be able to use the prima client instead of the raw SQL. What do you all think?

andriisoldatenko: correct, i'm not sure that you can do it using prisma migrations.
@adam2k AFAIK prisma migrate supports only data modeling! https://www.prisma.io/docs/understand-prisma/data-modeling

adam2k: Ah okay, I thought the Prisma migrations could handle more use cases.  Using `knex` works for me then.  We can revisit when Prisma migrate comes out of the experimental phase.

schnie: Should we use knex and raw SQL here or use the prisma client like we do in the rest of the app?

Standardizing seems like a good thing, but I could be missing something. What do y'all think? Could be nitpicking

adam2k (approved):
Looks good to me! 🎉 🚀 🥳

src/migrations/20200921231650_copy_invites_to_users.js
schnie: Should we use knex and raw SQL here or use the prisma client like we do in the rest of the app?

Standardizing seems like a good thing, but I could be missing something. What do y'all think? Could be nitpicking

src/generated/schema.graphql
```diff
@@ -317,7 +308,7 @@ type Mutation {
   createWorkspaceServiceAccount(category: String, label: String!, role: Role!, workspaceUuid: Uuid!): ServiceAccount
   deleteDeployment(deploymentUuid: Uuid!): Deployment
   deleteDeploymentServiceAccount(deploymentUuid: Uuid!, serviceAccountUuid: Uuid!): ServiceAccount
-  deleteInviteToken(inviteUuid: Uuid): Invite
+  deleteInviteToken(inviteUuid: Uuid, workspaceUuid: Uuid): Invite
```
andriisoldatenko, line 311: do we need to update astro-ui?
https://github.com/astronomer/astro-ui/blob/148732c556abf6384f2f8992610d8b49f20e137a/src/api/mutate/workspaceInviteDelete.js#L8
cc @samblackk @ryanahamilton

andriisoldatenko: nvm it fixed in https://github.com/astronomer/astro-ui/blob/f6d338e148273a0b94ea4d3cad2af10d622e0372/src/api/mutate/workspaceInviteDelete.js#L9

andriisoldatenko (approved):

andriisoldatenko:
practically LGTM.

src/migrations/20200921231650_copy_invites_to_users.js
kushalmalani: @schnie I think its worth exploring. Maybe we can file a separate ticket and look into it and not block this going into release.

src/lib/invites/index.js
```diff
@@ -18,3 +18,21 @@ export function inviteQuery(args) {
 
   return null;
 }
+
+/*
+ * Conforming to type Invite to maintain API response format.
+ * @param {Object} list of all invites.
+ * @return {Object} list of conformed response matching Invite type.
+ */
+export function inviteResponseConformation(invites) {
```
schnie, line 27:
```suggestion
export function inviteResponseConfirmation(invites) {
```

schnie (approved):

PR #490, opened 2020-10-06
No role binding creation when a user source is SYSTEM

kushalmalani:
## Description

No role binding creation when a user source is SYSTEM

## 🎟 Issue(s)

Resolves astronomer/issues#1812

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist


adam2k (approved):
This looks good to me! 👍

src/migrations/20200921231650_copy_invites_to_users.js
```diff
@@ -27,10 +27,17 @@ export async function up(knex) {
         ON CONFLICT ( username ) DO UPDATE SET token = '${invite.token}', source = '${invite.source}', "updatedAt" = CURRENT_TIMESTAMP RETURNING id`);
 
         // Creating respective RoleBindings
-        await trx.raw(`INSERT INTO ${schema}."RoleBinding" (id, role, "userId", "workspaceId", "updatedAt")
-        VALUES ( '${cuid()}', '${invite.role}', '${userId.rows[0].id}', '${
-          invite.workspaceId
-        }', CURRENT_TIMESTAMP)`);
+        if (invite.source === "SYSTEM") {
+          await trx.raw(`INSERT INTO ${schema}."RoleBinding" (id, role, "userId", "workspaceId", "updatedAt")
+          VALUES ( '${cuid()}', 'SYSTEM_VIEWER', '${userId.rows[0].id}', '${
```
andriisoldatenko, line 32: looks good, but not sure about default value.
cc @schnie

schnie: What are the options here for `invite.source`? I don't know if they would actually need a RoleBinding in this case. They'd just be a user with no workspaces and no deployments right?

andriisoldatenko: problem in prisma2 we introduces UDT https://github.com/astronomer/astronomer/blob/master/bin/migration-scripts/prisma1_to_prisma2.sql#L15 so we need some default value.

andriisoldatenko: and if we run:
```
postgres=# CREATE TYPE "houston$default"."Role" AS ENUM ('SYSTEM_ADMIN', 'SYSTEM_EDITOR', 'SYSTEM_VIEWER', 'WORKSPACE_ADMIN', 'WORKSPACE_EDITOR', 'WORKSPACE_VIEWER', 'DEPLOYMENT_ADMIN', 'DEPLOYMENT_EDITOR', 'DEPLOYMENT_VIEWER', 'USER', NULL);
ERROR:  syntax error at or near "NULL"
LINE 1: ...N', 'DEPLOYMENT_EDITOR', 'DEPLOYMENT_VIEWER', 'USER', NULL);

```

andriisoldatenko: but we can introduce `'NULL'` but i prefer to default some default ROLE or extend ENUM list with "DEFAULT" or something rather than using empty string or `NULL`.

kushalmalani: We maybe dont need roleBindings at all. We can just create the user record

src/migrations/20200921231650_copy_invites_to_users.js
```diff
@@ -27,10 +27,12 @@ export async function up(knex) {
         ON CONFLICT ( username ) DO UPDATE SET token = '${invite.token}', source = '${invite.source}', "updatedAt" = CURRENT_TIMESTAMP RETURNING id`);
 
         // Creating respective RoleBindings
-        await trx.raw(`INSERT INTO ${schema}."RoleBinding" (id, role, "userId", "workspaceId", "updatedAt")
-        VALUES ( '${cuid()}', '${invite.role}', '${userId.rows[0].id}', '${
-          invite.workspaceId
-        }', CURRENT_TIMESTAMP)`);
+        if (invite.source !== "SYSTEM") {
```
adam2k, line 30: Can we confirm that all other roles will require roleBinding?  Maybe I missed it in slack if you already discussed it.

adam2k (approved):
This looks good to me.  It was confirmed that the roleBinding isn't required until a user is added to a workspace.

src/migrations/20200921231650_copy_invites_to_users.js
kushalmalani: Yes, there are just 2 sources...SYSTEM and WORKSPACE....Only SYSTEM source wouldnt need a role assigned to it.

andriisoldatenko (approved):

PR #493, opened 2020-10-07
Fix location of deployment status getter

adam2k:
## Description

Move `getDeploymentStatus` inside of the poller implementation so the status can update correctly.

## 🎟 Issue(s)

Resolves astronomer/issues#1400

## 🧪 Functional Testing

1. Port forward prometheus from dev or staging to your local environment.
2. You can reconfigure Houston API to poll at a faster interval to see the status changes happen sooner.
3. Create a deployment in dev or staging and make a deployment with the same name in your environment.
4. Create an update to your local deployment (you should see the deployment go into `Deploying` mode after a little while)
5. Create an update to your remote deployment with the same name.
6. Once the revisions match for your local and the remote env then the status will move to a healthy state.
7. You can either run the GraphQL subscription or use Astro UI to see the results in the Websocket connection

## 📸 Screenshots

## 📋 Checklist


src/resolvers/subscription/deployment-status/index.unit.test.js
```diff
@@ -0,0 +1,64 @@
...
+    };
+
+    // Run the graphql subscription.
+    const iterator = await subscribe(null, args, prisma, pubsub);
+    const iteratorNext = await iterator.next();
+    const deploymentStatus = await iteratorNext.value.deploymentStatus;
+    const status = await deploymentStatus.status;
+
+    expect(iterator.errors).toBeUndefined();
+    expect(status).toEqual(DEPLOYMENT_HEALTHY);
```
andriisoldatenko, line 31: @adam2k can you please add:
```
expect(prisma)..toHaveBeenCalledWith({whatever you need});
```

adam2k: oh, that's a good point.  I think I can remove prisma here completely.  We're not actually calling prisma in either of these tests.

adam2k: Let me know if that looks good.  I've removed Prisma completely because it's not necessary in these tests.

andriisoldatenko (approved):

PR #495, opened 2020-10-09
fix: the label values must be strings.

andriisoldatenko:
## Description

The label values must be strings.

before this PR:
```
labels:
    platform: deadly-terrier
    workspace: ckg26kflz0125pvxt043gibfn
    revision: 1
```
after:
```
labels:
    platform: deadly-terrier
    workspace: ckg26kflz0125pvxt043gibfn
    revision: '1'
```

logs from commander:
```
INFO[0473] CHART PATH: /Users/andrii/Library/Caches/helm/repository/airflow-0.17.0.tgz  function=InstallRelease package=helm
ERRO[0474] unable to build kubernetes objects from release manifest: [unable to decode "": resource.metadataOnlyObject.ObjectMeta: v1.ObjectMeta.Name: Labels: ReadString: expects " or n, but found 1, error found in #10 byte of ...|evision":1,"tier":"a|..., bigger context ...|rier","release":"boreal-kinetic-5807","revision":1,"tier":"airflow","workspace":"ckg26kflz0125pvxt04|..., unable to decode "": resource.metadataOnlyObject.ObjectMeta: v1.ObjectMeta.Name: Labels: ReadString: expects " or n, but found 1, error found in #10 byte of ...|evision":1,"tier":"r|..., bigger context ...|rier","release":"boreal-kinetic-5807","revision":1,"tier":"resources","workspace":"ckg26kflz0125pvxt|..., unable to decode "": resource.metadataOnlyObject.ObjectMeta: v1.ObjectMeta.Name: Labels: ReadString: expects " or n, but found 1, error found in #10 byte of ...|evision":1,"workspac|..., bigger context ...|rier","release":"boreal-kinetic-5807","revision":1,"workspace":"ckg26kflz0125pvxt043gibfn"},"name":"|...]  function=InstallDeployment package=kubernetes
I
```

and after:
```
no errors.
```


src/lib/deployments/config/index.js
```diff
@@ -379,7 +379,8 @@ export function platform(deployment) {
   const labels = {
     platform: platformReleaseName,
     workspace: workspace.id,
-    revision
+    // Helm note: The label values must be strings. In yaml, that means all numeric values must be quoted.
+    revision: revision.toString()
```
adam2k, line 383: Is it possible for revision to be null?  This might crash.

adam2k, line 383:
```Suggestion
 const formattedRevision = revision && revision.toString();
 const labels = {
    platform: platformReleaseName,
    workspace: workspace.id,
    revision: formattedRevision
```

adam2k:
@andriisoldatenko how did you confirm the error?  Was it locally with commander (`make local`) and kind or did you have to run the entire Astronomer platform locally?

src/lib/deployments/config/index.js
adam2k: Note that revision is undefined in some of our tests too:

```
  ● platform › returns platform for less than v0.11.0

    TypeError: Cannot read property 'toString' of undefined
```

Maybe we should completely remove the `revision` key in that situation?

andriisoldatenko:
@adam2k yep easy test:
- houston-api and commander locally with real kind cluster 100% reproducible.

src/lib/deployments/config/index.js
andriisoldatenko: fixed tests. rather than adding `revision && revision.toString();`

src/generated/schema.graphql
```diff
@@ -103,40 +103,103 @@ type DeployInfo {
   nextCli: String
 }
 
+"""
```
adam2k, line 106: Did you mean for these changes to go into this PR too?

src/lib/deployments/config/index.js
adam2k: That's probably OK since we always have revision.  Does this bug only impact labels or any other helm values that we need to change as well?

src/generated/schema.graphql
andriisoldatenko: @adam2k removed from this PR>

adam2k: Thanks!  Should this go into a new PR?  I can open one up with only the generate schema changes.

src/lib/deployments/config/index.js
```diff
@@ -379,7 +379,8 @@ export function platform(deployment) {
...
   };
 
   // Pre-0.11.0 format.
```
adam2k, line 386: @schnie is this something that is only relevant to our platform before `0.11.0`?  Is this something that we can remove now.

src/generated/schema.graphql
adam2k: nvm...You already made one.

src/lib/deployments/config/index.js
andriisoldatenko: only `revision`.

src/lib/deployments/config/index.js
```diff
@@ -388,8 +389,11 @@ export function platform(deployment) {
     workspace: workspace.id
   };
 
+  const formattedLabels = mapValues(labels, l => String(l));
+  const formattedPlatform = mapValues(platform, p => String(p));
+
```
adam2k, line 394: @andriisoldatenko let me know what you think about this.  I figured we could make sure every label is a string here.  Otherwise we can revert back to converting only revision to a string.  It's safer if we type cast like this:

`revision: String(revision)`

andriisoldatenko: we need to test this change, i prefer to keep bugfix small.

adam2k: I'll revert to String(revision)

PR #508, opened 2020-10-14
Fix/create token query

adam2k:
## Description

Fix the createToken query after the migration that allows for `localCredentialId` to be null.

## 🎟 Issue(s)

Resolves astronomer/issues#1933

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist


andriisoldatenko (approved):

vishwas-astro (approved):
Approving, provided tests pass. Please rebase from main too.

andriisoldatenko:
@adam2k added tests.

src/resolvers/mutation/create-token/index.js
```diff
@@ -18,36 +17,20 @@ import { USER_STATUS_ACTIVE, USER_STATUS_PENDING } from "constants";
...
-    // `{ id, status, localCredential { password } }`
-  );
-
-  // There should be one or none.
-  const user = first(users);
+  });
 
   // Throw error if user not found.
-  if (!user) throw new ResourceNotFoundError();
+  if (!user || user.status !== USER_STATUS_ACTIVE)
```
andriisoldatenko: @adam2k @kushalmalani i changed this line because it breaks current statuses for user. So let's keep it as it was.
we are checking status in https://github.com/astronomer/houston-api/pull/508/commits/75be674beb9824e5679677b8db2e080a49df59c1#diff-68452291a2f1f1552a31c39bddd72f3bef2b15dcdb5d706c77b13e0f1bb436e4R56

schnie (approved):

src/resolvers/mutation/create-token/index.js
adam2k: @andriisoldatenko I think that's linking to the same change.  You think we should remove `user.status !== USER_STATUS_ACTIVE`?  We added this is so that we could correctly throw the `ResourceNotFoundError` when a record is discovered, but the User status is still pending.  We can switch to Pending if you think that's more accurate for all cases?

andriisoldatenko (approved):

src/resolvers/mutation/create-token/index.js
adam2k: @andriisoldatenko and I talked and we created a check for `localCredential.password` in this commit: https://github.com/astronomer/houston-api/pull/508/commits/4ca17cb59d63ba7d5d029cf432d5a4faae2d11f1

PR #517, opened 2020-10-20
Upgrade Jest to fix Prototype Pollution errors from lodash dependency

adam2k:
## Description

Snyk bot has identified a `Prototype Pollution` bug in lodash `4.17.15`.

## 🎟 Issue(s)

Resolves astronomer/issues#1995

## 🧪 Functional Testing

Make sure the tests still run correctly locally and within the CLI

## 📸 Screenshots

## 📋 Checklist


andriisoldatenko (approved):

src/schema/permissions.unit.test.js
```diff
@@ -1,5 +1,5 @@
+import { schema } from "../schema";
```
adam2k, line 1: @andriisoldatenko this was causing some jest errors because this test file is in a directory called `schema`, but it's really looking for `../schema.js`.  It's not pretty, but this works for now.  We can probably fix this by updating the jest config and/or `.bablerc`

kushalmalani (approved):

adam2k:
The tests here are failing with an exit 1 status.  They aren't hanging up locally, but it has something to do with the `schema` references.  This might be worth updating after updating `prisma` and the `nexus-prisma-plugin` OR we could try to upgrade to Jest `25` instead of `26` and see if the same problems persist (we're currently using Jest 24).

adam2k:
I tested using `Jest 25.5.4` and the `schema` error happens there as well.  This is going to take further investigation.  I think we should upgrade `Prisma` first.

andriisoldatenko:
@adam2k i think we already upgraded prisma so can we retry again?

adam2k:
After rebasing with main this error still exists: 

```
Summary of all failing tests
 FAIL  src/schema/permissions.unit.test.js
  ● Test suite failed to run

    Cannot find module './' from 'src/schema/permissions.unit.test.js'

    However, Jest was able to find:
    	'../schema.js'

    You might want to include a file extension in your import, or update your 'moduleFileExtensions', which is currently ['js', 'json', 'jsx', 'ts', 'tsx', 'node'].

    See https://jestjs.io/docs/en/configuration#modulefileextensions-arraystring

      1 | import { permissions } from "schema/permissions";
    > 2 | import { schema } from "schema";
        | ^
      3 | import { applyMiddleware } from "graphql-middleware";
      4 | import { graphql } from "graphql";
      5 | import casual from "casual";

      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:306:11)
      at Object.<anonymous> (src/schema/permissions.unit.test.js:2:1)
```

adam2k:
@andriisoldatenko this is working for me locally now for `yarn start` and `yarn coverage`.  I'll need to see if this will pass through the CI.  The `schema.js` and `schema` directory were conflicting with each other while running the Jest tests.  This appears to solve the issue, but now means that the `generated` schema files are located under `schema/generated`.  I think it's OK, but let me know if you find any issues.

PR #524, opened 2020-10-21
Hotfix/deployment deleted at

adam2k:
## Description

Fix a bug where deployments will not load because of a broken query.  This is a limitation of the version of `Prisma` that we are using where `findOne` only allows for a single param and `findFirst` isn't available for us to use in this version.

## 🎟 Issue(s)

Resolves astronomer/issues#1986

## 🧪 Functional Testing

1. Login into Astro UI with Houston API running
2. Create a workspace or visit an existing workspace
3. View or create a deployment for the particular workspace
4. The deployment should appear as normal
5. Copy the url of the deployment
6. Delete the deployment
7. Visit the url and the deployment should not appear

## 📸 Screenshots

**Creating a Deployment**

Screen Shot 2020-10-21 at 2 43 28 PM

**Viewing and then deleting a Deployment**

Screen Shot 2020-10-21 at 2 43 19 PM

**Attempting to view deleted deployment**

Screen Shot 2020-10-21 at 2 43 41 PM

**GraphQL Query - Valid Deployment**
Screen Shot 2020-10-21 at 2 49 09 PM

**GraphQL Query - Deleted Deployment**

Screen Shot 2020-10-21 at 2 52 01 PM

**GraphQL Query - Invalid Deployment**

Screen Shot 2020-10-21 at 2 53 21 PM

## 📋 Checklist


src/resolvers/query/workspace-deployment/index.unit.test.js
```diff
@@ -1,52 +1,84 @@
-import { schema } from "schema";
+import workspaceDeployment from ".";
 import casual from "casual";
-import { graphql } from "graphql";
+import cuid from "cuid";
+import { AIRFLOW_EXECUTOR_CELERY } from "constants";
 
-// Define our mutation
-const query = `
```
andriisoldatenko, line 6: any reason to remove default mutation and graphql?

andriisoldatenko: usually i prefer to keep such. tests to check schema trough graphql, so if somebody change args or something this query typically fails.

adam2k: It wasn't testing the implementation of `workspaceDeployment` that's why this bug was able to exist.  The code was covered, but we are mocking `await ctx.prisma.deployment.findOne`, so that method was returning whatever is mocked.  We can test the logic of `workspaceDeployment` now that branching logic exists, but it's still not perfect.

The only way we can test this accurately is with an integration test that does some DB setup by creating a `deletedAt` deployment and a valid deployment.  Then we can use Prisma to retrieve both from the DB and confirm that the `deletedAt` entry throws and the valid deployment doesn't throw.  In lieu of that full integration test.  I made three tests to account for the three cases that we now have in this method.

andriisoldatenko (approved):

src/resolvers/query/workspace-deployment/index.unit.test.js
```diff
@@ -1,52 +1,84 @@
...
+    });
+  });
+
+  test("throws for a deleted deployment", async () => {
+    // Setup a deletedAt date to trigger an error
+    deployment.deletedAt = new Date();
+    const findOne = jest.fn().mockImplementationOnce(() => deployment);
+
+    // Construct db object for context.
+    const prisma = {
```
kushalmalani, line 62: This is being repeated in every test case. We can declare it once outside or part of beforeEach. Same for args.

src/resolvers/query/workspace-deployment/index.unit.test.js
adam2k: @andriisoldatenko I actually added it back in and instead of throwing we can return null.  We can get the integration tests in at a later time too.

src/resolvers/query/workspace-deployment/index.unit.test.js
adam2k: Updated and did the same with `const args`

samblackk (approved):

kushalmalani (approved):
LGTM

PR #513, opened 2020-10-19
deployment add user role mutation

kushalmalani:
## Description

> Describe the purpose of this pull request.
deployment add user role mutation

## 🎟 Issue(s)

Resolves astronomer/issues#1762

## 🧪 Functional Testing

> List the functional testing steps to confirm this feature or fix.

## 📸 Screenshots

> Add screenshots to illustrate the validity of these changes.

## 📋 Checklist

- [x] Rebased from the main (or release if patching) branch (before testing)
- [x] Ran `yarn lint && yarn test` before taking out of draft
- [x] Added/updated applicable tests
- [ ] Tested against [Astro-UI](https://github.com/astronomer/astro-ui/) and [Astro-CLI](https://github.com/astronomer/astro-cli/) clients.
- [ ] Communicated to/tagged owners of respective clients potentially impacted by these changes.
- [ ] Updated any related [documentation](https://github.com/astronomer/docs/)

PR #526, opened 2020-10-21
deployment remove user role

kushalmalani:
## Description

Deployment remove user role

## 🎟 Issue(s)

Resolves astronomer/issues#1762

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist


adam2k:
I'm logging in as a workspace admin & system admin and trying to delete a viewer user, but I'm getting this error:

```
$ astroDev auth login
Username (leave blank for oAuth):

Please visit the following URL, authenticate and paste token in next prompt

http://localhost:5000/token

oAuth Token: eyJhb....

$ astroDev deployment user delete avandover@gmail.com --deployment-id=ckgjxo7kk0449txc9eixg7xsm
Your Astro CLI Version (0.21.0) is ahead of the server version (0.20.0).
Consider downgrading your Astro CLI to match. See https://www.astronomer.io/docs/cli-quickstart for more information.
Insufficient permissions.
Error: Insufficient permissions.
```

In Houston API the logs only show:

```
2020-10-22T18:41:06 ERROR Insufficient permissions.
```


Screen Shot 2020-10-22 at 6 43 25 PM

src/schema/permissions.js
```diff
@@ -168,6 +168,18 @@ export const permissions = shield(
           hasRBAC("system.deployments.update")
         )
       ),
+      deploymentRemoveUserRole: and(
+        isAuth,
+        hasRBAC("deployment.userRoles.update"),
+        hasRBAC("system.deployments.update")
```
adam2k, line 174: This is the bug that's preventing access to run this mutation for any users:

```Suggestion
        or(
            hasRBAC("deployment.userRoles.update"),
            hasRBAC("system.deployments.update")
        )
```

src/resolvers/mutation/deployment-add-user-role/index.unit.test.js
```diff
@@ -11,15 +11,15 @@ import {
 // Define our mutation
 const query = `
   mutation deploymentAddUserRole(
-    $userUuid: Uuid,
+    $userId: Id,
```
adam2k, line 14: You can leave this in this PR, but do you mind adding these changes to another PR so we can merge them in faster and then I can update the `Astro CLI` code that this affects?

src/resolvers/mutation/deployment-add-user-role/index.unit.test.js
```diff
@@ -29,7 +29,7 @@ const query = `
 `;
 
 describe("deploymentAddUserRole", () => {
-  let email, createRoleBindingId, workspaceId, userUuid, deploymentUuid;
+  let email, createRoleBindingId, workspaceId, userId, deploymentId;
```
adam2k, line 32: 🥳 Thanks for changing to `Id`!

adam2k (changes requested):
There is the permissions suggestion that needs to go in for this PR to work correctly.  The other comments are mostly nits.

I would suggest getting that `deployment-add-user-role` code into a different PR though because we'll be able to merge that fast and address the changes in `Astro CLI` for the code that's already merged in since it is a breaking change there.

src/resolvers/mutation/deployment-add-user-role/index.unit.test.js
kushalmalani: Its already in the other PR. I rebased this branch of that, so it has latest changes.

src/schema/permissions.js
kushalmalani: Yes, I will have to merge them in order, because i will need to resolve conflicts since they all touch the same files.

src/schema/mutation.js
```diff
@@ -151,6 +151,29 @@ export default mutationType({
       resolve: mutations.deploymentAddUserRole
     });
 
+    t.field("deploymentRemoveUserRole", {
+      type: "RoleBinding",
+      nullable: true,
```
andriisoldatenko, line 156:
```suggestion
      nullable: true,
      description : `Some useful text.`,
```

src/schema/mutation.js
```diff
@@ -151,6 +151,29 @@ export default mutationType({
...
+      args: {
+        userId: arg({ type: "Id" }),
```
andriisoldatenko, line 158: pls add `description`

src/schema/types.js
```diff
@@ -413,6 +413,11 @@ const Uuid = scalarType({
   serialize: value => value
 });
 
+const Id = scalarType({
```
andriisoldatenko, line 416: do you think we need this?

kushalmalani: Yes, houston throws a `Missing type Id, did you forget to import a type to the root query or mean ID, Uuid?` if we dont define that

src/schema/permissions.js
adam2k: If you merge and fix conflicts this bug will still be here, right?

src/resolvers/mutation/deployment-add-user-role/index.unit.test.js
adam2k: What I mean is this mutation is already merged into master and this PR shouldn't conflate the changes for `deploymentAddUserRole` https://github.com/astronomer/houston-api/pull/513.  I'm recommending that these changes are a hotfix/patch for the PR that you already have merged into main because the other three that are opened for the remaining work are unreleated to `deplymentAddUserRole` unless I'm missing another PR?

src/schema/permissions.js
kushalmalani: What is the deployment role of the user adam@astronomer.io? I dont see this on graphql.

src/resolvers/mutation/deployment-add-user-role/index.unit.test.js
kushalmalani: They are not, but all of them are still addressing the same ticket to support deployment level RBAC. I am happy to make similar change in a separate PR, we would just have another PR to review. Trying to move the ball fast, since there is more work to get to, once these PRs go in. Let me know if you still want to keep them separate.

src/schema/permissions.js
adam2k: I'm testing this locally.  The roles are `SYSTEM_ADMIN` and `WORKSPACE_ADMIN`.  This suggestion matches the rules for all of the other mutations that you added.  the `or (...` is only missing from `delete` here.

Both Add and Update Deployment User Roles have the `or` and they don't return the permissions error.  Locally when I add this code everything works as planned too.

src/resolvers/mutation/deployment-add-user-role/index.unit.test.js
adam2k: The mutation changes for `deploymentAddUserRole` are only here and in the implementation file, right?  

I think this has the potential to make the git history in main murky because these changes are unrelated to delete deployment user.  If we need to cherry-pick or roll back for any reason this PR squashed will impact delete deployment user role and add deployment user role.  

I can help you out if you want because I need to patch this in Astro CLI too.

kushalmalani: I can add the patch

kushalmalani: https://github.com/astronomer/houston-api/pull/531

src/resolvers/mutation/deployment-remove-user-role/index.js
```diff
@@ -0,0 +1,84 @@
...
+  });
+
+  if (isEmpty(roleBindings) || roleBindings.length === 0) {
+    throw new MissingRoleBindingError();
+  }
+
+  // Remove the RoleBinding.
+  return await ctx.prisma.roleBinding.delete({
+    where: {
+      id: get(first(roleBindings), "id")
```
andriisoldatenko, line 81: what if `get ` returns null?
cc @adam2k

andriisoldatenko (changes requested):

src/resolvers/mutation/deployment-remove-user-role/index.js
adam2k: if there is no default this won't return anything and the result will be `undefined`.  That would crash javascript, but on line 74 @kushalmalani is checking if roleBindings is empty and if the roleBindings.length is 0.  I think the assumption is that if there is at least one roleBinding then the roleBinding must have an `id`.

kushalmalani: Thats accurate

andriisoldatenko: ok

src/schema/permissions.js
kushalmalani: @adam2k Should be working now. Can you pull latest on houston and then login via cli with new token and validate it.

kushalmalani:
Changes merged in https://github.com/astronomer/houston-api/pull/522

PR #536, opened 2020-10-28
2034 - Openshift: Add AppConfg options for Prometheus and ElasticSearch

samblackk:
## Description

Allows sending elasticsearch.enabled and prometheus.enabled booleans to the client via `AppConfig` query.
**Client-side:** https://github.com/astronomer/astro-ui/pull/875

## 🎟 Issue(s)

Attached to https://github.com/astronomer/issues/issues/2034 

## 🧪 Functional Testing

1. In the houston config - set prometheus.enabled or elasticsearch.enabled to false.
2. Startup the API
3. Open the UI and go to a deployment
4. Notice whatever component you disabled no longer has a tab/page

## 📋 Checklist

- [x] Updated any related [documentation](https://github.com/astronomer/docs/)


src/generated/nexus.ts
```diff
@@ -109,9 +109,9 @@ export interface NexusGenScalars {
```
adam2k, line 109: @andriisoldatenko these are the same `nexus.ts` updates that I am experiencing with local dev.

src/resolvers/query/app-config/index.js
```diff
@@ -9,6 +9,8 @@ export default async function appConfig() {
     baseDomain: config.get("helm.baseDomain"),
     smtpConfigured:
       config.get("email.enabled") && !!config.get("email.smtpUrl"),
-    manualReleaseNames: config.get("deployments.manualReleaseNames")
+    manualReleaseNames: config.get("deployments.manualReleaseNames"),
+    prometheusEnabled: config.get("prometheus.enabled"),
```
adam2k, line 13: Is this going to impact the `deploymentStatus` implementation?  I'm assuming this would make the status always show `UNKNOWN` unless there is a way that Houston API can still connect to some other cluster `prometheus`?

samblackk: Yes, it will impact the depStatus. I think we might just want to default to `UNKNOWN` if prometheus is disabled. We need to account for places in the whole API where either ES or Prometheus is being used and provide graceful fallbacks if they're disabled. My PR is just 1 step towards this.

adam2k: @schnie maybe we should look into the long term idea for deployment status (using kube api) sooner than later so there won't be a degraded experience on OpenShift.

schnie: I think in most typical OpenShift environments we'll be able to keep prometheus enabled. The logging stack is the bigger issue in these worlds. That said, letting the platform gracefully degrade without these components is always good to think about.

I think we can rework the deployment status with the new AEO architecture. That stuff will be more controller based, we we'll have a better place to add all that logic.

adam2k: @schnie thanks!  Sounds good to me.

src/generated/nexus.ts
andriisoldatenko: ok

schnie:
@samblackk why did a lot of the schema go from required to non-required with this PR?

andriisoldatenko:
please rebase from main. make sure you are using correct prisma version locally/ remove prisma globally.

src/resolvers/query/app-config/index.js
vishwas-astro: @adam2k @schnie - Like @samblackk mentions do we need to find other places in the backend where ES or prometheus is being used and fallback gracefully? What are those places?

samblackk:
@andriisoldatenko rebased - seem to help the issue

andriisoldatenko (approved):
LGTM

schnie (approved):

schnie:
LGTM

src/schema/types.js
```diff
@@ -459,6 +459,14 @@ const AppConfig = objectType({
       nullable: true,
       description: "Can users specify their own release names for deployments?"
     });
+    t.boolean("prometheusEnabled", {
+      nullable: true,
+      description: "Is Prometheus (metrics) disabled?"
```
kushalmalani, line 464: Should the description be "Is Prometheus (metrics) enabled?" since this goes with key name? Same for Elastic Search

samblackk: Sorry - yes! Will update.

samblackk: Updated!

samblackk:
This is updated w/ all the feedback. We could get this into 0.22 if we merge today. Or we can hold until 0.23. 
Just let me know when to merge.

PR #540, opened 2020-10-28
Add scheduler replica selection

adam2k:
## Description

Add scheduler replica selection for any executor except for local executor.

## 🎟 Issue(s)

Resolves astronomer/issues#1866

## 🧪 Functional Testing

The deployment mutations that have an airflow version greater than 2.0.0 **and** the executor is not `LocalExecutor` can now specify how many scheduler replicas to request.

1. Start Houston API and visit the Graphql Playground
2. Create a deployment and update the Airflow version number to something `>= 2.0.0` in the database
3. Now try to update the deployment with scheduler replicas values (some number 1 - 4).

**Test mutation:**
```
mutation updateDeployment($deploymentUuid: Uuid!, $payload: JSON, $executor: ExecutorType, $workers: Workers, $webserver: Webserver, $scheduler: Scheduler, $config: JSON) {
  updateDeployment(deploymentUuid: $deploymentUuid, payload: $payload, executor: $executor, workers: $workers, webserver: $webserver, scheduler: $scheduler, config: $config) {
    id
    urls {
      type
      url
      __typename
    }
    properties
    description
    label
    releaseName
    status
    type
    version
    workspace {
      id
      __typename
    }
    config
    createdAt
    updatedAt
    __typename
  }
}
```

**Variables**
```
{
  "deploymentUuid":"ckgvk4lt60884i1c9p5qgb9xy",
  "executor": "CeleryExecutor",
  "scheduler": {
    "replicas": 4,
    "resources": {
      "limits": {
        "cpu": 1000,
        "memory": 1920
      },
      "requests": {
        "cpu": 500,
        "memory": 1920
      }
    }
  }
}
```
## 📸 Screenshots

**Successful Request:**

Screen Shot 2020-10-28 at 6 54 57 PM

**Airflow Version is behind `2.0.0`**

Screen Shot 2020-10-28 at 6 55 51 PM

**Replica count is out of acceptable range (1 - 4)**

Screen Shot 2020-11-03 at 10 42 51 AM

## 📋 Checklist


config/default.yaml
```diff
@@ -543,6 +543,9 @@ deployments:
         - flower
         - redis
       defaultExtraAu: 0
+      minSchedulerReplicas: 2
+      maxSchedulerReplicas: 4
+      minAirflowVersion: "2.0.0"
```
andriisoldatenko, line 548: please update `config/custom-environment-variables.yaml`

adam2k: Added.

config/default.yaml
```diff
@@ -543,6 +543,9 @@ deployments:
         - flower
         - redis
       defaultExtraAu: 0
+      minSchedulerReplicas: 2
```
vishwas-astro, line 546: I think I could be on airflow 2.0 but still choose to have only 1 scheduler replica. Right, @schnie ??

vishwas-astro:
@adam2k the scheduler replicas should not need to have anything with the executor choice? Should depend only on the airflow version chosen.
cc @schnie

config/default.yaml
adam2k: In the issue is says: `The default # of Schedulers = 2 with 10AU each (configurable by users)`.  Does that mean they can drop below the default? 

https://github.com/astronomer/issues/issues/1866

schnie: These don't need to be celery-specific. The number of schedulers is independent from the executor choice.

adam2k:
@samblackk do you mind taking a look and let me know if this is all that you'll need for the required Astro UI updates?

config/custom-environment-variables.yaml
```diff
@@ -170,6 +170,10 @@ deployments:
       __format: "json"
     connection: "DEPLOYMENTS__DATABASE__CONNECTION"
   airflowReleasesFile: "DEPLOYMENTS__AIRFLOW__RELEASES__FILE"
+  executors:
+    minSchedulerReplicas: "DEPLOYMENTS__EXECUTORS__MIN__SCHEDULER__REPLICAS"
+    maxSchedulerReplicas: "DEPLOYMENTS__EXECUTORS__MAX__SCHEDULER__REPLICAS"
+    minAirflowVersion: "DEPLOYMENTS__EXECUTORS__MIN__AIRFLOW_VERSION"
```
schnie, line 176: Are these still relevant with the current configuration structure?

config/default.yaml
schnie: Looks like these three settings are duplicated across the different executors. Since they are the same, should we consolidate to a common setting block?

config/default.yaml
andriisoldatenko: i think so.

adam2k: I updated to allow for `1`

config/custom-environment-variables.yaml
andriisoldatenko: @schnie what do you mean? should we restrict override it via env vars?

config/default.yaml
andriisoldatenko: good call.

config/custom-environment-variables.yaml
adam2k: @schnie is this comment related to consolidating the `config/default.yaml` comment too?

I can merge the settings in `config/default.yaml` and then this would work, right?

config/default.yaml
adam2k: Would we need these to vary for different executors?  If not then we can probably consolidate.  I'll need to update the code around how we're pulling these values too if we go that route.

config/custom-environment-variables.yaml
adam2k: Moved from `executors` to `components` and in `default.yaml` this is under `components: -name: "scheduler"`

src/lib/deployments/config/index.js
```diff
@@ -721,3 +722,41 @@ function parseImageTag(imageName) {
...
+  const schedulerConfig = find(components, ["name", "scheduler"]);
+  const {
+    minSchedulerReplicas,
+    maxSchedulerReplicas,
+    minAirflowVersion
+  } = schedulerConfig;
+  // Add executor scheduler defaults to the config
+  executorConfig.minSchedulerReplicas = minSchedulerReplicas;
+  executorConfig.maxSchedulerReplicas = maxSchedulerReplicas;
+  executorConfig.minAirflowVersion = minAirflowVersion;
```
adam2k, line 746: This implementation is a little tricky now because we have to add the `minSchedulerReplicas, maxSchedulerReplicas, minAirflowVersion` keys to the  `executorConfig` object.

Another option could be to pull these values in for `validateReplicaRange` and `isValidReplicaRange` functions.

@andriisoldatenko or @schnie what is your opinion?

src/lib/deployments/validate/index.unit.test.js
```diff
@@ -0,0 +1,46 @@
...
+import casual from "casual";
+import { AIRFLOW_EXECUTOR_DEFAULT } from "constants";
+
+describe("validate", () => {
+  const prisma = {
+    deployment: {
+      findMany: jest.fn()
+    }
+  };
+  const id = casual.uuid;
```
kushalmalani, line 12: Can we start using cuid() instead of casual.uuid?

andriisoldatenko (approved):

andriisoldatenko:
@schnie PTAL.

src/lib/deployments/validate/replicas/index.js
```diff
@@ -0,0 +1,145 @@
...
+}
+
+/**
+ * Compare the airflowVersion to the minimum required airflow version
+ * @param  {String} airflowVersion
+ * @param  {String} minAirflowVersion
+ * @return {Boolean} is the version behind
+ */
+function isAirflowVersionBehind(airflowVersion) {
+  const { minAirflowVersion } = getSchedulerReplicas();
```
ashb, line 74: @adam2k You dun a booboo here :)

```
Cannot destructure property 'minAirflowVersion' of '(0 , _config.getSchedulerReplicas)(...)' as it is undefined.
```

Which also means that this isn't covered by unit tests.

adam2k: Which environment are you using?  This means an env var is missing from one of the environment charts.

We have coverage over these lines, but everywhere we run the tests those env vars are available.

PR #553, opened 2020-11-05
Remove NOSUPERUSER from create user statement

jpweber:
## Description

Needed because of an issue with azure flexible postgres that this causes the following error.
`must be superuser to create superusers`
Even though we are not creating a super user.

## 🎟 Issue(s)

Related to  https://github.com/astronomer/issues/issues/1111

## 🧪 Functional Testing

This worked fine on azure to be able to create deployments. However testing on other clouds has not been done yet and needs to be done. Likewise any adverse effects of this change has not been tested for.  

## 📸 Screenshots

## 📋 Checklist


jpweber:
@andrii said to open this PR and platform dev team and work on the testing.

src/lib/deployments/database/index.js
```diff
@@ -177,7 +177,7 @@ export async function ensureUser(conn, user, password) {
   const exists = first(result.rows).exists;
   if (!exists) {
     await conn.raw(
-      `CREATE USER ${user} WITH LOGIN NOSUPERUSER NOCREATEDB NOCREATEROLE NOINHERIT NOREPLICATION CONNECTION LIMIT -1 ENCRYPTED PASSWORD '${password}';`
+      `CREATE USER ${user} WITH LOGIN NOCREATEDB NOCREATEROLE NOINHERIT NOREPLICATION CONNECTION LIMIT -1 ENCRYPTED PASSWORD '${password}';`
```
adam2k, line 180: @schnie this `ensureUser` is only used upon DB creation (the only reference that I found is here: https://github.com/astronomer/houston-api/blob/2d1a063e62e56ba010d43423d1e742ba8370dd3d/src/lib/deployments/database/index.js#L213)

Is it correct that this user is only for the first person into the system?  Is it safe to default to whatever the DB selects for the `SUPERUSER` setting to allow for compatibility for `Azure`?

If so, I'd recommend that we update the name of this function to something like `createFirstSystemUser` or something similar so this is SQL isn't inadvertently used anywhere else.  We can also update the `exists` check to make sure there aren't _any_ users in the DB yet.

adam2k: Also note that `NOSUPERUSER` is the default role. https://www.postgresql.org/docs/current/sql-createrole.html#id-1.9.3.78.6

schnie: This is all within the context of creating a new Airflow deployment. So we `ensureDatabase`, `ensureUser`, `ensureSchema`, etc. This chunk of the code is running after the new Airflow database has been created and we are ensuring (create or skip) that we have the user created. This users credentials then get passed into the Airflow helm values and get pushed into a secret that is mounted by the Airflow pods (or really just pgbouncer, these days).

schnie: @adam2k yea the fact that we're just specifying the default value of `NOSUPERUSER` seems to make it OK to drop here right? 

My only question is if this is actually behaving correctly in Azure. Like just need to ensure that this user _only_ has access to the single database it was created with. Seems like it would, but just feels like something to check on. Also makes me wonder if Azure supports `SUPERUSER`, or if that feature is dropped or reserved for system stuff or something.

adam2k: I'm checking in on the Azure docs and it looks like it _should_ accept `NOSUPERUSER`.

https://docs.microsoft.com/en-us/azure/postgresql/howto-create-users#how-to-create-database-users-in-azure-database-for-postgresql

@jpweber is there an Azure playground or somewhere that I can get access to look into this issue?  I'm wondering what the reason is setting this role is failing when it's in the Azure docs.

adam2k: Realized that you're looking at Azure Flexible server.  I'm checking docs here now: https://docs.microsoft.com/en-us/azure/postgresql/flexible-server/how-to-manage-server-portal

jpweber: I tested it running on azure flexible server so I know it works there.

adam2k: @jpweber does this issue _only_ occur for the Azure Flexible Server or for their standard postgresql server too?  It seems like something might be going on with the user access only in the `Azure Flexible Server` scenario.  I'm wondering if we should build an interface here to take into account the nuance for this particular provider?

jpweber: It only occurs on azure flexible server. We won't be using their standard Postgres because of other connection string related issues. For completeness we also do _not_ have this issue on other providers.

adam2k: @jpweber do you have a way to quickly check in the `Azure` DB that a user added to a new deployment doesn't have access to other DBs?  We'll have to add a non-sysadmin user and create two deployments, then add the user to both to confirm. 

I think this is safe to merge if that works.  If you can give me access to Astro UI I can do all the account setup and then I'd probably just need your help to confirm the DB access.

andriisoldatenko:
is it possible to perform practical tests with:
- [ ] postgresql in cluster mode (default in kind)
- [ ] google cloud SQL
- [ ] aws RDS
- [ ] azure postgresql
than we can say no regression and ready to merge. I can help to perform tests

ianstanton (approved):

vishwas-astro:
@adam2k @andriisoldatenko @kushalmalani  - Need to get this into 0.23 (slated for 11/19)

Can you guys please help test and merge this ASAP?

adam2k (approved):
This seems OK to me.  We can test the change in dev and see if it causes any negative side effects.  It is working in Azure, as per @jpweber's tests.

andriisoldatenko:
practically LGTM (tested only in local kind)

andriisoldatenko (approved):

andriisoldatenko:
@jpweber @adam2k go ahead and merge it and than let's ship to dev cluster.

PR #569, opened 2020-11-12
Default airflowVersion to latest available #2121

andriisoldatenko:
### Description

Default airflowVersion to latest available #2121 #569 

### 🎟 Issue(s)

Resolves https://github.com/astronomer/issues/issues/2121

Practical tests:

img

i

@schnie do we need to change logic if user creates deployment without specifying airflow version from astro-cli/graphql?

For instance right now we use `1.10.7` should we fallback to 1.10.12 or 2.0.0 (max available one)?


src/lib/deployments/config/index.js
```diff
@@ -692,7 +692,11 @@ export async function airflowVersions() {
 
   // Semver.coerce truncates Longer versions to simply ones (4.6.3.9.2-alpha2 becomes 4.6.3).
   const coercionVersions = map(rawVersions, semver.coerce);
-  return uniq(map(coercionVersions, "version"));
+
+  return uniq(map(coercionVersions, "version")).sort((a, b) =>
```
vishwas-astro, line 696: @andriisoldatenko could we add in the logic somewhere in this flow where we are using the latest "stable" AC version only as default. For eg: when we start allowing 2.0 beta AC image via updates.astronomer.io as an "edge" instead of "stable" image, I dont want it to be the default selected image.

May be we dont need to do this in this PR, but there is a requirement bubbling up to support airflow 2.0 beta image in our cloud soon and I want to prepare for that.

cc @schnie @astrorfox @paolaperaza

astrorfox: Let's defer that part of the requirement to the future. We'll also want to look at how we address builds tagged as EOL at the same time.

vishwas-astro: hmm ..Dont we still need it prior to shipping it as an LTS and making it available to our EE customers or may be, we can ship that logic as a patch on the 0.23 LTS release

astrorfox: Ah, I see where you're going @vishwas-astro. Let's have 0.23 look for a stable tag and ignore everything else. We can introduce other tags (early, EOL, etc.) and responses to them later, but at least this insulates 0.23.0 from any breaking changes there.

andriisoldatenko: @vishwas-astro IMO it's out of scope of this ticket isn't ?

andriisoldatenko: @vishwas-astro could you please create new ticket and move conversation this this ticket?

schnie: Although it doesn't matter yet, we already are filtering to `stable` here. The first line in this function calls the function above to get the images, which are filtered to `stable`. https://github.com/astronomer/houston-api/pull/569/files#diff-76bbbdeaa45e56ed3cc7dc4a6d00210ea4c81a0ff883cbd594b8c6cae42ba76aR678-R680

schnie (approved):
LGTM

src/lib/deployments/config/index.js
adam2k, line 696: This is a nit:

```Suggestion
  const uniqueVersions = uniq(map(coercionVersions, "version"));
  return uniqueVersions.sort((a, b) =>
```

adam2k (approved):
LGTM! 🎉  One nit for readability.

src/lib/deployments/config/index.js
astrorfox: @schnie to confirm, we can add a 'non-stable' entry into the updates.astronomer.io file and it will not impact this behavior?

andriisoldatenko: nope. so it's pretty safe.

schnie: I believe we _can_ add a non-`stable` entry updates.astronomer.io. This code should filter those out even though they'll exist in the local db.

astrorfox: Thank you, both. Can we add a test for this case to ensure it stays that way?

PR #571, opened 2020-11-13
Create Workspace Service Account Synthetic Role bindings

kushalmalani:
## Description

Create Workspace Service Account Synthetic Role bindings

## 🎟 Issue(s)

Resolves astronomer/issues#2156

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist


schnie (approved):

andriisoldatenko:
practically LGTM, i was able to astro deploy locally via workspace service account.

andriisoldatenko (approved):

src/lib/rbac/index.js
```diff
@@ -201,6 +202,64 @@ export async function getAuthUser(prisma, authorization) {
...
+        }
+      },
+      deletedAt: null
+    },
+    include: {
+      workspace: true
+    }
+  });
+
+  // Generate fake rolebindings for deployment level admin.
```
adam2k, line 237: I'm not understanding the comments about `fake roleBindings`.  From the issue it seems like they're implicit roleBindings, so that the Service Account users can still operate on workspace deployments, right?  

Can you change all the comments from `fake` to `implicit` or note something about how this code exists for `Service Accounts`.  I think it'll make sense when somebody comes across this code later.

src/lib/rbac/index.js
```diff
@@ -201,6 +202,64 @@ export async function getAuthUser(prisma, authorization) {
...
+      workspace: null,
+      deployment: { id: deployment.id }
+    };
+  });
+
+  // Combine with existing real roleBindings.
+  const roleBindings = [...user.roleBindings, ...deploymentRoleBindings];
+
+  // Return a modified user, spreading existing rolebindings, with new fake ones.
+  return { ...user, roleBindings };
```
adam2k, line 260: Can we move all this transform code into a central location, so it's not buried in with the `lib/rbac` code?

It might not cause bugs now, but it probably will in the future.

@schnie or @andriisoldatenko do you all have ideas on where these object transforms should go?  Should we create a `service` directory that lives outside of `lib`?  Then we could have a subdirectory like `service/transforms`?

src/lib/rbac/index.js
```diff
@@ -231,6 +290,10 @@ async function addSyntheticRoleBindings(prisma, user) {
   return addUserRoleBinding(user);
 }
 
+async function addServiceAccountSyntheticRoleBindings(prisma, user) {
+  return addUserRoleBinding(await addDeploymentRoleBindings(prisma, user));
```
adam2k, line 294: This is a nit:

Can you invoke `await addDeploymentRoleBindings(prisma, user)` on the line above and pass the result into `addUserRoleBinding`?  It'll make this a little easier to read.

src/resolvers/types/service-account/index.js
```diff
@@ -13,13 +13,13 @@ export function apiKey({ apiKey, createdAt }) {
...
+    get(parent, "roleBinding.workspaceId") ||
+    get(parent, "roleBinding.deploymentId")
   );
 }
 
 export function entityType(parent) {
-  return get(parent, "roleBinding.workspace.id")
+  return get(parent, "roleBinding.workspaceId")
     ? ENTITY_WORKSPACE
     : ENTITY_DEPLOYMENT;
```
adam2k, line 24: Should this always default to `ENTITY_DEPLOYMENT`?  Are there potential downstream errors?  Should we throw if `get(parent, "roleBinding.workspaceId")` and `get(parent, "roleBinding.deploymentId")` do not exist instead?

src/lib/rbac/index.js
schnie: I dunno, I feel like `lib/rbac` is the central location and its related to RBAC. All its doing is mapping Workspace-level Roles to additional Deployment-level Roles. This code was here before Deployment-level RBAC, @kushalmalani is just bringing it back to life for Service Accounts to fix an issue we didn't anticipate. Doesn't really feel like a `service` to me.

adam2k: I didn't realize this code was reverted from a previous iteration.  I think we can leave as-is and maybe improve the comments.

I'm thinking of this in terms of how we are doing transforms in general.  There are other areas where transforms on objects occur, and because nothing is typed, in those cases it's more difficult to know what the shape of the object is at any given point.  I think it's something that we can work on as we go.  It's probably out of scope for this issue.

src/lib/rbac/index.js
adam2k: Don't worry about doing this for now because this code was reverted from a previous change.

src/lib/rbac/index.js
schnie: Agree on the lingo here

src/resolvers/types/service-account/index.js
kushalmalani: I think this should be ok. I believe we only use this for a label on astroUI. Plus the logic is a year old, not sure what would break trying to refactor that. Out of the scope for this ticket. We can look at it separately so we dont block the patch for this. What do you think? @adam2k

PR #575, opened 2020-11-17
Fix for system admin user unable to view workspace deployments

kushalmalani:
## Description

Fix for system admin user unable to view workspace deployments

## 🎟 Issue(s)

Resolves astronomer/issues#2155

## 🧪 Functional Testing

## 📸 Screenshots

Screen Shot 2020-11-16 at 5 21 56 PM

## 📋 Checklist


src/resolvers/query/workspace-deployments/index.js
```diff
@@ -39,16 +41,23 @@ export function deploymentsQuery(args, ctx) {
...
-      in: compact(deploymentIds)
+  const deploymentIds = [];
+  ctx.user.roleBindings.filter(rb => {
+    if (get(rb, "role") === SYSTEM_ADMIN) {
+      isSuperAdminUser = true;
     }
+    deploymentIds.push(get(rb, "deployment.id", null));
   });
 
+  // Super Admin User should be able to see all deploymnets under that workspace
```
adam2k, line 52:
```Suggestion
  // Super Admin User should be able to see all deployments under that workspace
```

src/resolvers/query/workspace-deployments/index.js
```diff
@@ -24,6 +25,7 @@ export default async function workspaceDeployments(parent, args, ctx) {
 export function deploymentsQuery(args, ctx) {
   // Pull out some args.
   const { releaseName, workspaceUuid } = args;
+  let isSuperAdminUser = false;
 
   // Build query structure.
   let query = { where: { AND: [] } };
```
adam2k, line 31: I think this can be a `const`

kushalmalani: I am overriding he variable to true down in code if we find a system admin role.

src/resolvers/query/workspace-deployments/index.js
```diff
@@ -39,16 +41,23 @@ export function deploymentsQuery(args, ctx) {
...
-  query.where.AND.push({
-    id: {
-      in: compact(deploymentIds)
+  const deploymentIds = [];
+  ctx.user.roleBindings.filter(rb => {
+    if (get(rb, "role") === SYSTEM_ADMIN) {
+      isSuperAdminUser = true;
     }
+    deploymentIds.push(get(rb, "deployment.id", null));
   });
```
adam2k, line 50: Something like this might cut down on the lines of code here.
```Suggestion
  const isSuperAdminUser = find(rb, r => {
    deploymentIds.push(get(rb, "deployment.id", null));
    return r.role === SYSTEM_ADMIN
  }) !== undefined;
```

src/resolvers/query/workspace-deployments/index.js
adam2k: Maybe I'm missing something...what line does that happen on?  Also, why would we change the data type of a variable?

src/resolvers/query/workspace-deployments/index.js
kushalmalani: what are we trying to optimize here? Your suggestion seems way too difficult to read and understand.

src/resolvers/query/workspace-deployments/index.js
adam2k: Oh!  You're talking about `let isSuperAdminUser = false;`.  My comment is about `let query = { where: { AND: [] } };`

src/resolvers/query/workspace-deployments/index.js
adam2k: Mainly so you don't need to redefine `let isSuperAdminUser = false;`

adam2k: There is probably a better way to do it with lodash, but that was the quickest thing that I could come up with.  Another idea is to do two loops.  One where you check for `isSuperAdminUser` and another where you push the `rb.deployment.id`s into the `deploymentIds`?

kushalmalani: Do we really need to refactor this? Unless you think this would cause a significant change

adam2k (approved):
This LGTM! 🎉

PR #581, opened 2020-11-19
Create deployments with right UID/GID for the distro

ashb:
The Debian image has some hacks in place to "work" when run as 100:101
(the defaults from the Alpine image) but it has a few draw backs.

This change makes us use Debian images by default for new deployments,
and then on image push examines the labels to work out what UID/GID we
should run as.

Previously the dist that was used for new deployments was "ill-defined"
-- it depended on what order the tags for the version were added to the
AirflowReleases table. It is now an explict choice (sadly based on the
string appearing in the tag -- the best info we have) rather than random
choice.

## 🎟 Issue(s)

Resolves astronomer/issues#2100

## 🧪 Functional Testing

## 📸 Screenshots

## 📋 Checklist


ashb:
I haven't yet added tests, (or even checked this actually works!) but feedback on the approach?

andriisoldatenko (approved):

src/lib/deployments/config/index.js
```diff
@@ -706,7 +707,23 @@ export async function airflowVersions() {
  * @reuturn {String} Default Airflow image.
  */
 export async function defaultAirflowImage() {
-  return first(filter(await airflowImages(), i => i.channel === "stable"));
+  const releases = await airflowImages();
+
+  // Work out what the first version is
+  const version = releases[0].version;
+
+  const releasesOfVersion = filter(releases, i => i.version == version);
```
adam2k, line 715: Is this a typo or are you intentionally trying to avoid equal type?

```Suggestion
  const releasesOfVersion = filter(releases, i => i.version === version);
```

src/lib/deployments/config/index.js
```diff
@@ -806,3 +823,47 @@ export function removeLegacyDeploymentReplicas(deployment, args) {
...
+  // Play it safe -- old images didn't have any of these labels, and those were always Alpine images.
+  const distro = get(
+    currentImage.labels,
+    "io.astronomer.docker.distro",
+    "alpine"
+  );
+
+  const defaultValues = {
+    alpine: { uid: 100, gid: 101 },
+    debian: { uid: 50000, gid: 50000 }
```
adam2k, line 849: Should these values go into the `constants/index.js`?

adam2k (commented):
I think this looks good, but wanted confirmation on the `releasesOfVersion` comment for the `==` vs `===`.

src/lib/deployments/config/index.js
ashb: I rarely use `===` is all :) ENOTAJSDEV

src/lib/deployments/config/index.js
andriisoldatenko: @adam2k which part do you prefer to put to constants ?

src/lib/deployments/config/index.js
adam2k: I'll commit it and do some further testing.

src/lib/deployments/config/index.js
adam2k: We have one array in the `constants` right now, so I could see either only adding the numbers to the constants: `100`, `101`, `5000`, `50000`  OR we could add two objects ` { uid: 100, gid: 101 }` and `{ uid: 50000, gid: 50000 }`.

I'm kind of leaning towards the objects as constants.

andriisoldatenko: prefer only numbers too.

andriisoldatenko:
@ashb @adam2k PTAL.

src/lib/deployments/config/index.js
```diff
@@ -806,3 +827,47 @@ export function removeLegacyDeploymentReplicas(deployment, args) {
     get(args, "config.scheduler") && delete args.config.scheduler.replicas;
   }
 }
+
+export async function containerUidGid(deployment) {
+  const currentImage = await prisma.dockerImage.findFirst({
+    where: { deploymentId: deployment.id },
+    orderBy: { createdAt: "desc" }
```
ashb, line 834: updatedAt maybe instead? (Just thinking if someone re-pushes an existing image, do we update the updateAt column)

ashb (commented):
Lgtm other than one small comment

ashb:
@andriisoldatenko LGTM - I can't approve it as it's my own PR :grin: Do we want to wait for Adam to review again, or should we merge this and the chart change?

src/lib/deployments/config/index.js
andriisoldatenko: @ashb good point, but:
```
Unknown arg `updatedAt` in orderBy.updatedAt for type DockerImageOrderByInput. Did you mean `createdAt`? Available args:
```

andriisoldatenko:
@ashb practically LGTM:
```
2020-11-20T15:12:43 INFO Final helm values:
uid: 100
gid: 101
rbacEnabled: true
ingress:
    enabled: true
    baseDomain: local.astronomer.io
    class: deadly-terrier-nginx
    webserverAnnotations:
        nginx.ingress.kubernetes.io/server-snippet: "\n             location ^~ /api/ {\n               if ($host = 'deployments.local.astronomer.io' ) {\n                 return 404;\n               }\n               proxy_pass http://deployments.local.astronomer.io/translunar-cosmic-1367/airflow/$request_uri;\n             }"

```

and db records:
```
                                                  labels

-------------------------------------------------------------------------------------------------------------------------------
-------------------------------------------------------------------------------------------------------------------------------
----------------------------------------------------------------------------------------------------------
 {"maintainer": "Astronomer <humans@astronomer.io>", "io.astronomer.docker": "true", "io.astronomer.docker.distro": "alpine", "
io.astronomer.docker.module": "airflow", "io.astronomer.docker.component": "airflow", "io.astronomer.docker.ac.version": "1.10.
12-1", "io.astronomer.docker.airflow.onbuild": "true", "io.astronomer.docker.airflow.version": "1.10.12"}
(1 row)

```

andriisoldatenko:
found small issue:
```
small issue, in current PR
3:21
https://github.com/astronomer/ap-airflow/blob/master/1.10.12/buster/Dockerfile#L23

1.10.12/buster/Dockerfile:23
ARG ASTRONOMER_UID="50000"
<https://github.com/astronomer/ap-airflow|astronomer/ap-airflow>astronomer/ap-airflow | Added by GitHub
3:22
as you can see label value is string
3:22
 {"maintainer": "Astronomer <humans@astronomer.io>", "io.astronomer.docker": "true", "io.astronomer.docker.uid": "50000", "io.a
stronomer.docker.distro": "debian", "io.astronomer.docker.module": "airflow", "io.astronomer.docker.component": "airflow", "io.
astronomer.docker.ac.version": "", "io.astronomer.docker.airflow.onbuild": "true", "io.astronomer.docker.airflow.version": "1.1
0.12"}
houston record ^
3:22
2020-11-20T15:17:35 INFO Final helm values:
uid: '50000'
gid: 50000
rbacEnabled: true
ingress:
    enabled: true
3:22
do you think i need to convert all default values to strings also?
3:22
just to sync or convert value from label to int?
```

ashb:
I think convert from label to int

andriisoldatenko:
> I think convert from label to int

done in 17b59d862f58130c128aef99f72536bddba9fd01

src/lib/deployments/config/index.js
ashb: Oh, I was looking AirflowRelease which does have an updatedAt.

NM

ashb:
K, we good to merge?

